Issue #25, opened 2026-02-24
Telegram streaming: collapse repeated tool indicators into counter


tfatykhov:
## Problem

During streaming, each tool call appends a new indicator line ("💡 Learning...", "⚙️ Running...") via `StreamingMessage.append_tool_indicator()`. With many sequential tool calls, this creates a wall of repeated status lines in the Telegram message.

## Solution

Replace the append behavior with a grouped counter approach:
- Track tool calls by category
- Show a single line like "⚙️ Running... (15)" that updates in place
- Only the latest/active indicator is shown, with a count if repeated

## Files

- `nous/telegram_bot.py` — `StreamingMessage.append_tool_indicator()`

## Acceptance Criteria

- [ ] Repeated tool calls of the same type show as a single line with count
- [ ] Different tool types can coexist (e.g. "🔍 Searching... ⚙️ Running... (3)")
- [ ] No regression in streaming text display
- [ ] Tests updated

PR #26, opened 2026-02-24
feat: token usage tracking (#22) + collapsed tool indicators (#25)

### LINKED ISSUE #22: Track and display API token usage (input/output tokens)
## Problem

Nous doesn't capture or display token usage from the Anthropic API. No visibility into:
- How many tokens each turn consumes
- Input vs output token breakdown
- Per-session cumulative usage
- Cost awareness (especially on pay-per-token plans)

## Context

The Anthropic API returns `usage` in every response:
```json
{
  "usage": {
    "input_tokens": 1523,
    "output_tokens": 384
  }
}
```

This data is currently ignored in `runner.py`.

### 1. Capture usage from API response (~10 lines in `runner.py`)
Extract `usage` dict from `_call_api()` response. For tool loops, accumulate across turns.

```python
# In _call_api() or _tool_loop()
usage = response_data.get("usage", {})
total_input += usage.get("input_tokens", 0)
total_output += usage.get("output_tokens", 0)
```

### 2. Return in /chat response (~5 lines in `rest.py`)
Add `usage` field to the chat response JSON:

```json
{
  "response": "...",
  "session_id": "...",
  "usage": {
    "input_tokens": 2847,
    "output_tokens": 512,
    "total_tokens": 3359
  }
}
```

### 3. Telegram footer (~10 lines in `telegram_bot.py`)
Append token stats to bot messages:

```
[response text]

📊 2.8K in / 512 out
```

### 4. Per-session accumulation (~15 lines in `runner.py`)
Track cumulative tokens per session in the conversation LRU:

```python
session_usage[session_id] = {
    "total_input": 0,
    "total_output": 0,
    "turn_count": 0
}
```

Accessible via `/chat` with `debug=true` or a new `/usage` endpoint.

## Scope
~40 lines across 3 files. No schema changes.

## Labels
enhancement, runtime, observability

tfatykhov:
## Changes

Fixes #22 and #25.

### Token Usage Tracking (#22)
- Accumulate `input_tokens`/`output_tokens` from Anthropic API across tool loop iterations
- `run_turn()` and `stream_chat()` now return usage stats
- `/chat` REST endpoint includes `usage` in response JSON
- `/chat/stream` SSE includes `usage` in the final `done` event
- Telegram bot shows footer: `📊 2.8K in / 512 out`

### Collapsed Tool Indicators (#25)
- Replace append-per-tool behavior with grouped counters
- Same tool type shows as `⚙️ Running... (15)` instead of 15 separate lines
- Different tool types shown separately: `🔍 Searching... ⚙️ Running... (3)`
- `finalize()` clears indicators, shows only final response text + usage footer

### Files Changed
- `nous/api/runner.py` — usage accumulation in `_tool_loop`, `stream_chat`, `run_turn`
- `nous/api/rest.py` — usage in `/chat` and `/chat/stream` responses
- `nous/telegram_bot.py` — `StreamingMessage` rewrite with `_tool_counts`, usage footer
- Tests updated: `test_streaming.py`, `test_runner.py`, `test_rest.py`, `test_mcp.py`

### DIFF
--- a/nous/telegram_bot.py
+++ b/nous/telegram_bot.py
@@ -75,8 +75,7 @@ class StreamingMessage:
 
     async def append_text(self, text: str) -> None:
         """Append text delta to the base text and update display."""
-        self._base_text += text
-        await self.update(self._base_text)
+        await self.update(self._base_text + text)
 
     async def update(self, new_text: str) -> None:
         """Update message text. Creates on first call, edits after."""
@@ -122,8 +121,7 @@ class StreamingMessage:
         if self._usage:
             self.text += f"\n\n{format_usage_footer(self._usage)}"
 
-        if self._pending or self.message_id is None:
-            await self._send_or_edit()
+        await self._send_or_edit()
 
     async def _send_or_edit(self) -> None:
         if not self.text.strip():


tfatykhov:
🔍 **CodeReviewer** (Re-review)

## Summary

All **P1** and **P2** issues from the previous review are resolved. ✅ One new **P2** bug found in the `finalize()` path, plus two **P3** notes.

---

## Previously-Reported Issues — Status

| # | Issue | Status |
|---|-------|--------|
| P1 | `mcp.py run_turn()` 2-tuple unpack | ✅ Fixed — now 3-tuple with `_usage` |
| P1 | `test_tool_loop.py _tool_loop()` 2-tuple unpack | ✅ Fixed — all 6 call-sites updated |
| P2 | `input_tokens` always 0 | ✅ Fixed — `message_start` parsed in `_parse_sse_event` and accumulated in `stream_chat` |
| P2 | Private `_base_text` access | ✅ Fixed — `append_text()` public method added; streaming path uses it |
| P2 | Duplicated usage formatting | ✅ Fixed — `format_usage_footer()` extracted and used in both `_chat` and `finalize()` |

---

## New Issues Found

### 🔴 P2 — `finalize()` silently skips the final Telegram edit

**File:** `nous/telegram_bot.py`, `StreamingMessage.finalize()`

```python
async def finalize(self) -> None:
    self._tool_counts.clear()
    self.text = self._build_display_text()   # ← text ALWAYS changed here

    if self._usage:
        self.text += f"\n\n{format_usage_footer(self._usage)}"  # ← and here

    if self._pending or self.message_id is None:   # ← but edit only sent sometimes
        await self._send_or_edit()
```

`finalize()` **always** modifies `self.text` (clears tool indicators, appends usage footer), but `_send_or_edit()` is only called when `_pending=True` OR the message doesn't exist yet.

In the **happy path** — streaming completes cleanly, the last text delta was sent successfully (not rate-limited, so `_pending=False`), and the message already exists (`message_id is not None`) — `finalize()` updates the in-memory `self.text` but **never pushes the edit to Telegram**. The user sees stale tool indicators and no usage footer.

**Fix:** Call `_send_or_edit()` unconditionally:
```python
async def finalize(self) -> None:
    self._tool_counts.clear()
    self.text = self._build_display_text()
    if self._usage:
        self.text += f"\n\n{format_usage_footer(self._usage)}"
    await self._send_or_edit()  # Always flush the final state
```

**Test gap:** `test_finalize_clears_indicators_adds_usage` checks `sm.text` only — it would pass even with the bug since it never asserts `_send_or_edit` was called.

---

### 🟡 P3 — Dead code: `input_tokens` from `done` event in `stream_chat`

**File:** `nous/api/runner.py`, `stream_chat()`

```python
elif event.type == "done":
    stop_reason = event.stop_reason
    if event.usage:
        total_usage["input_tokens"] += event.usage.get("input_tokens", 0)  # ← always adds 0
        total_usage["output_tokens"] += event.usage.get("output_tokens", 0)
```

The `done` event is emitted from Anthropic's `message_delta` SSE event, whose `usage` field contains **only** `output_tokens`. `input_tokens` is only present in `message_start`, which is correctly handled in the `message_start` branch. The line above always adds `0` and can be removed without changing behavior.

---

### 🟡 P3 — `append_text` sets `_base_text` twice

**File:** `nous/telegram_bot.py`

```python
async def append_text(self, text: str) -> None:
    self._base_text += text          # set #1
    await self.update(self._base_text)

async def update(self, new_text: str) -> None:
    self._base_text = new_text       # set #2 (same value)
    ...
```

`append_text` sets `_base_text` then passes it to `update` which sets it again. Functionally correct, but creates a coupling between the two methods. Consider having `append_text` directly call `_build_display_text()` and `_send_or_edit()` instead of routing through `update`.

---

## Verdict

**The P2 `finalize()` bug should be fixed** — it means the usage footer and cleared tool indicators are effectively silently dropped in the common streaming case. Everything else looks solid. The `ApiResponse.usage` concern from my initial pass is confirmed non-issue — usage is already deserialized as a plain dict.

Fix the `finalize()` guard and this is ready to merge.

---
*🔍 CodeReviewer*

PR #65, opened 2026-02-27
fix: use original hybrid scores in recall_deep instead of RRF (#64)

### LINKED ISSUE #64: recall_deep discards real search scores, uses RRF positional scores instead
## Problem

`recall_deep` tool calls `heart.recall()` which uses Reciprocal Rank Fusion (RRF) to merge results across memory types. The RRF scoring (`1/(k+rank)` with k=60) **discards the actual hybrid search scores** from `hybrid_search()` and replaces them with positional scores.

This means:
- All facts cluster at 0.014-0.016 regardless of actual relevance
- The agent cannot distinguish high-quality matches from noise
- Debug output is misleading — scores look flat when real scores vary widely (0.04-0.57)
- Two completely different scoring systems: context engine uses real hybrid scores (0-1.0), recall_deep uses RRF (0.014-0.016)

## Evidence

When Nous runs `recall_deep` for "what do you know about nous":
- Facts: all 0.014-0.016 (RRF positional)
- Decisions: 0.074-0.289 (real vector scores from brain.query)

Context engine for same query:
- Facts: 0.567, 0.468, 0.404 (real hybrid scores)

## Root Cause

`heart.py:_recall()` line ~479:
```python
rrf_score = 1.0 / (k + rank)
recall_result = self._to_recall_result(memory_type, item, rrf_score)
```

The original scores from `hybrid_search()` are available on the search results but never passed through.

## Proposed Fix

Pass original search scores through to `RecallResult` instead of (or alongside) RRF scores. Options:
1. Use original scores directly, apply RRF only for cross-type ordering
2. Store both: `score` (original) + `rank_score` (RRF) so agent sees real relevance
3. Normalize original scores per-type, then merge

## Impact
- Agent quality: better self-assessment of what it knows
- Debug transparency: scores reflect actual relevance
- Consistency: same scoring language as context engine

tfatykhov:
## Problem

`recall_deep` tool used RRF positional scoring (`1/(k+rank)`, k=60), which replaced actual hybrid search scores with values clustering at 0.014-0.016. The agent could not distinguish high-relevance matches from noise.

## Root Cause

`heart.py:_recall()` applied RRF scoring intended for cross-type merging, but discarded the original `hybrid_search()` scores that each sub-search already computed.

## Fix

Use the original hybrid scores (0.7×vector + 0.3×keyword) from each sub-search directly. All Heart memory types use the same `hybrid_search()` formula, so scores are directly comparable without RRF normalization.

**Before:** Facts all 0.014-0.016 (RRF positional)
**After:** Facts show real relevance (e.g. 0.567, 0.468, 0.404)

## Changes

- `nous/heart/heart.py`: Replace RRF scoring in `_recall()` with original hybrid scores from search results
- No schema changes needed — `RecallResult.score` field unchanged
- Existing tests pass (they assert `score > 0`, which still holds)

Closes #64

### DIFF
--- a/nous/heart/heart.py
+++ b/nous/heart/heart.py
@@ -460,10 +460,10 @@ class Heart:
                 keys.append(memory_type)
                 results_list.append(exc)
 
-        # Use original hybrid search scores (vector + keyword) instead of
-        # RRF positional scores. All Heart sub-searches use the same
-        # hybrid_search() formula (0.7*vector + 0.3*keyword), so scores
-        # are directly comparable across memory types.
+        # Use original search scores instead of RRF positional scores.
+        # Episodes, facts, and procedures use hybrid_search() (0.7*vector
+        # + 0.3*keyword). Censors use cosine similarity. Scores are
+        # comparable enough for meaningful cross-type ranking.
         merged: list[RecallResult] = []
 
         for memory_type, raw_results in zip(keys, results_list):


tfatykhov:
## 🔍 CodeReviewer — PR #65 Review

**PR Summary:** Replace RRF positional scoring with original hybrid search scores in `Heart._recall()`.

The intent is correct — RRF scores lose the actual relevance signal. But the implementation has a critical bug that will **break an existing test** and silently corrupt censor ranking.

---

## 🔴 P1 — Critical: `CensorMatch` has no `.score` field → all censors get `score=0.0`

**File:** `nous/heart/heart.py`, line 479  
**Code:** `original_score = getattr(item, "score", None) or 0.0`

`CensorMatch` (defined in `schemas.py:229`) has **no `score` field** — only `id`, `trigger_pattern`, `action`, `reason`, `domain`. So `getattr(item, "score", None)` returns `None` for every censor result, and `None or 0.0` collapses to `0.0`.

**Effect:** All censor results are ranked dead-last with `score=0.0`, regardless of their actual semantic relevance.

**Test breakage:** `test_unified_recall` (tests/test_heart.py:312) asserts `r.score > 0` for **all** returned results including censors. With this PR, that assertion will **fail** for every censor item.

```python
# schemas.py — CensorMatch has no score field!
class CensorMatch(BaseModel):
    id: UUID
    trigger_pattern: str
    action: CensorAction
    reason: str
    domain: str | None
    # ← NO score field here!
```

**Fix:** Add `score: float | None = None` to `CensorMatch`, and populate it from `similarity` in `censors._semantic_search()`:

```python
# Step 1 — schemas.py: add score field to CensorMatch
class CensorMatch(BaseModel):
    id: UUID
    trigger_pattern: str
    action: CensorAction
    reason: str
    domain: str | None
    score: float | None = None  # ← Add this

# Step 2 — censors.py _semantic_search(): capture similarity per-id and pass it through
scores_by_id = {row.id: float(row.similarity) for row in rows}
# ... fetch censors ...
return [
    CensorMatch(
        id=c.id,
        trigger_pattern=c.trigger_pattern,
        action=c.action,
        reason=c.reason,
        domain=c.domain,
        score=scores_by_id.get(cid),  # ← populate from similarity
    )
    for cid in ids
    if (c := censors.get(cid)) is not None
]
```

---

## 🟡 P2 — Comment inaccuracy: censors do NOT use `hybrid_search()`

**File:** `nous/heart/heart.py`, lines 463–466 (comment)

The comment states:
> *"All Heart sub-searches use the same hybrid_search() formula (0.7\*vector + 0.3\*keyword), so scores are directly comparable across memory types."*

This is **not accurate for censors**. `censors._search()` branches to either `_semantic_search` (cosine similarity only, no keyword component) or `_keyword_search` (ILIKE only). Neither goes through `hybrid_search()`.

- `episodes.search()` → `hybrid_search()` ✅
- `facts.search()` → `hybrid_search()` ✅
- `procedures.search()` → `hybrid_search()` ✅
- `censors.search()` → custom vector OR keyword path ❌

Scores are broadly in the 0–1 range so cross-type ranking is still reasonable, but the comment overstates the comparability guarantee. Suggest updating it to be precise.

---

## 🟡 P2 — `None or 0.0` masks zero-score vs missing-score

**File:** `nous/heart/heart.py`, line 479

```python
original_score = getattr(item, "score", None) or 0.0
```

`or 0.0` treats both `None` (missing field) and `0.0` (valid but zero relevance) identically. Prefer the explicit check:

```python
raw = getattr(item, "score", None)
original_score = raw if raw is not None else 0.0
```

Functionally equivalent in this codebase, but clearly expresses intent and avoids surprising someone who adds a type with a legitimate 0.0 score.

---

## 🔵 P3 — Tests lack score-range assertions for censor type

**File:** `tests/test_heart.py`, line 346

Once the P1 fix lands, add a bounds assertion to guard against regression:

```python
for r in results:
    assert isinstance(r, RecallResult)
    assert r.type in ("episode", "fact", "procedure", "censor")
    assert r.score > 0, f"Expected score > 0 for type={r.type}, got {r.score}"
    assert r.score <= 1.0, f"Score out of expected range: {r.score}"
```

---

## ✅ What's correct in this PR

- The core insight is right: RRF positional scores `1/(k+rank)` discard the actual relevance signal. For episode, fact, and procedure results the new approach correctly preserves real hybrid scores.
- The merge + sort logic is clean and readable.
- The `getattr(..., None)` fallback is the right defensive pattern — it just needs `CensorMatch` to actually carry a score.

---

## 🚦 Verdict: **DO NOT MERGE** — 1 P1 must be fixed first

| # | Severity | Issue | Location |
|---|----------|-------|----------|
| 1 | 🔴 P1 | `CensorMatch` has no `.score` → censors get `score=0.0`, breaks `test_unified_recall` | `schemas.py:229`, `censors.py:358`, `heart.py:479` |
| 2 | 🟡 P2 | Comment overstates score comparability — censors bypass `hybrid_search()` | `heart.py:463–466` |
| 3 | 🟡 P2 | `None or 0.0` — prefer explicit `is not None` check | `heart.py:479` |
| 4 | 🔵 P3 | Tests need score-range assertions for censor type | `test_heart.py:346` |

The fix is small: add `score: float | None = None` to `CensorMatch` and thread the similarity value through `censors._semantic_search()`. Once done, re-review welcome.

— 🔍 CodeReviewer

Issue #68, opened 2026-02-27
Conversation history has no token management — risk of context window overflow


tfatykhov:
## Problem

The tiered context system (Spec 008) manages the system prompt well, but conversation history has zero token management.

### Current State
- `MAX_HISTORY_MESSAGES = 20` — hard cutoff by count, not tokens
- Full messages sent (including long bash outputs, tool results)
- No token counting before API calls
- No summarization of old messages
- No truncation of long tool outputs
- Conversation is in-memory only — lost on restart

### Risk
A 20-message conversation with bash tool outputs could easily exceed 50K+ input tokens. With `max_tokens` increased to 16384 for output, total per-turn cost could spike.

### What OpenClaw Does (Reference)
- Compacts context when hitting ~58% of window
- Summarizes old messages
- Tracks token usage per turn
- Truncates tool outputs

### Proposed Improvements (Phased)

**Phase 1 — Safety Net**
- Count input tokens before API call (tiktoken or char estimate)
- Truncate tool outputs over N chars (e.g. 4000)
- Log warning when conversation exceeds token threshold

**Phase 2 — Smart Management**
- Token-based history window (e.g. 30K token budget for history)
- Summarize old messages beyond the window
- Persist conversation to DB for restart recovery

**Phase 3 — Full Compaction**
- Automatic compaction when approaching context limit
- Rolling summary that preserves key decisions and context
- Integration with episode system (conversation → episode on compaction)

## Context
- System prompt: ~2-3K tokens (managed by Spec 008)
- Conversation: up to 50K+ tokens (UNMANAGED)
- Model: Sonnet 4.6 (200K context, 64K output)
- Current max_tokens being raised from 4096 to 16384

PR #69, opened 2026-02-27
feat: 008.1 Phase 1 — Tool Output Pruning + Token Estimation

tfatykhov:
## Spec 008.1 Phase 1 Implementation

Addresses #68 (Conversation history has no token management)

### New Files
- `nous/api/compaction.py` — `ConversationCompactor` (Layer 1 pruning) + `TokenEstimator` (calibrated estimation)
- `nous/api/models.py` — Extracted `Message`, `Conversation`, `ApiResponse` from runner.py to avoid circular imports
- `tests/test_compaction.py` — 30 test cases covering pruning, estimation, identification, edge cases

### Changes
- **runner.py**: Import from models.py, instantiate compactor in `__init__`, hook `prune_tool_results()` into both `_tool_loop()` and `stream_chat()` after tool results appended, calibrate estimator from API usage
- **config.py**: 9 new config fields with `NOUS_` env var aliases (6 Layer 1, 3 Layer 2 stubs)

### How It Works
- **Soft-trim**: Tool results > 4000 chars → keep first 1500 + last 1500 chars with marker
- **Hard-clear**: Tool results > 6 messages old → replace with placeholder
- **Protection**: Last 2 tool results always preserved (model just acted on these)
- **Calibration**: `TokenEstimator` starts at chars/4, improves via EMA from actual API `input_tokens`
- **Feature flag**: `NOUS_TOOL_PRUNING_ENABLED=true` (on by default)

### Backward Compatible
- `Message`, `Conversation`, `ApiResponse` re-exported from runner.py — all existing imports work
- Pruning only activates during tool loops when tool results exist
- No behavior change when `tool_pruning_enabled=false`

### Spec
Approved v5: `docs/implementation/008.1-conversation-compaction.md` (3/3 reviewer approval across 3 review rounds)

### DIFF
--- a/tests/test_compaction.py
+++ b/tests/test_compaction.py
@@ -1,7 +1,5 @@
 """Tests for Spec 008.1 Phase 1: Tool Output Pruning + Token Estimation."""
 
-import pytest
-
 from nous.api.compaction import ConversationCompactor, TokenEstimator
 from nous.config import Settings
 


tfatykhov:
## 🔍 CodeReviewer — PR #69 Review
### Spec 008.1 Phase 1: Tool Output Pruning + Token Estimation

**Files reviewed:** `compaction.py` (226 lines new), `models.py` (39 lines new), `runner.py` (~50 lines modified), `config.py` (~35 lines new), `test_compaction.py` (354 lines new, 28 tests)

---

### ✅ What Looks Good

- **Pruning logic is correct** — `prune_tool_results` properly identifies tool-result messages, protects the last N, ages correctly (distance from end), and applies soft-trim → hard-clear in the right order
- **Image content preservation** — `_has_image_content` correctly skips image blocks from both soft-trim and hard-clear paths
- **Hook placement** — Both `stream_chat()` (line ~755) and `_tool_loop()` (line ~899) insert pruning after `messages.append(tool_results)` and before the next API call — exactly right
- **TokenEstimator** — Clean EMA implementation with proper zero-guards on `calibrate()`
- **models.py extraction** — Clean separation. `Conversation` adds `summary` and `compaction_count` with defaults, preserving backward compatibility
- **Test coverage** — 28 tests covering positive/negative identification, soft-trim, hard-clear, protection zones, image skipping, disabled flag, edge cases (empty, all-protected), and TokenEstimator convergence. Solid.
- **Config fields** — Consistent with existing `validation_alias` pattern in Settings. Defaults match spec.
- **Phase 2 stubs** — `should_compact()` and `find_cut_point()` are implemented early with basic tests. Good for Phase 2 readiness.

---

### P2 (Should Fix)

#### 1. `stream_chat()` calibration is systematically biased during tool loops

```python
# Lines ~791-796 (stream_chat, after tool loop completes)
if self._compactor and total_usage.get("input_tokens"):
    input_chars = sum(len(str(m.get("content", ""))) for m in messages)
    self._compactor.estimator.calibrate(
        input_chars, total_usage["input_tokens"]
    )
```

`total_usage["input_tokens"]` is the **sum** across all API calls in the streaming tool loop. But `input_chars` is computed from the **final** message state. Since each API call sends progressively larger messages, `sum(all_calls_tokens) / final_chars` overestimates the true ratio.

Example with 2 tool iterations (true ratio 0.25):
- Call 1: 10K chars → 2,500 tokens
- Call 2: 15K chars → 3,750 tokens
- Calibration: 6,250 / 15,000 = 0.417 (should be ~0.25)

Contrast with `_tool_loop()` which correctly calibrates **per-call** (lines ~847-851).

**Suggestion:** Either (a) calibrate per-iteration inside the streaming tool loop (matching `_tool_loop`), or (b) track accumulated chars across iterations (sum of chars sent per call) rather than just the final state. The 20K safety margin absorbs this for now, but it creates a systematic upward bias in the estimator.

#### 2. No validation that `tool_soft_trim_head + tool_soft_trim_tail < tool_soft_trim_chars`

With misconfigured settings (e.g., head=3000, tail=3000, threshold=4000), the "trimmed" output would be **larger** than the original (6000 chars + marker text vs 4001 chars). Consider adding a `model_validator` in Settings:

```python
@model_validator(mode="after")
def _validate_trim_sizes(self) -> "Settings":
    if self.tool_soft_trim_head + self.tool_soft_trim_tail >= self.tool_soft_trim_chars:
        raise ValueError(
            "tool_soft_trim_head + tool_soft_trim_tail must be < tool_soft_trim_chars"
        )
    return self
```

#### 3. Backward compatibility relies on implicit re-export (fragile)

Multiple test files import from `nous.api.runner`:
```python
# test_runner.py, test_tool_loop.py, test_streaming.py
from nous.api.runner import ApiResponse, Conversation, Message
```

This works because `from nous.api.models import ApiResponse, Conversation, Message` in runner.py makes them available in runner's namespace. But it's **implicit** — adding `__all__` to runner.py or changing the import to `import nous.api.models` would silently break downstream imports.

**Suggestion:** Add explicit re-exports in runner.py:
```python
from nous.api.models import ApiResponse, Conversation, Message  # noqa: F401 — re-exported
```
Or add to `__all__`. The `# noqa` comment documents the intent.

---

### P3 (Nice to Have)

#### 1. `soft_trimmed` counter counts items, not messages

The `soft_trimmed` variable increments per tool-result **item** within a message, not per message. If one message has 3 tool results that all get trimmed, `soft_trimmed=3`. The log message says "soft-trimmed=3" which could be misread as 3 messages. Consider renaming to `soft_trimmed_items` or adding item vs message distinction in the log.

#### 2. Missing test for `should_compact()` with `compaction_enabled=False`

The `should_compact()` stub has a guard for `compaction_enabled`, but no test verifies it returns `False` when disabled. Minor since it's Phase 2 but good for completeness.

#### 3. `find_cut_point` test is conditional

```python
def test_snaps_to_user_boundary(self):
    ...
    cut = compactor.find_cut_point(messages, keep_recent_tokens=300)
    if cut > 0:
        assert messages[cut]["role"] == "user"
```

The `if cut > 0` means this test can silently pass without actually asserting anything (if the estimator thinks everything fits in 300 tokens). Consider using a smaller `keep_recent_tokens` value or larger messages to guarantee `cut > 0`.

---

### Summary

| Severity | Count | Status |
|----------|-------|--------|
| P1 (Critical) | 0 | — |
| P2 (Should Fix) | 3 | See above |
| P3 (Nice to Have) | 3 | See above |

No P1 issues found. The core pruning logic is correct and well-tested. The P2 items are worth addressing before merge — especially #1 (calibration bias) and #2 (config validation). P3 items are optional improvements.

🔍 *CodeReviewer*

PR #70, opened 2026-02-27
feat: 008.1 Phase 2 — History Compaction

tfatykhov:
## Spec 008.1 Phase 2 Implementation

Addresses #68 — Layer 2: History Compaction

### What It Does
When estimated input tokens exceed `NOUS_COMPACTION_THRESHOLD` (default 100K):
1. Find cut point (keep ~20K tokens recent, snap to user boundary)
2. Summarize old messages into structured checkpoint via background model
3. Replace old messages with [summary, ack] + recent window
4. If summary fails → fall back to truncation (no data loss)

### Changes
- **compaction.py** (+276 lines): Full `compact()`, `_summarize()`, `_validate_summary()`, `_serialize_for_summary()`, `ApiCaller` Protocol, checkpoint/update prompts
- **runner.py** (+49 lines): `model_override` on `_call_api`/`_build_api_payload`, compaction trigger in `stream_chat()` with per-session `asyncio.Lock` + double-check, lock cleanup on LRU eviction, `_format_messages()` returns all when compaction enabled, `_format_history_text()` summary-aware
- **test_compaction_phase2.py** (442 lines): 30+ tests

### Key Design Decisions
- Compaction trigger is **explicit** in `stream_chat()`, not hidden in formatting
- `_format_messages()` stays sync and pure
- Per-session `asyncio.Lock` with double-check pattern prevents concurrent compaction
- Structured checkpoint format with Conversation Dynamics section
- Iterative UPDATE prompt merges into existing summary (no photocopy-of-photocopy)
- Format validation (regex, case-insensitive) with honest acknowledgment that content validation is deferred
- Fallback to truncation on any failure

### Feature Flag
`NOUS_COMPACTION_ENABLED=false` — off by default. Enable with `NOUS_COMPACTION_ENABLED=true`.

### Spec
Approved v5: `docs/implementation/008.1-conversation-compaction.md` (3/3 reviewer approval)

### DIFF
--- a/nous/api/compaction.py
+++ b/nous/api/compaction.py
@@ -349,6 +349,10 @@ class ConversationCompactor:
                 "Compaction failed: %s - falling back to truncation", e
             )
             conversation.messages = conversation.messages[cut_point:]
+            conversation.summary = None
+            conversation.compaction_count = max(
+                0, conversation.compaction_count - 1
+            )
             return
 
         # Rebuild messages with summary prefix
--- a/tests/test_compaction.py
+++ b/tests/test_compaction.py
@@ -3,7 +3,6 @@
 from nous.api.compaction import ConversationCompactor, TokenEstimator
 from nous.config import Settings
 
-
 # ------------------------------------------------------------------
 # TokenEstimator Tests
 # ------------------------------------------------------------------


tfatykhov:
🔍 **CodeReviewer** — PR #70 Review (Spec 008.1 Phase 2: History Compaction)

## Summary

Clean, well-structured implementation that closely follows the approved spec. The core compaction logic, double-check locking, fallback to truncation, and validation are all solid. Tests are comprehensive. Found **1 P2** and **3 P3** issues.

---

## P2 (Should Fix)

### 1. Stale `conversation.summary` on iterative compaction failure

**File:** `nous/api/compaction.py`, `compact()` fallback path

When `compact()` fails during an **iterative** compaction (i.e., `conversation.summary` already exists from a prior successful compaction), the fallback does:

```python
conversation.messages = conversation.messages[cut_point:]
return
```

This truncates messages (removing the synthetic summary prefix at indices 0-1) but leaves `conversation.summary` set to the old value. On the **next** compaction attempt:

1. `conversation.summary` is not `None` → `serialize_start = 2`
2. But `messages[0:2]` are now **real messages**, not the synthetic prefix
3. Two real messages get silently skipped during serialization → **data loss in summary**

**Fix:** Clear the summary in the fallback path:

```python
except Exception as e:
    logger.error("Compaction failed: %s - falling back to truncation", e)
    conversation.messages = conversation.messages[cut_point:]
    conversation.summary = None  # Prevent stale serialize_start=2
    return
```

This loses the old summary context but ensures consistent state. The alternative (keeping stale summary) causes a worse cascading bug.

---

## P3 (Nice to Have)

### 1. Missing concurrency test

The spec lists `test_concurrency_lock` in the testing strategy but no concurrency test is present. A basic test verifying two concurrent `compact()` calls on the same session don't corrupt state would increase confidence in the lock pattern.

### 2. Deprecated `asyncio.get_event_loop().run_until_complete()`

Tests use `asyncio.get_event_loop().run_until_complete()` which is deprecated since Python 3.10. Consider `asyncio.run()` or `@pytest.mark.asyncio` with `pytest-asyncio` for cleaner async test patterns.

### 3. `TestFormatMessages` / `TestFormatHistoryText` test inline simulation

These tests simulate the logic inline rather than testing through `AgentRunner._format_messages()` / `_format_history_text()`. This is understandable given AgentRunner's complex dependencies, but means the tests wouldn't catch a divergence between the simulation and the actual method.

---

## Verification Checklist

| # | Check | Status |
|---|-------|--------|
| 1 | `compact()` guards: cut_point > 0, index alignment | ✅ Both `ValueError` raises present |
| 2 | Alternation fix with for/else pattern | ✅ `found` flag pattern correctly handles no-user-in-recent |
| 3 | Iterative update skips synthetic prefix (serialize_start=2) | ✅ Logic correct, but see P2 for failure path |
| 4 | `_validate_summary()` regex case-insensitive | ✅ Pre-compiled `_SECTION_PATTERNS` with `re.IGNORECASE` |
| 5 | `model_override` propagation | ✅ `_call_api` → `_build_api_payload` → `model_override or self._settings.model` |
| 6 | Double-check pattern under lock | ✅ Re-fetches messages + re-checks `should_compact` inside lock |
| 7 | Lock cleanup on LRU eviction | ✅ `self._compaction_locks.pop(evicted_id, None)` |
| 8 | Fallback to truncation on API failure AND bad summary | ✅ Both caught by try/except (bad summary raises ValueError) |
| 9 | `_format_messages()` returns all when compaction enabled | ✅ Bypasses `MAX_HISTORY_MESSAGES` cap |
| 10 | `_format_history_text()` includes summary | ✅ Prepends `[Previous context summary]` block |
| 11 | Tests comprehensive and correct | ✅ Good coverage; see P3 notes |

**Config/model fields** (`background_model`, `summary`, `compaction_count`, `compaction_enabled`, `compaction_threshold`, `keep_recent_tokens`): Verified all exist from Phase 1 — no missing dependencies.

---

## Verdict

**1 P2 to fix**, then LGTM. The stale summary issue is a real bug that would cause silent data loss on an already-degraded path. Everything else is solid work that matches the spec closely.

tfatykhov:
🔍 **CodeReviewer** — PR #70 Re-review (commit 995d24d)

## P2 Fix Verified ✅

The fix correctly addresses the stale summary issue:

```python
conversation.messages = conversation.messages[cut_point:]
conversation.summary = None                          # ← NEW: prevents serialize_start=2 on next compaction
conversation.compaction_count = max(                  # ← NEW: consistent state
    0, conversation.compaction_count - 1
)
```

The `compaction_count` decrement is a nice addition I hadn't considered — it keeps the count consistent with reality (a failed compaction shouldn't count). The `max(0, ...)` guard prevents underflow.

## Minor Note

The existing `test_fallback_on_api_failure` and `test_fallback_on_bad_summary` tests assert `conv.summary is None` after failure, but both start with a fresh conversation (`summary=None`), so they pass with or without the fix. A test that sets `conv.summary = "old"` + `conv.compaction_count = 1` before triggering a failure would directly verify the iterative failure path. Not blocking — the fix is obviously correct.

## Lint fix

Import sorting fix in `test_compaction.py` — clean. ✅

## Verdict

**No remaining issues. LGTM — ready to merge.** 🚀

tfatykhov:
🔍 **CodeReviewer** (Re-review)

## P2 Fix Verification: Stale `conversation.summary` on Compaction Fallback

**Commit:** 995d24d
**Status:** ✅ Fix is correct and complete

### What was fixed

The `except Exception` handler in `compact()` (fallback-to-truncation path) now:

1. **`conversation.summary = None`** — Clears the stale summary so the next compaction cycle uses `CHECKPOINT_SYSTEM_PROMPT` (fresh summary) instead of `UPDATE_SYSTEM_PROMPT` (incremental update against a now-invalid base). Also correctly resets `serialize_start` to 0 on the next call since the `if conversation.summary` guard will be False.

2. **`conversation.compaction_count = max(0, conversation.compaction_count - 1)`** — Accurately reflects that the compaction did not succeed. The `max(0, ...)` guard prevents underflow to negative values on first-compaction failures.

### Verification

Traced the data flow:
- After fallback: `summary=None`, messages truncated from front, count decremented
- Next `compact()` call: `serialize_start=0` (all old messages included), `_summarize()` receives `existing_summary=None` → uses fresh `CHECKPOINT_SYSTEM_PROMPT` ✅
- Success path still works correctly: sets `summary=checkpoint_text`, increments count, cleans `turn_contexts` ✅

### Minor observation (P3, not blocking)

The fallback path does not trim `turn_contexts` (unlike the success path which does `conversation.turn_contexts = conversation.turn_contexts[-keep_contexts:]`). Stale `turn_contexts` entries referencing removed turns waste a small amount of memory but are harmless since they only feed token estimation calibration. Not worth blocking the PR.

### Verdict

**No P1 or P2 issues remaining.** The fix correctly addresses the stale summary problem. Ready to merge.

PR #79, opened 2026-02-28
fix: 008.3 episode summary backfill + active flag lifecycle

tfatykhov:
Implements Spec 008.3.

_update_summary(): backfill title, summary, lessons_learned from structured JSON + refresh embedding.
_end(): set active=false on close.
4 new tests.

### DIFF
--- a/docs/implementation/008.4-summary-quality.md
+++ b/docs/implementation/008.4-summary-quality.md
@@ -1,0 +1,289 @@
+# Spec 008.4: Episode Summary Quality Improvements
+
+**Status:** Draft
+**Depends on:** 008.3 (Backfill fix), 006 (Event Bus)
+**Related:** F007 (Metrics & Growth), F010 (Memory Improvements)
+
+## Problem
+
+The episode summarizer produces structured summaries that work but have systematic quality gaps. With 008.3 fixing the plumbing (backfilling text columns), the next step is improving *what* gets summarized and *how well*.
+
+### Current Quality Issues
+
+#### 1. Outcome Classification is Weak
+Almost every episode returns `"outcome": "partial"` regardless of actual completion:
+
+| Episode | Actual Outcome | Summarizer Says |
+|---------|---------------|-----------------|
+| Weather check → answered | Resolved | `partial` |
+| Research doc written and committed | Resolved | `partial` |
+| A2A communication established | Resolved | `partial` |
+
+**Root cause:** The summarizer prompt has no definition of what constitutes "resolved" vs "partial" vs "informational". The LLM defaults to `partial` when uncertain.
+
+#### 2. No Decision Awareness
+The summarizer only sees the transcript. It doesn't know about:
+- Brain decisions made during the episode (confidence, reasoning, outcome)
+- Deliberation traces (thinking blocks)
+- Tools used and their results
+
+A session where Nous made 3 architecture decisions appears identical to a casual weather check from the summarizer's perspective — just text.
+
+#### 3. Lessons Are "What Happened" Not "What Was Learned"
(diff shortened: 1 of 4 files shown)


tfatykhov:
🔍 **CodeReviewer** — PR #79 Review

## Summary

The production code changes (`_update_summary` backfill + `_end` active=false) are **correct and well-structured**. However, the new tests have two bugs that will cause failures.

---

## P1 — Critical (Must Fix)

### 1. Tests reference `detail.active` but `EpisodeDetail` has no `active` field

**Files:** `tests/test_episodes.py` lines 284, 287, 308 | `nous/heart/schemas.py`

`test_end_sets_active_false` and `test_end_with_lessons_and_active_flag` both assert on `detail.active`, but:
- `EpisodeDetail` (schemas.py L38-61) does **not** have an `active` field
- `_to_detail()` (episodes.py L497-520) does **not** pass `active` to the constructor

These tests will raise `AttributeError: "EpisodeDetail" object has no attribute "active"`.

**Fix:** Add `active: bool = True` to `EpisodeDetail` in schemas.py, and add `active=episode.active,` to `_to_detail()` in episodes.py.

### 2. `test_update_summary_partial_data` asserts `lessons_learned is None` but it will be `[]`

**File:** `tests/test_episodes.py` line 272

```python
assert updated.lessons_learned is None  # Not set — no "key_points"
```

`_to_detail()` line 514 converts `None` → `[]`:
```python
lessons_learned=episode.lessons_learned or [],
```

And `EpisodeDetail.lessons_learned` is typed as `list[str]` (not optional). So the actual value will be `[]`, not `None`.

**Fix:** Change to `assert updated.lessons_learned == []`

---

## P2 — Should Fix

### 3. Semantic behavior change needs documentation

`list_recent()` (L332) and `search_recent_by_embedding()` (L468) both filter `active = true`. With this PR, **ended episodes will no longer appear in these queries**. This is almost certainly intended, but should be called out in the PR description since it changes query behavior for any caller that previously relied on finding ended-but-active episodes.

### 4. No test for embedding refresh in `_update_summary`

The new embedding refresh logic (L152-158) is only exercised if `self.embeddings` is not None. The test fixtures use mock embeddings, so this path should execute, but there's no assertion that `embedding` was actually updated after `update_summary`. Consider adding:
```python
assert updated.embedding is not None  # or check it changed
```
(Minor — the happy path is implicitly covered if the mock doesn't raise.)

---

## P3 — Nice to Have

### 5. No test for non-dict summary

The `isinstance(summary, dict)` guard (L143) is good defensive coding. A test passing a string as `summary` would lock down the behavior: `structured_summary` gets set but no backfill occurs.

### 6. No test for `_update_summary` with nonexistent episode

If `episode_id` doesn't match, `_update_summary` silently does nothing. A test would document this contract.

---

## Correctness Checklist

| Question | Answer |
|----------|--------|
| Does `_update_summary()` backfill title, summary, lessons_learned? | ✅ Yes, correctly using `summary.get()` with truthy checks |
| Does `_end()` set `active=False`? | ✅ Yes (L200) |
| Are the 4 new tests correct? | ❌ Two will fail (P1 #1 and #2 above) |
| Edge case: structured_summary not a dict? | ✅ Guarded by `isinstance(summary, dict)` |
| Edge case: embeddings is None? | ✅ Guarded by `if self.embeddings:` |
| Existing tests broken by active=false? | ✅ No — existing `test_end_episode` doesn't check `active`, and `test_list_recent` uses non-ended episodes |

---

**Verdict:** Fix the two P1 issues (schema missing `active` + wrong assertion), then this is good to merge. The production code is clean and correct.

PR #11, opened 2026-02-23
feat: add runtime layer — REST API, MCP server, agent runner (005)

tfatykhov:
## Summary

- **AgentRunner** with conversation management, LRU eviction (MAX_CONVERSATIONS=100), and reflection on session end
- **11 REST endpoints** via Starlette (health, status, chat, decisions, episodes, facts, frames, censors, calibration)
- **5 MCP tools** (nous_ask, nous_decide, nous_teach, nous_recall, nous_status) via `StreamableHTTPServerTransport`
- **Starlette lifespan** for clean async startup/shutdown of Brain, Heart, CognitiveLayer
- **Config additions**: host, port, anthropic_api_key, agent identity, MCP toggle, LLM model/max_tokens
- **Brain.list_decisions()** with pagination + tag fetch, **CognitiveLayer.list_frames()** public delegation
- **Dockerfile** + docker-compose `nous` service
- **228/228 tests passing** including 33 new tests (10 runner, 15 REST, 8 MCP)

## Key fixes applied during development

- Replaced sync `TestClient` with async `httpx.AsyncClient` + `ASGITransport` (event loop mismatch)
- Rewrote MCP transport for `mcp` library 1.26.0 API (`StreamableHTTPServerTransport`)
- Fixed conftest SAVEPOINT handler (removed sync listener on async connection)
- Fixed `test_guardrail_jsonb` for CEL condition format
- Fixed `calibration.reviewed_count` → `reviewed_decisions` attribute

## Test plan

- [x] Full test suite: 228/228 passing, 0 regressions
- [x] Runner tests: conversation lifecycle, LRU eviction, API error handling
- [x] REST tests: all 11 endpoints with async httpx client
- [x] MCP tests: all 5 tool handlers
- [ ] Manual smoke test with `docker compose up`

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #114, opened 2026-03-04
F014: Frame Reasoning Scaffolds

tfatykhov:
## Summary
Add structured reasoning templates (semi-formal scaffolds) to each of Nous cognitive frames.

## Motivation
Paper #11 showed structured reasoning templates improve accuracy by up to 30%. Each frame should supply a reasoning scaffold — not telling the agent what to conclude, but what steps to take.

## Phases
- **Phase 1** — Static Scaffolds (~3h): 7 frame-specific templates + complexity gate
- **Phase 2** — Memory Quality Gates (~2h): Pre-commit checks on learn_fact and record_decision
- **Phase 3** — DB-Driven Templates (~4h): reasoning_templates table, domain-specific variants

## CSTP Extension
- Template serving to connected agents
- Reasoning audit trails
- Domain template marketplace

## Research Basis
Papers #10, #11 from the LLM agent memory/cognition collection

## Files
- docs/features/F014-reasoning-scaffolds.md (+297 lines)

### DIFF
--- a/docs/features/F014-reasoning-scaffolds.md
+++ b/docs/features/F014-reasoning-scaffolds.md
@@ -1,6 +1,6 @@
 # F014 — Frame Reasoning Scaffolds
 
-**Status:** Draft (v3 — revised after architecture review)  
+**Status:** Draft (v4 — approved with required revisions)  
 **Author:** Nous (with Tim)  
 **Created:** 2026-03-05  
 **Revised:** 2026-03-05  
@@ -48,10 +48,24 @@ Why pilot Decision first:
 
 ### Success Criteria (2-week measurement window)
 
+**Success (expand to next frame):**
 - Average reason count per decision increases (baseline: current avg)
 - Confidence calibration improves (fewer 0.95+ decisions that get marked as failures)
 - No increase in scaffold-related token costs > 15% per decision turn
 - Qualitative: decision descriptions become more structured without feeling formulaic
+
+**Failure (kill or redesign):**
+- Token cost increase > 25% with no measurable quality improvement
+- Scaffold steps are parroted mechanically without genuine reasoning (cargo cult)
+- Decision recording frequency drops (scaffold overhead discourages recording)
+- Tim reports the agent feels formulaic or robotic in decision conversations
+
+**Pre-pilot baseline task:** Before enabling scaffolds, run a query against the `decisions` table to capture:
+- Average reason count per decision (last 30 days)
+- Confidence distribution histogram (buckets: 0-0.5, 0.5-0.7, 0.7-0.85, 0.85-0.95, 0.95-1.0)
+- Decision-to-failure rate by confidence bucket
+- Average token usage per decision-frame turn
+Store as a fact in Heart for comparison at pilot end.
 
 ### File Structure
 
@@ -92,6 +106,12 @@ DECISION_SCAFFOLD_SHORT = (
 )
 
 
+import os
+
+# Kill switch — set NOUS_REASONING_SCAFFOLDS=false to disable all scaffolds
+SCAFFOLDS_ENABLED = os.getenv("NOUS_REASONING_SCAFFOLDS", "true").lower() in ("true", "1", "yes")
+
+
 def get_scaffold(frame_id: str, turn_in_frame: int, message_length: int) -> str | None:
     """Return the appropriate scaffold for a frame and turn.
 
@@ -103,6 +123,10 @@ def get_scaffold(frame_id: str, turn_in_frame: int, message_length: int) -> str 
     Returns:
         Scaffold string, or None if no scaffold applies.
     """
+    # Kill switch
+    if not SCAFFOLDS_ENABLED:
+        return None
+
     # Phase 1: Only Decision frame gets a scaffold
     if frame_id != "decision":
         return None
@@ -126,8 +150,8 @@ def _get_frame_instructions(self, turn_context: TurnContext) -> str:
     # Check for reasoning scaffold (replaces tool nudges for scaffolded frames)
     scaffold = get_scaffold(
         frame_id=frame_id,
-        turn_in_frame=turn_context.turn_in_frame,  # requires tracking
-        message_length=len(turn_context.user_message or ""),
+        turn_in_frame=turn_context.turn_in_frame,  # tracked in runner.py turn loop
+        message_length=turn_context.message_length,  # set by runner.py from user message
     )
     if scaffold:
         return scaffold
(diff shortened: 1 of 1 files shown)


tfatykhov:
## 🏗️ Architecture Review — F014: Frame Reasoning Scaffolds

### Overall Assessment: ✅ Sound design, a few structural concerns to address

The phased approach is well thought out and the spec demonstrates strong understanding of the existing architecture. The core idea — injecting structured reasoning scaffolds per frame — is architecturally clean and aligns with Nous's frame-centric design. Below are my detailed findings.

---

### 1. Architectural Fit ✅

**Good alignment with existing patterns:**
- Scaffolds extend `_get_frame_instructions()` in `runner.py`, which is already the designated extension point for frame-specific prompt content
- The phased approach (static strings → brain integration → DB-driven) follows Nous's iterative pattern
- Research subtask scaffold correctly uses `build_subtask_prefix()` instead of the frame system — proper separation
- Config flag `NOUS_REASONING_SCAFFOLDS=true` follows existing feature flag patterns

**`FrameType` enum alignment concern (minor):**
The spec references 7 frames including `initiation`, but `schemas.py::FrameType` only defines 6 (task, question, decision, creative, conversation, debug). The `initiation` frame exists only in `FRAME_TOOLS` in `runner.py` (line 45), not in the `FrameType` enum. The spec correctly skips initiation scaffolds, but Phase 3's DB schema should align with however initiation is formalized.

---

### 2. Coupling & Separation of Concerns ⚠️

**Critical: Two separate frame content injection paths**

There are currently **two independent paths** for injecting frame-related content into the system prompt:

1. **`ContextEngine._format_frame()`** in `context.py` — produces frame description + questions, goes through `_truncate_to_budget()` with `budget.frame` (500 tokens)
2. **`Runner._get_frame_instructions()`** in `runner.py` — appends tool instructions AFTER the budget-controlled system prompt, **bypassing the budget system entirely**

The spec proposes adding scaffolds (150-300 tokens) to path #2. This means **scaffold tokens are invisible to the budget system**. Today this is manageable because tool instructions are ~80-120 tokens, but after adding scaffolds, path #2 could inject 300-500 tokens completely outside budget control.

**Recommendation:** Either:
- (a) Account for scaffolds in the budget by increasing `budget.frame` and moving scaffold injection into `ContextEngine`, or
- (b) Document this as intentional (scaffolds are "fixed overhead" not subject to dynamic budgeting) and add a comment in `_build_system_prompt()` noting the unbudgeted cost

Option (b) is pragmatically fine for Phase 1 since scaffold tokens are bounded and predictable, but it should be explicit.

---

### 3. Phased Approach & Tech Debt Analysis ✅

**Phase 1 → Phase 3 migration path is clean:**
- Phase 1 scaffolds are simple string constants — easy to extract into DB rows
- The `steps` JSONB schema in Phase 3 maps directly to the Phase 1 step format (NAME: instruction)
- `frame_turn_count` on `TurnContext` is needed regardless — no throwaway work

**One tech debt concern: scaffold string location**

The spec says "Create shared scaffold definitions (used by both `_get_frame_instructions()` and `build_subtask_prefix()`)" but doesn't specify *where*. If scaffolds land as constants in `runner.py`, they'll need extraction during Phase 3. 

**Recommendation:** Put scaffold string constants in a dedicated module (e.g., `nous/cognitive/scaffolds.py`) from Phase 1. Both `runner.py` and `tools.py` import from there. When Phase 3 adds DB-driven templates, this module becomes the fallback/default layer. Zero rewrite needed.

---

### 4. Token Budget Analysis ⚠️

**The compression strategy is good but the budget math needs checking:**

Current `budget.frame` is **500 tokens** (all frames). The `_format_frame()` output is typically 50-100 tokens (frame name + description + questions). This leaves 400-450 tokens of headroom *within the budget system*.

But as noted in §2, scaffolds are injected via `_get_frame_instructions()` which is **outside** the budget. So the real question is total system prompt growth:

| Component | Current | With Scaffold (Turn 1) | With Scaffold (Turn 2+) |
|-----------|---------|----------------------|------------------------|
| Frame (budgeted) | ~80 tok | ~80 tok | ~80 tok |
| Tool instructions (unbudgeted) | ~100 tok | ~100 tok | ~100 tok |
| Scaffold (unbudgeted) | 0 tok | ~250 tok | ~40 tok |
| **Total frame-related** | **~180 tok** | **~430 tok** | **~220 tok** |

The Turn 1 cost (+250 tokens) is a ~138% increase in frame-related prompt overhead. For the `decision` frame with a 12K total budget, this is ~2% — acceptable. For `conversation` (3K budget, no scaffold) — N/A. For `question` (6K budget) — ~4% — fine.

**The compression strategy (65% reduction by turn 2) is the right call.** Multi-turn cost stays manageable.

**One gap:** The spec doesn't address what happens when `budget.frame` truncation in `ContextEngine` cuts the frame description short. If truncation removes `questions_to_ask`, the scaffold might reference reasoning steps that conflict with truncated frame guidance. This is unlikely with current 500-token frame budgets but worth a comment.

---

### 5. Scaffold ↔ Deliberation ↔ run_python Interaction ✅

**Well-defined boundaries:**

- **Scaffolds → Deliberation:** The spec correctly positions Phase 2 enhancements as extensions to the existing `_should_deliberate()` gate (which already checks `_DELIBERATION_FRAMES = {"decision", "debug"}`). The scaffold doesn't create a parallel deliberation system — it enriches the existing one.

- **Scaffolds → run_python:** The spec explicitly states scaffolds influence *what* to compute in run_python, not *how* the sandbox executes. The `run_python` sandbox (in `tools.py`) has its own `learn_fact` wrapper with write caps — scaffolds don't bypass that. Clean boundary.

- **Scaffolds → record_decision:** The existing `_check_safety_net()` in runner.py already logs warnings when decision frames don't call `record_decision`. Scaffolds reinforce this at the prompt level (step 7: DECIDE → record_decision). Good layered defense.

**One suggestion:** Phase 2 proposes "check if scaffold reasoning steps were followed" in `_should_deliberate()`. Be careful here — `_should_deliberate()` currently returns a simple bool based on frame_id. Adding scaffold-step verification would change its contract from "should we deliberate?" to "is the deliberation well-formed?". Consider a separate `validate_deliberation()` method to keep `should_deliberate` clean.

---

### 6. Phase 3 Schema Review ⚠️

**The `reasoning_templates` schema is reasonable, with two concerns:**

1. **No `compressed_form` column.** Turn 2+ compression currently relies on generating a 1-line summary from step names. If Phase 3 allows custom templates with long step names, auto-compression might produce oversized reminders. Add an optional `compressed_form TEXT` column so template authors can specify their own compressed version.

2. **Domain selection mechanism undefined.** The spec says "domain-specific template selection based on intent signals" but doesn't specify what intent signals are or where they come from. Currently, `FrameSelection` has `default_category` and `default_stakes` — are these the intent signals? Clarifying this now prevents Phase 3 from requiring a new classification layer.

---

### 7. Open Questions — My Input

> 1. Should scaffolds be visible in the agent's response, or purely internal?

**Purely internal (thinking block).** Visible scaffolds would make responses feel robotic. The value is in structured *thinking*, not structured *output*.

> 2. Template inheritance — should subtasks inherit the parent's scaffold?

**Own scaffold based on frame_type.** Parent scaffold context may not apply to a child subtask in a different frame. The current `build_subtask_prefix()` already takes `frame_type` independently — keep that pattern.

> 3. Should Phase 2 fact quality checks become blocking?

**Advisory for at least 2 release cycles**, then evaluate based on false-positive rate. Blocking without tuning will cause frustrating fact storage failures.

> 4. How should scaffold compression handle frame switches?

**Reset turn counter on frame change** (as the spec already proposes). If the user switches from conversation → decision → conversation → decision, the second decision entry gets the full scaffold again. This is correct — context may have shifted.

---

### Summary

| Area | Rating | Key Finding |
|------|--------|-------------|
| Architectural fit | ✅ Good | Clean extension of existing patterns |
| Coupling | ⚠️ Minor | Two frame injection paths; scaffolds bypass budget system |
| Phase 1→3 path | ✅ Good | No throwaway work; clean migration |
| Token budget | ⚠️ Minor | Compression strategy is sound; document unbudgeted overhead |
| Component interaction | ✅ Good | Well-defined boundaries between scaffolds, deliberation, run_python |
| Phase 3 schema | ⚠️ Minor | Add compressed_form column; clarify domain selection signals |

**Verdict:** Ready to proceed to implementation with the above items addressed (or documented as intentional). No blocking architectural issues.

---
*🏗️ ArchitectureReviewer*

PR #126, opened 2026-03-08
F023: Memory Admission Control — A-MAC-inspired fact scoring gate

tfatykhov:
## Summary

Adds spec for F023 — Memory Admission Control, an A-MAC-inspired 5-factor scoring gate that evaluates every candidate fact **before** it enters long-term storage.

## Problem
Currently `learn_fact()` stores everything the LLM says to store. There is zero defense against:
- Hallucinated facts entering long-term memory
- Near-redundant facts that pass the 0.95 dedup threshold
- Low-utility transient observations diluting recall quality

## 5-Factor Scoring
- **Grounding** — ROUGE-L against recent conversation turns (anti-hallucination)
- **Novelty** — Cosine similarity, continuous range (not binary dedup)
- **Type Prior** — Category-based weights (preferences=0.95, tools=0.60)
- **Recency** — Exponential decay, 24h half-life
- **Utility** — LLM assessment (Phase 2 only)

## 4 Phases
1. Rule-based scoring (~250 LOC)
2. LLM utility factor (~60 LOC)
3. Calibration feedback loop via F017 (~150 LOC)
4. Contradiction-aware admission via F022 (~80 LOC)

Shadow mode rollout: scores without rejecting for first week.

### DIFF
--- a/docs/features/F023-memory-admission-control.md
+++ b/docs/features/F023-memory-admission-control.md
@@ -1,0 +1,473 @@
+# F023 — Memory Admission Control
+
+> **Status:** Planned
+> **Priority:** P1
+> **Author:** Nous (spec), Tim (requirements)
+> **Created:** 2026-03-08
+> **Depends on:** F002 (Heart Module — `FactManager.learn()`), F017 (Context Quality Gate — usage tracking)
+> **Research:** 016 — Agent Memory Synthesis (Gap G1: Memory Evolution), A-MAC paper (Mar 2026, Workday AI)
+> **Inspired by:** A-MAC (Adaptive Memory Admission Control) — 5-factor scoring for memory admission
+
+---
+
+## Problem Statement
+
+Nous stores every fact the LLM says to store. There is no quality gate on memory **admission**.
+
+**What exists today:**
+- `learn_fact()` is called by the LLM whenever it decides something is worth remembering
+- Near-duplicate detection (cosine > 0.95) prevents exact repeats
+- Contradiction detection (cosine 0.85–0.95) emits warnings but **does not block storage**
+- Subject-based supersession replaces old facts about the same subject
+
+**What's missing:**
+- **No hallucination check** — if the LLM fabricates a fact, it enters long-term storage unchallenged. There is zero verification that the fact is grounded in actual conversation content.
+- **No utility assessment** — a transient observation ("the API returned a 200") is stored with the same weight as a durable preference ("Tim prefers Celsius").
+- **No novelty gate beyond dedup** — a fact that's 90% similar to an existing one passes dedup (which requires >0.95) but adds almost nothing new.
+- **No type-based admission policy** — all categories (preference, technical, person, tool, concept, rule) are treated identically at admission time.
+
+**Result:** Memory grows without curation. Over time, low-value facts dilute recall quality. The context quality gate (F017) tries to filter at retrieval time, but the damage is done — noise is already in the database competing for embedding similarity scores.
+
+**The principle:** It's cheaper to prevent a bad memory than to filter it out later. Admission control is the first line of defense.
+
+---
+
+## Research Backing
+
+### A-MAC (March 4, 2026 — Workday AI)
+
+A-MAC proposes 5-factor composite scoring for memory admission:
+
+| Factor | What it measures | Method | Latency |
+|--------|-----------------|--------|---------|
+| **Utility (𝒰)** | "Will this be useful in future?" | Single LLM call, temp=0, cached | ~2580ms |
+| **Confidence (𝒞)** | "Is this grounded in conversation?" | ROUGE-L against recent turns | ~18ms |
+| **Novelty (𝒩)** | "Do we already know this?" | Cosine similarity via Sentence-BERT | ~32ms |
+| **Recency (ℛ)** | "How fresh is the source?" | Exponential decay, λ=0.01/hr | <1ms |
+| **Type Prior (𝒯)** | "What kind of information is this?" | Rule-based pattern matching | ~14ms |
+
+**Key findings:**
+- **Type Prior is the single most impactful factor** — removing it drops F1 by 0.107
+- Composite score with threshold θ*=0.55 achieves F1=0.583, precision=0.417, recall=0.972
+- 31% faster than A-MEM (2644ms vs 3831ms)
+- The LLM utility call is the bottleneck; all other factors are rule-based and fast
+
+### Relevance to Nous
+
+| A-MAC Factor | Nous Equivalent | Gap |
+|-------------|----------------|-----|
+| Utility (𝒰) | F017 usage tracking (post-hoc) | No pre-admission utility assessment |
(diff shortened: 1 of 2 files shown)


tfatykhov (commented):
🔍 **CodeReviewer** — Spec Review: F023 Memory Admission Control

## Summary

Well-researched spec with strong theoretical backing (A-MAC paper). The 5-factor scoring architecture is sound, the phase rollout plan is excellent, and the complementary relationship with F017 (admission vs retrieval filtering) is a good architectural choice.

However, there are **2 P1 critical issues** that would make the system non-functional as specified, **4 P2 issues** that need resolution before implementation, and **2 P3 suggestions**.

---

## P1 — Critical Issues

### P1-1: `confidence=1.0` bypass makes admission a no-op by default

The spec states (Bypass conditions section):
> `input.confidence == 1.0` (explicit override by LLM)

But `FactInput.confidence` defaults to `1.0` in `nous/heart/schemas.py`:
```python
confidence: float = Field(default=1.0, ge=0.0, le=1.0)
```

And the `learn_fact` tool also defaults to `confidence=1.0`:
```python
async def learn_fact(content, ..., confidence: float = 1.0, ...):
```

**This means every fact learned via the standard tool call bypasses admission scoring by default.** The LLM almost never explicitly sets confidence — it just calls `learn_fact(content=..., category=...)`. The entire admission system would be a no-op.

**Fix options:**
1. Change the bypass to require an explicit `override=True` flag (not tied to confidence)
2. Change the default confidence to something lower (e.g., 0.8) — but this has ripple effects
3. Only bypass when confidence is explicitly passed AND equals 1.0 (requires tracking whether it was explicitly set vs defaulted)

Option 1 is cleanest — add an `admission_override: bool = False` parameter to `learn_fact` and `FactInput`.

### P1-2: FactManager has no access to conversation messages for grounding check

`FactManager.__init__` takes `(db, embeddings, agent_id)`. It has no reference to `WorkingMemoryManager`, `ConversationState`, or any source of recent conversation turns.

The grounding check needs the last N conversation messages, but the spec says only _"Uses the last N messages from working memory / episode transcript"_ without specifying how `FactManager` gets them.

The `_learn()` call chain: `learn_fact()` (tools.py) → `heart.learn()` → `FactManager.learn()` → `FactManager._learn()`. None of these currently pass conversation context.

**Fix — the spec must specify the plumbing:**
- Option A: Add `recent_messages: list[str] | None = None` parameter to `learn()` / `_learn()`, and have `tools.py` fetch them from `ConversationState` before calling
- Option B: Give `FactManager` a reference to `WorkingMemoryManager` at construction time
- Option C: Have `_compute_admission_score()` query `ConversationState` from the database directly using `self.db`

Option C is simplest (no API changes to callers) but couples FactManager to conversation storage.

---

## P2 — Should Fix

### P2-1: ROUGE-L grounding blind spot for tool-sourced facts

The spec acknowledges this in Open Questions #3 and #6 but offers no solution. This is not an edge case — tool output is a **primary** source of facts in Nous. `web_fetch`, `web_search`, `run_python` results all feed into `learn_fact`.

If the grounding window only checks user/assistant messages (not tool call results), a large fraction of legitimate facts will get low grounding scores. The spec needs to specify:
- Which message types are included (user, assistant, tool results?)
- How tool output is accessed (it is in the conversation message array as tool_result messages)
- Whether this changes the grounding window size considerations

This should be resolved in the spec, not deferred to Open Questions.

### P2-2: `FactDetail` schema mismatch

The spec says `_rejected_detail()` returns a `FactDetail` with `admitted=False`, but `FactDetail` has no `admitted` field. Current fields: `id`, `agent_id`, `content`, `category`, `subject`, `confidence`, `source`, `learned_at`, `last_confirmed`, `confirmation_count`, `superseded_by`, `contradiction_of`, `active`, `tags`, `created_at`, `contradiction_warning`.

The spec claims _"No schema changes for Phase 1"_ but this requires adding at minimum `admitted: bool` and possibly `admission_score: float` and `rejection_reason: str` to `FactDetail`. These are Pydantic schema changes (not DB schema), but they need to be specified.

### P2-3: `category="rule"` bypass is risky

The spec bypasses admission for `category="rule"` because _"rules are always explicit declarations."_ But the LLM assigns categories — a hallucinated fact could be categorized as "rule" and bypass all checks. This creates a second bypass loophole alongside the confidence=1.0 issue.

The Type Prior table already gives rules 0.90, which is generous. Let rules go through scoring with their high prior rather than bypassing entirely. If a rule is hallucinated, grounding will catch it.

### P2-4: Recency factor adds minimal signal in Phase 1

With a 24h half-life, most facts learned during active conversation (age ~0) score ~1.0. Even 6-hour-old context scores 0.84. The factor barely differentiates in practice because the vast majority of `learn_fact` calls happen during live conversation.

The only meaningful case is F016 pre-prune extraction, where facts come from tool output that could be minutes to hours old. Consider:
- Whether 20% weight for a factor that is almost always ~1.0 is dead weight
- Whether `source_timestamp` is even reliably available (the spec assumes it but `FactInput` has no timestamp field — it has `source_episode_id` which could be used to derive age, but that is not addressed)

---

## P3 — Nice to Have

### P3-1: Type Prior table has dead entry for "rule"

Since rules bypass admission entirely (per the bypass conditions), the `TYPE_PRIORS["rule"] = 0.90` entry is dead code. Either remove the bypass (recommended, see P2-3) or remove the table entry. Having both is confusing.

### P3-2: INDEX.md ordering

The PR inserts F023 between F022 and F021, resulting in the order: F022, F023, F021. Numerical order would be F021, F022, F023. This is a pre-existing issue (F021 was already out of order) but the PR perpetuates it.

---

## What Works Well

1. **Problem statement** — Clearly articulated with concrete examples. The "prevent vs filter" principle is sound.
2. **Research backing** — A-MAC mapping to Nous gaps is thorough and honest.
3. **Phase rollout** — Shadow → soft → full mode is excellent risk management.
4. **Continuous novelty** — Extending binary dedup to continuous scoring is a meaningful improvement.
5. **Rejection UX** — Feeding rejection info back to the LLM is well-designed.
6. **Scope control** — "What We are NOT Building" section shows disciplined scoping.
7. **Metrics** — Comprehensive observability plan.
8. **F017 complementarity** — Using F017 usage data for Phase 3 calibration is an elegant feedback loop.

---

## Verdict

**DO NOT implement until P1 issues are resolved.** The confidence=1.0 bypass (P1-1) makes the entire system a no-op, and the missing conversation access plumbing (P1-2) means the highest-weighted factor (grounding) cannot be implemented as specified.

After P1 fixes, this is a strong spec that should produce a valuable feature. The P2 issues are design refinements that can be addressed in the spec update.

**Issue count: 2 P1 | 4 P2 | 2 P3**

tfatykhov:
Closing — F023 spec is already on main (docs/features/F023-memory-admission-control.md) committed directly as 22a55e7. The branch is CONFLICTING because the main version diverged. The F023 spec is already captured in the repo; the Admission Control Dashboard (F021.1) also partially implements these ideas. No action needed.

PR #140, opened 2026-03-16
fix(F022): auto-inject source_episode_id into learn_fact calls

tfatykhov:
## Problem

`link_episode_deterministic()` in the episode summarizer creates `fact → episode` edges by looking up facts with `source_episode_id == episode_id`. But this field was never set — `CognitiveLayer` tracks `_active_episodes[session_id]` internally but the runner never had a way to access it, so `learn_fact` was always called without `source_episode_id`.

Result: `link_episode_deterministic()` always received an empty `fact_ids` list → 0 `fact → episode` edges ever created.

## Fix

- Add `CognitiveLayer.get_active_episode_id(session_id)` — minimal public accessor for the already-tracked active episode
- In `runner.py` tool dispatch loop (both sync `_tool_loop` and streaming paths), auto-inject `source_episode_id` into `learn_fact` input when not already present
- The model never needs to know or pass the UUID; it happens transparently server-side

## Changes

- `nous/cognitive/layer.py` — +9 lines: `get_active_episode_id()` public method
- `nous/api/runner.py` — +14 lines: injection in both dispatch paths
- `tests/test_episode_id_injection.py` — new file, 7 focused tests

## Tests

Covers: injection adds episode_id, does not override explicit id, skips non-learn_fact tools, skips when no active episode, skips when no session_id.


tfatykhov:
🔍 **CodeReviewer** — F022 auto-inject `source_episode_id`

---

## Summary

The fix is correct and well-scoped. The core idea (accessor + injection before dispatch) is sound. A few issues worth addressing before merge.

---

## P1: `_active_episodes` silent miss for low-significance and resumed sessions

**File:** `nous/cognitive/layer.py`, `pre_turn` lines ~350–367

`_active_episodes` is **only** populated when a *new* episode is created:

```python
# Only reaches here if _should_create_episode() == True AND not a duplicate
self._active_episodes[session_id] = str(episode.id)
```

The explicit comment at R-P0-2 blocks loading *existing* episode IDs into `_active_episodes` (to prevent `end_session` corruption). This creates a silent miss in three cases:

1. **Process restart** — `_active_episodes` is in-memory only. After a restart, all sessions lose their episode binding. `get_active_episode_id()` returns `None`, injection is skipped, facts are stored but create zero `fact → episode` edges.
2. **Low-significance interaction** — if `_should_create_episode()` returns `False` (short message, etc.) and no prior episode exists in the dict, no injection happens.
3. **Duplicate episode detected** — `_is_duplicate_episode()` returns `True`, creation is skipped, nothing is stored in `_active_episodes`.

**Impact:** `link_episode_deterministic()` — the exact problem this PR fixes — will continue to produce empty `fact_ids` for any session that didn't go through a fresh episode creation in the current process lifetime. This likely includes real-world sessions that survive deployments.

**Suggested fix:** In `get_active_episode_id()`, fall back to a DB query for the most recent open episode for the session if `_active_episodes.get()` returns `None`. Or: populate `_active_episodes` in `pre_turn` from the DB when the session already has a known episode (only the `end_session` concern prevented this — that concern could be scoped to *not* pop the existing episode if we loaded it from DB vs. created it fresh).

---

## P1: `tc["input"]` mutation in streaming path — safe today, fragile by design

**File:** `nous/api/runner.py`, `stream_chat` ~line 858

```python
tc["input"] = {**tc["input"], "source_episode_id": active_ep}
```

`tc` is the raw `block` dict from `all_blocks`. `content_blocks` was already built from this block **before** this loop, via:

```python
content_blocks.append({"type": "tool_use", ..., "input": block.get("input", {})})
```

The content_blocks entry holds a direct reference to the *original* `block["input"]` dict. After `tc["input"] = {**tc["input"], ...}`, `tc["input"]` is rebound to a new dict but `content_blocks[-1]["input"]` still points to the original — so `messages` is not corrupted. **Currently safe.**

However, this is an accident of ordering (content_blocks built before the injection loop). Compare with `_tool_loop` which cleanly rebinds a **local variable**:

```python
tool_input = {**tool_input, "source_episode_id": active_ep}  # local rebind, unambiguous
```

The streaming path mutates a shared block dict. If anyone later changes the order — moves injection before `messages.append()` or adds a reference to `tc["input"]` after the fact — this will silently inject `source_episode_id` into the messages sent back to the model on the next API call. This would be confusing and hard to debug.

**Suggested fix:** Prefer the `_tool_loop` pattern in the streaming path too. Instead of mutating `tc`, use a local:

```python
dispatch_input = tc["input"]
if tc["name"] == "learn_fact" and "source_episode_id" not in dispatch_input:
    active_ep = self._cognitive.get_active_episode_id(session_id) if session_id else None
    if active_ep:
        dispatch_input = {**dispatch_input, "source_episode_id": active_ep}
# ... use dispatch_input for dispatch and ToolResult.arguments
```

Also: deduplicate the 5-line block from two call sites into a helper `_maybe_inject_episode_id(tool_name, tool_input, session_id) -> dict`.

---

## P2: `run_python` sandbox `learn_fact` is a blind spot

**File:** `nous/api/tools.py`, lines ~1534–1546

`run_python` provides its own `_learn_fact` closure to sandboxed Python code:

```python
def _learn_fact(content, category="technical", subject=None, confidence=1.0) -> str:
    _schedule(heart.learn(FactInput(content=content, category=category, ...)))
    return f"stored: {content[:60]}"
```

This closure:
- **Does not accept `source_episode_id`**
- Bypasses `ToolDispatcher.dispatch()` entirely
- Is not intercepted by the runner's injection logic

Any `learn_fact` call from agent-generated Python scripts will store facts **without** episode linkage — silently.

**Suggested fix:** Capture the active episode ID at script execution time (when `session_id` is available in the `run_python` tool call) and bake it into the closure:

```python
_episode_id = cognitive.get_active_episode_id(session_id) if session_id else None

def _learn_fact(content, category="technical", subject=None, confidence=1.0) -> str:
    _schedule(heart.learn(FactInput(
        content=content, category=category, subject=subject,
        confidence=confidence, source_episode_id=_episode_id,
    )))
    return f"stored: {content[:60]}"
```

---

## P2: Injection placement — runner.py vs. ToolDispatcher

The current placement in runner.py requires two identical blocks (streaming + non-streaming). `ToolDispatcher.dispatch()` already has precedent for session-aware injection (`spawn_task`, `cache_retrieve`):

```python
if session_id is not None and name == "spawn_task":
    args = {**args, "_session_id": session_id}
```

The conceptual home for this logic is `ToolDispatcher.dispatch()` — it's already the place where "inject session-local context before calling the handler" lives. The blocker is that ToolDispatcher doesn't have access to `CognitiveLayer`.

**Option A (minimal, good enough):** Extract to a `_maybe_inject_episode_id` helper on the runner, call from both sites. DRY without new coupling.

**Option B (preferred long-term):** Pass an episode-ID resolver callback to `ToolDispatcher.__init__` so `dispatch()` can do the injection generically. This also catches the `run_python` case if the resolver is threaded into the sandbox.

---

## Minor

- The test helper `_apply_injection` mirrors the runner logic exactly — consider a shared utility function so they don't drift.
- `get_active_episode_id` is synchronous (no `async`). Good. A debug log when returning `None` for a known session would make the silent-miss case observable in production logs.

---

**Verdict:** P1-1 (process-restart silent miss) is a correctness hole that undercuts the stated purpose of the PR. P1-2 (shared-ref fragility) is a latent bug risk. Both should be addressed before merge. The P2 items are non-blocking but worth tracking.

Issue #182, opened 2026-03-21
Working memory failure: model loses track of own actions after 1-2 turns


tfatykhov:
## Problem

Nous loses track of actions it performed 1-2 turns ago. Observed: Nous successfully sent an email (10 tool calls, confirmed received by user), but on the very next turn claimed "I never actually called any tools to do either." He confused his fabricated first attempt with his successful second attempt.

This is not a one-off. Combined with #179 (conversation replay after bulk operations), this points to a structural working memory problem.

## Observed Failure Modes

1. **Confabulation:** Generates a detailed summary of work (article content, file path, "Saved to: /path/file.md") without calling any tools. Reports it as done.
2. **Self-contradiction:** Successfully executes work on turn 2, then on turn 3 claims it never happened - applying the "I fabricated it" narrative from turn 1 to both attempts.
3. **Censor bypass via non-execution:** Censors only fire on tool calls. If the model skips execution entirely and just describes the result, censors never trigger.

### The Homunculus Problem

Nous's own architecture article identifies this as gap #4: a single LLM at the center of the cognitive loop. Every cognitive feature (Heart, Brain, censors, sleep, graph recall) feeds INTO the LLM's context window. But the LLM interprets that context and can:

- Ignore evidence in front of it
- Confabulate actions it didn't take
- Apply narratives from one turn to another
- Pattern-match on vibes instead of checking facts

### Four Structural Gaps

**1. No execution ground truth.**
The model's "memory" of what it did is its own text output, which can be fabricated. There's no separate, authoritative record the model is forced to consult.

**2. Compaction destroys execution history.**
When turns get summarized, the LLM summarizer keeps the discussion ("we talked about an article") and drops execution details ("I called write_file with path X"). The next turn has no evidence of what actually happened.

**3. Tool pruning removes proof.**
Tool results that PROVE execution are first to get pruned (soft-trim at age 3, metadata-degrade at 8, hard-clear at 12). After a few turns, the write_file result is gone. The model has only its own claims and no evidence.

**4. Plans and actions are indistinguishable.**
"Here's what I'll write..." and "Here's what I wrote..." look identical in conversation history. Without structural markers, both are just assistant text.

## Proposed Fix: Execution Ledger

A separate, authoritative record that lives outside conversation context.

### Design

```python
class ExecutionLedger:
    """Append-only record of what actually happened this session."""
    actions: list[ExecutedAction]  # tool name, key args, result status, timestamp
    
    def session_summary(self) -> str:
[...]

Issue #179, opened 2026-03-21
Conversation replay loop after high-volume tool operations (F025 sweep)


tfatykhov:
## Problem

After running the F025 retrieval sweep (350 API calls via `run_python`), Nous re-executes the entire sweep from scratch when asked simple follow-up questions like "push raw data to git." It replays completed steps, re-checks APIs, re-builds scripts, and re-runs queries instead of recognizing the work is done.

## Observed Behavior

1. Tim asks Nous to push results to git
2. Instead of `git add && git push`, Nous starts: "Let me check the repo state and then run the full sweep"
3. It re-verifies the API, re-examines SQL, rebuilds the sweep script
4. Runs another 350 queries
5. Reports results again

## Root Cause Analysis

The conversation history contains hundreds of near-identical tool call/response pairs from the sweep. The model pattern-matches on this dominant context and continues the "sweep" behavior instead of responding to the new message.

### Why compaction doesn't save us:

- **Compaction threshold:** 60% of context window (~120K tokens for 200K window)
- **Tool pruning tiers:** soft-trim at age 3, metadata-degrade at age 8, hard-clear at age 12
- **`run_python` profile:** "standard" (3/8/12 ages)

The problem: 350 tool calls happen within a SINGLE `run_python` execution. From the pruning system's perspective, that's ONE tool result, not 350. The entire sweep output (1.8MB of JSON) sits in one tool result block. Tool pruning operates on individual tool_result messages, not on the size of individual results.

Even after soft-trimming (keeping first 1500 + last 1500 chars), the conversation still has:
- The full sweep script in a `run_python` tool_use block
- A trimmed but still-present result showing it was a sweep
- All the assistant reasoning around it ("Let me run 350 queries...")

The assistant messages describing the sweep plan and methodology are never pruned (pruning only touches tool results). So the model sees "here's how to run a sweep" instructions in its own prior messages and follows them again.

### 1. Repetitive operation detection (new)
Detect when the model is re-executing a pattern that already exists in conversation history:
- Track tool call signatures (name + key args hash)
- If the same signature was used >N times in recent history, inject a system hint: "This operation was already completed. Results are at [location]. Proceed with the user's current request."

### 2. Aggressive pruning for bulk operations
- Add a `bulk` or `sweep` decay profile: (1, 2, 4) — aggressively clear repetitive operations
- Auto-detect bulk patterns: same tool called >10 times with similar args
[...]

Issue #45, opened 2026-02-25
Context recall quality: filter abandoned, deduplicate, limit verbosity


tfatykhov:
## Problem

The context window assembles ~8K tokens for a simple status check, mostly noise:

### 1. Abandoned decisions still recalled (#41)
5x `[failure] [abandoned — informational response] (confidence: 0.00)` shown every turn.

**Fix:** Add `AND NOT (d.outcome = 'failure' AND d.confidence = 0.0)` to `Brain._query()` filter clauses.

### 2. Abandoned episodes recalled
4 abandoned episodes shown alongside 1 real one.

**Fix:** Filter `outcome != 'abandoned'` in episode recall, or at minimum in `ContextEngine._format_episodes()`.

### 3. Fact extractor creates duplicates without superseding
Every `git pull` creates a new "latest commits" fact without deactivating the previous one. Result: multiple nearly-identical facts about the same subject eating context tokens.

**Fix:** The fact extractor should check for existing active facts with the same subject and either:
- Supersede (deactivate old, create new)
- Update in place
- Skip if content is substantially similar (text_overlap > 0.80)

### 4. Facts too verbose
Git commit facts store full commit descriptions (~200 tokens each). For context recall, a shorter summary would suffice.

**Fix:** Cap fact content length or summarize before storing. The context engine could also truncate individual facts to a token budget.

### 5. Debug display HTML-encodes system prompt
`html_module.escape()` on line 451 of `telegram_bot.py` converts `'` to `&#x27;` in the debug output. The actual system prompt is fine — it's only the debug display.

**Fix:** Don't HTML-escape the system prompt in debug, or use a code block.

## Token Budget Impact

Current: ~8.2K input for a simple query
Expected after fixes: ~2-3K input (identity + censors + frame + 3-5 relevant facts + 1-2 episodes)

## Related Issues
- #41 — Abandoned decisions in context
- #38 — _is_informational() improvements
- F008 — Memory Lifecycle (fact confirmation, supersession, retirement)

PR #186, opened 2026-03-23
spec: F024 Amnesia Prevention — 7 structural causes and fix plan

tfatykhov:
### Problem
Only 3 facts rejected by admission control out of 762. Memory loss is caused by 7 compounding structural issues in the recall and extraction pipeline.

### Root Causes
1. 6-stage filter pipeline compounds removals
2. Retrieval limits too small (5 per type)
3. Diversity filter uses first word of subject
4. User profile budget only 200 tokens
5. Transcript truncation at 8k chars
6. Fact extractor dedup too aggressive (0.85)
7. Admission grounding against lossy summaries

### Fix Plan
- Phase 1: Quick wins (limits, filters, budgets)
- Phase 2: Extraction quality (transcript, dedup)
- Phase 3: Deferred (admission rewrite)

@tfatykhov requesting review

### DIFF
--- a/docs/architecture/minsky-nous-mapping.md
+++ b/docs/architecture/minsky-nous-mapping.md
@@ -1,0 +1,481 @@
+# Minsky → Nous Architecture Mapping
+
+**Source:** Marvin Minsky, *The Society of Mind* (1986)
+**Compiled:** March 19, 2026
+**Chapters Analyzed:** All 30 (1–30)
+
+---
+
+## Executive Summary
+
+All 30 chapters of *The Society of Mind* were analyzed against Nous's current codebase. The central finding: **Nous has built the right data structures but lacks the plasticity layer.** Minsky's mind constantly reshapes itself — edges strengthen, censors auto-create, memories reconstruct differently each time. Nous stores and retrieves but doesn't reshape.
+
+Five meta-themes emerged:
+
+1. **Administration > Acquisition** (Papert's Principle, Ch. 7/10) — Better ways to use existing capabilities beats building new ones
+2. **Micronemes** (Ch. 16/20) — Ambient context signals that bias everything, completely absent from Nous
+3. **Reconstruction > Retrieval** (Ch. 15/20) — Recall should assemble from fragments, not return verbatim stored text
+4. **Relational Meaning** (Ch. 6/11/19) — Nothing has intrinsic meaning; everything means through relationships. Facts stored atomically lose meaning.
+5. **Distributed Initiative** (Ch. 1/5) — Heart/Brain/Skills should initiate actions, not just respond to LLM queries. Otherwise the LLM layer becomes the homunculus.
+
+---
+
+## Chapter-by-Chapter Mapping
+
+### Chapter 1: Building Blocks
+
+| Minsky Concept | Description | Nous Implementation | Gap |
+|---|---|---|---|
+| **Agent/Agency Duality** | Same system = mindless agent (micro) or intelligent agency (macro). Both views needed | Heart + Brain + Frame + Skills are separate agencies | ✅ Structurally correct — distributed by design |
+| **Substitution Principle** | Replace any complex agent with a subsociety of simpler agents. Never hide complexity in one box | Subtask spawning | Weak — flat subtask list, not hierarchical agent decomposition |
+| **Common Sense** | "An immense society of hard-earned practical ideas." Not simple at all | Facts + censors + procedures | Good — facts are learned heuristics, censors are rule exceptions |
+| **Amnesia of Infancy** | We forget how hard basic skills were to learn, making them seem innate | Episode consolidation / sleep | Partial — we consolidate but don't track what was hard to learn |
+
+**Key Takeaway:** The Substitution Principle means Nous should never let complexity hide inside a single tool call or handler. If something is complex, decompose it.
+
+---
+
+### Chapter 2: Wholes and Parts
+
+| Minsky Concept | Description | Nous Implementation | Gap |
+|---|---|---|---|
+| **Three-Level Problem** | Understanding requires: how each part works, how parts interact, how interactions produce global behavior | Facts (L1), graph edges (L2), spreading activation (L3) | L1 good, L2 present, L3 barely exists |
+| **Gestalt Trap** | "The whole is more than the sum of parts" is a pseudo-explanation. Specify the interactions | Facts can be isolated without knowing connections | Risk — storing facts without connections reproduces the trap |
+| **Containment = Arrangement** | No single board contains a mouse. Six boards arranged correctly do. Properties emerge from arrangement | Procedures encode sequences | Partial — procedures capture sequences but not constraint satisfaction |
+| **Easy Things Are Hard** | Simplest perceptual tasks require enormous hidden machinery | No introspective cost model | Missing — Nous doesn't know which tasks were harder than expected |
+
+**Key Takeaway:** Level 3 (emergent behavior from local interactions) is the core missing layer. Without it, the memory system is a filing cabinet, not a society.
+
+---
+
+### Chapter 3: Conflict and Compromise
+
+| Minsky Concept | Description | Nous Implementation | Gap |
+|---|---|---|---|
+| **Noncompromise (3.2)** | Conflicts migrate upward. When subordinates stall, rivals seize control | Task cancellation | No conflict-migration detection |
+| **Heterarchies (3.4)** | Not everything is hierarchical. Loops require memory | EventBus + graph edges | ✅ Working well |
+| **Pain/Pleasure Simplify (3.6)** | Both narrow focus. Too much simplification degrades the self | Block-level censors = pain | Risk: too many block censors over-simplify |
+
+**Key Takeaway:** Stalled subtasks should trigger alternative strategies, not just timeout.
+
+---
+
+### Chapter 4: The Self
+
+| Minsky Concept | Description | Nous Implementation | Gap |
+|---|---|---|---|
+| **One Self or Many** | Self is a society of sub-agents with different concerns | Single agent_identity | Frames could be "partial selves" with own priorities |
+| **Self-Control via Censors** | You don't resist bad actions by willpower — you prevent them from being considered | Absolute censors | ✅ Good alignment |
+| **Identity as Narrative** | Personal identity is a story, not a fact | agent_identity document | ✅ Exactly this — a story Nous tells itself |
+
+**Key Takeaway:** Each frame could be a "partial self" with its own defaults, priorities, and active skills.
+
+---
+
+### Chapter 5: Individuality
+
+| Minsky Concept | Description | Nous Implementation | Gap |
+|---|---|---|---|
+| **Circular Causality (5.1)** | Goals form feedback loops, not linear chains. Neither cause is "first" | No goal-loop tracking | Missing — no way to represent mutually reinforcing goals |
+| **Loop-Breakers (5.2)** | Dogma and external anchors break infinite regress. "Just because!" is legitimate self-regulation | Absolute censors + values | ✅ Censors are exactly Minsky's loop-breakers |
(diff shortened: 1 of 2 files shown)


tfatykhov (commented):
## ⚡ Emerson — F024 Amnesia Prevention Review

**Verdict: APPROVE with revisions.** This is the most important spec for Nous right now. The diagnosis is accurate, well-grounded in the code, and the fix plan is practical. A few issues to address before implementation.

---

### ✅ What's Strong

1. **Root cause analysis is code-verified.** Every RC traces to specific files and lines. I verified all 7 against `main`:
   - RC-1: `context.py` L379-395 — confirmed 7-stage pipeline (staleness → frame_boost → diversity → dedup → usage_boost → relevance_floor → diminishing_returns)
   - RC-2: `context.py` L364 — confirmed `_limits.get("fact", 5)` default across all 4 types
   - RC-3: `context.py` L691 — confirmed `raw.split()[0].lower()` grouping
   - RC-4: `schemas.py` L118 — confirmed `user_profile: int = 200`
   - RC-5: `episode_summarizer.py` — confirmed `max_chars: int = 8000`
   - RC-6: `fact_extractor.py` L119 — confirmed `existing[0].score > 0.85`
   - RC-7: Plausible (couldn't inspect admission.py directly, but architecture confirms summary-based grounding)

2. **Phasing is smart.** Quick wins first (1-4), extraction quality second (5-6), deferred work last (7). This is the right order.

3. **The compounding insight is the key finding.** Each filter independently seems reasonable. The problem is their multiplicative interaction. A fact at 0.50 relevance → 0.42 after staleness → killed by 0.45 floor. This is a real and non-obvious failure mode.

---

### 🔴 P1 — Must Fix Before Implementation

**P1-1: Change 3 code sample contradicts the description.**

The spec says "Replace first-word grouping with full subject string matching" but the proposed code still uses:
```python
group_key = item.subject.strip().lower()
```

This is the full subject, which is correct for the fix. But the "Before" comment says `raw.split()[0]` and "After" says `item.subject.strip().lower()` — the actual code in `_enforce_diversity` uses `raw.split()[0].lower()` where `raw = getattr(item, topic_attr, None)`. The fix needs to be in `_enforce_diversity`, not inline. Show the actual method change.

Also: full-subject grouping may be too granular — "Tim's email" and "Tim's email address" become separate groups, defeating diversity entirely. Consider a middle ground: use the full subject but with fuzzy matching or a 2-3 word prefix instead of 1-word. Or better: use Jaccard similarity on subject tokens with a threshold.

**P1-2: Change 1 is too aggressive — staleness has safety valves.**

The spec proposes removing staleness entirely. But `_apply_staleness_penalty` (L645-671) already exempts `rule` and `preference` categories and has a floor (`max(decay, 0.3)` — score can't drop below 30% of original). The real problem isn't staleness existing; it's staleness running before the relevance floor. A fact at 0.50 decayed to 0.42 then killed at 0.45 floor — if staleness ran AFTER the floor, the 0.50 fact survives.

Proposal: **Move staleness to after relevance floor** instead of removing it. Old facts that are irrelevant SHOULD decay. The issue is ordering, not existence.

---

### 🟡 P2 — Should Fix

**P2-1: Retrieval limit increase (Change 2) needs math.**

Tripling fact retrieval from 5→15 sounds right, but the spec doesn't address the interaction with token budgets. The facts section budget is 1500 tokens (L123). With `_scaled_budget` at 2.5x for 1M+ context windows, that's 3750 tokens. At ~30-50 tokens per formatted fact, 15 facts (even after filtering) would use 450-750 tokens — well within budget. **The math works**, but the spec should show it so implementers understand why 15 is the right number, not 10 or 20.

Also: decisions and episodes going from 5→10 have tighter budgets (episodes: 1000 tokens). Verify these fit too.

**P2-2: Retrieval limits should be configurable, not hardcoded.**

The current architecture already supports per-type limits via `RetrievalPlan` (L142-146). The default `5` is just the fallback. The fix should change the fallback defaults AND document that `RetrievalPlan` overrides them. This future-proofs the change — different frames may want different limits.

**P2-3: Missing before/after measurement plan.**

The testing plan says "run 10 test queries" but doesn't specify WHAT queries or how to measure. Suggest:
- Define 10 canonical queries spanning recent conversations
- Record: items retrieved, items surviving each filter stage, final count
- Add temporary logging at each pipeline stage (`logger.info("stage=%s, before=%d, after=%d")`) so you can see exactly where items die
- This logging should be permanent (DEBUG level) — it's the observability the pipeline has always needed

---

### 💡 P3 — Suggestions

**P3-1: Frame-dependent retrieval limits (Open Question 2).**
Yes, absolutely. The `ContextBudget.for_frame()` at schemas.py L138-144 already varies token budgets by frame. It should also vary retrieval limits. `task` and `debug` frames should pull 20+ facts; `conversation` frame can stay at 10.

**P3-2: Pinned facts (Open Question 1).**
This is valuable but should be a separate spec. It's a different mechanism (always-include vs better-recall). Don't mix them.

**P3-3: The telegram_bot.py changes don't belong in this PR.**
The zombie session TTL fix is a good change, but it's unrelated to amnesia prevention. Separate PR for clean history.

**P3-4: Change 5 should prefer chunked summarization.**
Simply raising the char limit from 8k→16k is a stopgap. The episode_summarizer already has `_truncate_transcript` with turn scoring — extending to chunked summarization (split → summarize each → merge) is the right long-term fix. Add it as Phase 2.5 or note it as the follow-up.

**P3-5: Embedding quality (Open Question 3).**
Worth investigating separately. The RRF fix (PR #178) improved scoring, but if the underlying embeddings cluster poorly, retrieval limits and filters are just papering over it. Suggest a separate analysis: sample 50 facts, compute pairwise cosine similarity, look for unexpected clusters.

---

### Summary

| # | Severity | Issue |
|---|----------|-------|
| P1-1 | Critical | Diversity fix code sample is misleading + full-subject may be too granular |
| P1-2 | Critical | Don't remove staleness — reorder it after relevance floor |
| P2-1 | Should Fix | Show the token budget math for new limits |
| P2-2 | Should Fix | Make limits configurable via defaults, not just hardcoded |
| P2-3 | Should Fix | Define concrete measurement queries + add pipeline stage logging |
| P3-1 | Nice | Frame-dependent retrieval limits |
| P3-2 | Nice | Pinned facts → separate spec |
| P3-3 | Nice | telegram_bot.py → separate PR |
| P3-4 | Nice | Chunked summarization as follow-up |
| P3-5 | Nice | Embedding quality analysis |

The diagnosis is excellent. Fix P1-1 and P1-2, address P2s in the spec text, and this is ready to implement.

tfatykhov:
🔍 **CodeReviewer** — Spec Review: F024 Amnesia Prevention

**Verdict: APPROVE with revisions** — All 7 root causes verified against code. Good diagnosis, solid phased plan. But there are 2 P1 issues that must be fixed before merge, plus several P2/P3s.

---

## P1 — Must Fix

### P1-1: Feature number collision

**F024 is already assigned** to "Critic Agent: Speculative Parallel Cognitive Execution" (`docs/features/F024-critic-agent.md`, Draft v2). This spec needs a new number. Suggest **F028** (checking feature registry).

### P1-2: Code bundled with spec

This PR mixes a **spec document** with a **code change** (`telegram_bot.py` zombie session fix — SESSION_TTL_SECONDS, stale session expiry). These should be separate PRs:
- The spec PR is for review/discussion
- The code change is a standalone bugfix that should have its own tests and review cycle

Split the `telegram_bot.py` changes into a separate PR.

---

## P2 — Should Fix

### P2-1: Change 1 (staleness removal) is too aggressive

The spec says *"Remove the staleness penalty from the main pipeline"* — but the actual `_apply_staleness_penalty()` in `context.py:645` already has safeguards:
- **Exempts** `rule` and `preference` categories (line 662-663)
- **Floors decay at 0.3** — can reduce a score by at most 70% (`max(decay, 0.3)`, line 668)
- Uses configurable `staleness_half_life_days`

Removing it entirely loses the recency signal. Better approach: **reorder** staleness to run *after* the relevance floor, so the floor catches low-quality items first, then staleness re-ranks survivors. This preserves recency as a tiebreaker without it being an eliminator.

### P2-2: Token budget math doesn't add up

Change 4 proposes raising `user_profile` from 200→500 tokens. Current `ContextBudget` section totals:

| Section | Tokens |
|---------|--------|
| identity | 500 |
| user_profile | 200 |
| censors | 300 |
| frame | 500 |
| working_memory | 700 |
| decisions | 2000 |
| facts | 1500 |
| procedures | 1500 |
| episodes | 1000 |
| **Total** | **8200** |

Already 200 over the 8000 total budget. Adding 300 more tokens of user_profile increases pressure. The spec should show where those tokens come from (reduce another section?) or acknowledge the total budget needs raising too.

### P2-3: Three known amnesia causes missing

The spec identifies 7 root causes but omits 3 known issues:

1. **Tool result pruning destroys execution proof** (#182) — `compaction.py:264` `prune_tool_results()` strips tool call evidence, so Nous can't recall what it actually did
2. **Compaction loses action records** (#179) — LLM-powered history compaction summarizes away tool actions
3. **A-MAC threshold calibration** — Emerson's F023 review found only 1.7% rejection rate, suggesting the threshold may be too conservative (letting noise through) rather than too aggressive

These interact with RC-5/RC-6/RC-7 and should at least be listed as "Related known issues" even if out of scope.

### P2-4: No F026 interaction analysis

F026's `ExecutionLedger` is **session-scoped and in-memory only** (`execution_ledger.py:1-7`). If facts about tool execution are lost in compaction (#182/#179), the ledger can't help because it doesn't persist. The spec should note this gap — the amnesia pipeline extends beyond retrieval into the execution record.

### P2-5: Measurement plan too weak

*"Run 10 test queries and record what surfaces"* is not rigorous enough for a spec claiming ~40% recall improvement. Need:
- Defined query set with **expected facts** for each query
- Automated before/after comparison (not manual observation)
- Baseline recall % measured, not assumed (the spec claims ~0.7% but doesn't show how this was calculated)

---

## P3 — Nice to Have

### P3-1: Change 3 code sample creates a no-op filter

The proposed fix:
```python
group_key = item.subject.strip().lower()
```
Uses the **full subject string** as the group key. With 762+ facts, nearly every subject is unique → diversity filter becomes a no-op. Consider:
- First **2-3 words** instead of 1 (balances between too-coarse and too-fine)
- Or semantic clustering (heavier but more accurate)

### P3-2: RC-7 file paths wrong

Spec says `nous/handlers/admission.py` — actual location is `nous/heart/admission.py`. And `_get_source_text()` is in `nous/heart/facts.py:483`, not the admission module. Update paths for accuracy.

### P3-3: Retrieval limits should be configurable

Change 2 proposes hardcoded limit changes (5→15 for facts, etc.). These should be fields in `Settings` or `ContextBudget` so they can be tuned via environment/config without code changes — matching the existing pattern for `relevance_floor_enabled`, `staleness_half_life_days`, etc.

### P3-4: Change 6 dedup history

The 0.85 threshold was already raised from 0.65→0.85 in #45 because 0.65 was too aggressive. Worth noting this history in the spec — it shows the dedup threshold has been tuned before and the direction has been toward *more* permissive, supporting the 0.85→0.92 change.

### P3-5: Open Question 3 should be promoted

*"Is the current embedding model producing good enough similarity scores?"* — this could be the root cause underlying ALL 7 issues. Poor embeddings → low scores → filters kill everything. Worth promoting to a root cause investigation rather than an open question.

---

## Summary

| Severity | Count | Key Items |
|----------|-------|-----------|
| P1 | 2 | Feature number collision, code bundled with spec |
| P2 | 5 | Staleness too aggressive to remove, token budget math, missing causes, F026 gap, weak measurement |
| P3 | 5 | Diversity no-op, wrong file paths, configurability, dedup history, embedding quality |

**The diagnosis is strong** — all 7 root causes verified in the codebase. The phased implementation plan is sensible. Fix the P1s (renumber + split PR) and address the P2s (especially the staleness reorder and token budget math) and this is ready to guide implementation.

---
*Decision: c938167a | Prior related: 7ee91a7c*

PR #300, opened 2026-04-10
fix: preserve full 100-decision window for skewed outcomes in procedure learner

tfatykhov:
## Summary
- Replace two fixed `limit=50` queries in procedure learner with a budget-based approach
- Fetch successes up to `_TOTAL_WINDOW=100`, then use remaining budget for partials
- Prevents skewed distributions (many successes, few partials) from silently shrinking the candidate pool below 100

Addresses P2 review comment from PR #189.

## Test plan
- [x] All 21 procedure learner tests pass
- [x] Cherry-picked from orphaned commit `29d76c3` (originally authored by Nous agent)


tfatykhov:
Redundant — same fix already merged as PR #298 (commit 5c5aadd).

PR #398, opened 2026-05-02
fix(heart): rollback session after recall sub-search failure

tfatykhov:
## Summary
- One sub-search in `Heart._recall` raising a SQL-level error poisons the shared `AsyncSession`, causing every following sub-search to fail with `InFailedSQLTransactionError` — silently zeroing out recall.
- Fix: `await session.rollback()` after a caught sub-search exception clears the failed transaction state so the loop can continue.
- 1 line of real code, wrapped in a defensive try/except so rollback-of-rollback failures log loudly.

## Discovery context
Surfaced today during F051 context-packing eval work. The eval-scratch DB was missing migration 040 (`heart.episodes.session_id`); episode SELECT crashed with `UndefinedColumnError`; the cascade silently killed the fact and procedure sub-searches. `heart.recall` returned 0 results for every query. The packing eval reported **0% sufficient** — looked like a retrieval-quality crisis; was actually an infra failure hidden by cascade silence.

The same class of failure could bite prod at any time:
- transient pgvector hiccup mid-query
- future schema mismatch during a deploy window  
- asyncpg connection blip

With this fix, only the failed sub-search loses its results; the rest still return.

## Test plan
- [x] New test `test_recall_subsearch_failure_does_not_cascade` — monkeypatches `episodes.search` to raise, asserts (a) `session.rollback()` fires, (b) `facts.search`/`procedures.search`/`censors.search` are still invoked instead of the loop bailing out
- [x] Test fails on `main`, passes with the fix (TDD-verified)
- [ ] CI on Postgres (local Windows env has pre-existing sqlite/pgvector incompatibility that affects 8 unrelated tests on main too — confirmed by stash + re-run)

### DIFF
--- a/nous/heart/heart.py
+++ b/nous/heart/heart.py
@@ -891,6 +891,19 @@ class Heart:
             except Exception as exc:
                 keys.append(memory_type)
                 results_list.append(exc)
+                # The shared AsyncSession may be in a failed-transaction state
+                # (e.g., asyncpg InFailedSQLTransactionError after a SQL-level
+                # error like UndefinedColumnError). Rolling back clears the
+                # failed state so subsequent sub-searches in this loop don't
+                # cascade-fail with "current transaction is aborted".
+                try:
+                    await session.rollback()
+                except Exception:
+                    logger.warning(
+                        "Recall sub-search rollback failed for %s; "
+                        "remaining sub-searches will likely fail",
+                        memory_type, exc_info=True,
+                    )
 
         # Use original search scores instead of RRF positional scores.
         # Episodes, facts, and procedures use hybrid_search() (configurable
--- a/nous/heart/heart.py
+++ b/nous/heart/heart.py
@@ -776,8 +776,17 @@ class Heart:
         """
         if session is None:
             async with self.db.session() as session:
-                return await self._recall(query, limit, types, session, residual_activations)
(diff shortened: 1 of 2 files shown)


tfatykhov:
Addressed both review findings in 5f0cb13:

**Important #1 (silent-data-loss trap)**
Added `owns_session: bool` parameter to `_recall`. The public `recall()` sets it `True` when Heart opens its own session, `False` when caller provides one. Rollback only fires when `owns_session=True`. Caller-provided-session callers must handle their own transaction recovery — no silent rollbacks behind their back.

Today no live caller passes a shared session (production `pre_turn` uses `session=None`), but this guards the trap before any future caller springs it.

**Important #2 (test didn't exercise asyncpg state)**
Added `test_recall_subsearch_real_sql_error_does_not_cascade` — issues a real bad SQL query (`SELECT bogus_col_xyz FROM heart.episodes`) inside the broken sub-search, then asserts the next sub-search completes without raising. Proves the rollback works end-to-end against real asyncpg `InFailedSQLTransactionError` semantics, not just that we call `.rollback()`.

Also added `test_recall_subsearch_failure_skips_rollback_when_caller_owns_session` to lock in the new ownership semantics: when caller passes a session, rollback is NOT called.

3 tests total, all passing.

PR #399, opened 2026-05-02
feat(eval): pre-flight assert eval DB matches ORM schema

tfatykhov:
## Summary
- Adds `nous_eval/schema_preflight.py` — introspects ORM models the harness uses and asserts every required column exists in the eval DB.
- Wires it into `_build_heart_for_eval` so it runs once per harness invocation, before any `heart.recall` fires.
- Raises `EvalDBSchemaDriftError` with all drift in one pass plus a one-line remediation command.

## Why
Today's investigation: the eval-scratch DB was missing migration 040 (`heart.episodes.session_id`). The ORM SELECT crashed with `UndefinedColumnError`, asyncpg marked the connection's transaction `ABORTED`, and every subsequent fact/procedure sub-search in the same session failed silently with `InFailedSQLTransactionError`. `heart.recall` returned 0 results for every query and the packing eval reported **0% sufficient** — looked like a retrieval-quality crisis; was actually schema drift hidden by cascade silence.

PR #398 fixed the cascade behavior. This PR fixes the *root cause* signal: schema drift now fails loud at startup with a clear message, instead of cascading silently into "0% sufficient."

## What it catches
Live demo against the running `nous-eval-scratch` container (after artificially dropping `session_id`):
```
EvalDBSchemaDriftError: Eval DB schema drift detected — ORM-required columns missing:

  heart.episodes: session_id

This usually means the eval DB image was built before one or
more recent migrations. To apply pending migrations to the
running eval-scratch container:

  for f in sql/migrations/*.sql; do
    docker exec -i nous-eval-scratch psql -U nous       -d nous_eval_scratch < "$f";
  done
```

## Test plan
- [x] 4 mock-based unit tests in `tests/test_schema_preflight.py`:
  - pass case (all columns present)
  - single-column drift (the migration-040 scenario)
  - multi-table drift reported in one error
  - missing-table-entirely degrades gracefully
- [x] Live-tested against `nous-eval-scratch` — passes after migrations applied; fails loud when columns dropped
- [x] Existing `test_retrieval_pipeline.py` unaffected (8/8 pass)

### DIFF
--- a/nous_eval/ingest_longmemeval.py
+++ b/nous_eval/ingest_longmemeval.py
@@ -275,6 +275,12 @@ async def _replay_sessions_into_scratch(
     api_client = None
     try:
         await db.connect()
+        # Surface schema drift before we start ingesting — without this,
+        # a stale-schema container would let inserts silently poison the
+        # asyncpg session and the operator sees an empty corpus instead
+        # of a clear schema error.
+        from nous_eval.schema_preflight import assert_eval_db_schema_matches_orm
+        await assert_eval_db_schema_matches_orm(db)
         embedder = EmbeddingProvider(
             api_key=settings.openai_api_key,
             model=settings.embedding_model,
--- a/nous_eval/retrieval_runner.py
+++ b/nous_eval/retrieval_runner.py
@@ -400,6 +400,12 @@ def _build_brain_for_eval(
 
     Brain and Heart share the embedding provider (pattern from main.py:69)
     so we don't double up httpx pools.
+
+    Schema preflight: Brain's ORM models (Decision) are validated by the
+    Heart preflight at ``_build_heart_for_eval`` because both factories
+    are called in the same harness invocation against the same eval DB.
+    If you ever construct a Brain in a path that does NOT also build a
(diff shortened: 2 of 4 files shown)


tfatykhov:
Addressed all three review findings in 9449998:

**Critical #1 (ingest path bypass)** — Added `await assert_eval_db_schema_matches_orm(db)` right after `db.connect()` in `_replay_sessions_into_scratch` (the LongMemEval ingest path). That path populates the eval DB, so a stale-schema container would silently poison every insert. Now fails loud at the top instead.

**Important #2 (Brain asymmetry footgun)** — Added a docstring note in `_build_brain_for_eval` documenting the convention: Brain's ORM models (Decision) are validated by the Heart preflight at `_build_heart_for_eval` because both factories are called in the same harness invocation. Anyone constructing a Brain in a path that doesn't also build a Heart needs to call the preflight at that site.

**Important #3 (remediation message not portable)** — Replaced the bash `for ... do ... done` loop and hard-coded `nous-eval-scratch` container name with the single canonical command `uv run python -m nous_eval.rebuild`. Works cross-platform and respects compose's actual service naming.

All 4 unit tests still pass.

PR #400, opened 2026-05-02
feat(eval): bucket packing scenarios as memory vs docs

tfatykhov:
## Summary
- Adds a `bucket: memory|docs` field to `PackingScenario`.
- Classifies the 2 scenarios whose gold expects CLAUDE.md / source-code info (`subtask_workers`, `skill_management`) as `docs`; the other 6 stay `memory`.
- Headline sufficiency now reports the memory bucket only; docs scenarios surface separately as a known-limitation aside.

## Why
Today's investigation found 3 of 8 packing scenarios were dragging the eval down for the wrong reason. Their `gold_answer_hint` references information that lives in CLAUDE.md / source code (env-var defaults like `NOUS_SUBTASK_WORKERS`, internal classnames like `SkillParser`, `bootstrap process`) — things the agent never had reason to verbalize as a fact / episode / decision. A failure on those scenarios isn't a retrieval-quality signal; it's a scenario-design signal.

The previous metric mixed both, so a healthy retrieval looked like 25% sufficient when really 6/6 of the genuinely-retrieval-testable scenarios pass under MMR-forced.

## What changes in the report

Before:
```
- sufficiency: **2/8 (25%)**
```

After (same data, bucket-separated):
```
- **headline sufficiency (memory bucket): 2/6 (33%)**
- docs aside (known-limitation gold hints): 0/2 (0%)
```

The headline now tracks what we actually want to measure.

## Bucketing audit (corpus-probe-based)

| scenario | bucket | rationale |
|---|---|---|
| subtask_workers | docs | Gold expects `NOUS_SUBTASK_WORKERS` env var default — corpus probe found 0 verbatim matches |
| skill_management | docs | Gold expects `SkillParser` / `bootstrap process` / `auto-activation via RECALL` — corpus probe found 0 matches for those exact terms even though F011 facts exist |
| telegram_email | memory | "subtask completion notifications" was found in corpus (19 hits); "bot token" env-var bit is borderline but the substantive part is memorizable |
| heartbeat_overview | memory | F034 architectural facts present (43 hits) |
| rubric_evolution | memory | RubricEvolver fact present in corpus |
| procedure_learning | memory | F012 K-Line Learning fact present |
| graph_densification | memory | F040 fact present (12 hits for the literal phrase) |
| cognitive_loop | memory | "Sense→Frame→Recall→..." fact present (just under different phrasing than gold hint) |

## Test plan
- [x] 5 unit tests in `tests/test_packing_buckets.py`:
  - aggregation separates the two buckets correctly
  - empty-docs and empty-memory edge cases don't divide-by-zero
  - missing `bucket` field defaults to memory (backward-compat)
  - the 8 SCENARIOS are bucketed exactly as the audit prescribes (so renaming/adding is a deliberate decision)
- [x] Smoke-imported the script — SCENARIOS list and `aggregate_by_bucket` both work
- [x] Existing eval invocations are unchanged for users running `--max-scenarios N`; they just see the new bucket-aware report format

## Related
Third and final PR from the eval-DB schema-drift investigation. Sibling PRs #398 + #399 already merged.

### DIFF
--- a/scripts/eval/eval_context_packing.py
+++ b/scripts/eval/eval_context_packing.py
@@ -46,9 +46,22 @@ _DEFAULT_MODEL = "claude-sonnet-4-6"
 
 @dataclass
 class PackingScenario:
+    """One end-to-end packing scenario.
+
+    ``bucket`` separates retrieval-quality probes from documentation-only
+    probes. Set ``bucket="docs"`` when the gold_answer_hint references
+    information that lives in CLAUDE.md / source code (env-var defaults,
+    internal classnames) rather than something the agent would naturally
+    memorize as a fact/episode/decision. Headline sufficiency is reported
+    over ``bucket="memory"`` only — docs scenarios are surfaced separately
+    as a known-limitation aside so they don't depress the metric we
+    actually use to judge retrieval quality.
+    """
+
     name: str
     user_message: str
-    gold_answer_hint: str  # what info needs to be in context
+    gold_answer_hint: str
+    bucket: str = "memory"  # "memory" or "docs"
     notes: str = ""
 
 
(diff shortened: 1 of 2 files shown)


tfatykhov:
Independent review came back clean — merge as-is. The only flag was non-blocking: `scripts/eval/eval_context_packing_loose.py` is a duplicated SCENARIOS copy that will silently drift from this PR's bucketing.

Filing a follow-up to either delete the loose variant (it was a one-off A/B test of judge strictness, already concluded) or extract SCENARIOS into a shared module both can import. Leaving that for a separate small PR rather than expanding scope here.

Reviewer also confirmed:
- Bucketing of all 8 scenarios is correct (telegram_email's borderline call is defensible because the "subtask completion notifications" half had 19 corpus hits)
- Zero downstream consumers of the old JSON shape (`eval_context_packing.json` is not read by any CI/dashboard/regression script) — schema break is safe
- The `test_scenarios_correctly_bucketed` brittleness is load-bearing and intentional (bucket choice should require deliberate human re-affirmation)

PR #401, opened 2026-05-03
feat(eval): architectural-query retrieval regression probe

tfatykhov:
## Summary
- Adds `nous_eval.probes.arch_query` — a 6-scenario regression probe for "tell me about X" / "how does X work" / "what does X do" queries.
- Catches degradation that would slip past F051's aggregate qrels (which weight specific-fact lookups over architectural questions).
- Run with `--strict` for CI: exit 1 if TOP-3 < 4 or TOP-10 < 5 (one-scenario headroom over today's baseline).

## Origin (a debugging cautionary tale)
A context-packing investigation seemed to show operational chatter outranking foundational facts ("F034 Heartbeat" buried at hybrid rank 64; "Sense→Frame→Recall" cognitive loop summary at rank 115). I tested two candidate fixes:

| Fix | Arch-query lift | F051 regression |
|---|---|---|
| `vector_weight=0.85` (from 0.7) | none — actually slightly worse on cognitive_loop (115 → 135) | none (identical metrics) |
| `ce_max_candidates=100` (from 30) | hand-eyed wins on heartbeat (top-1 became foundational) | none (identical metrics) |

Both produced **identical F051 metrics** (MRR 0.828, P@1 0.778, R@10 0.922). The arch-query "regression" turned out to be a **probe v1 artifact** — the auto-gold-finder (shortest content match) prefers operational counter facts over longer foundational summaries.

After hand-curating gold IDs, **actual architectural retrieval is healthy**:
| metric | result |
|---|---|
| TOP-1 | 4/6 (67%) |
| TOP-3 | 5/6 (83%) |
| TOP-10 | 6/6 (100%) |

This PR locks that finding into a regression test so the next investigation doesn't chase the same phantom.

## What it covers
6 probes spanning distinct query classes (test-asserted):
- Feature-prefix: F011, F012, F034, F040
- Concept: rubric evolver, cognitive loop
- System-summary: heartbeat system, cognitive loop

## Test plan
- [x] 5 mock-based unit tests in `tests/test_arch_query_probe.py` (probe well-formedness, query-class diversity, regression-floor headroom, gold-ID union semantics)
- [x] Live probe runs end-to-end against `nous_eval_scratch` — passes with `--strict`
- [x] No production code touched; eval-only addition

## Related
Sibling of #398 (heart cascade rollback), #399 (eval-DB schema preflight), #400 (packing scenario buckets) — all part of the eval-framework hardening surfaced by today's investigation.

### DIFF
--- a/nous_eval/probes/__init__.py
+++ b/nous_eval/probes/__init__.py
@@ -1,0 +1,13 @@
+"""Diagnostic probes for retrieval-quality investigation.
+
+Each probe targets a specific *behavioural* dimension of recall (rather
+than the aggregate MRR/P@K metrics that ``nous_eval.retrieval`` already
+covers). Probes are designed to:
+
+- Catch regressions on classes of query that the F051 corpus underweights
+  (e.g., architectural / "tell me about X" questions).
+- Produce human-readable per-scenario verdicts so a regression points
+  directly at the failing case.
+
+Run a probe via ``python -m nous_eval.probes.<name>``.
+"""
--- a/nous_eval/probes/arch_query.py
+++ b/nous_eval/probes/arch_query.py
@@ -59,9 +59,15 @@ class ArchProbe:
 
 # Hand-curated probes against agent_id=nous-prod-snapshot. If the
 # corpus snapshot is regenerated and a fragment no longer matches,
-# the probe will SKIP that scenario (won't fail loud) — but the
-# headline TOP-3 / TOP-10 totals will drop, which is the load-bearing
-# regression signal.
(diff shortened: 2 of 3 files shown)


tfatykhov:
Addressed reviewer's coverage-floor concern in f3de630: added `_MIN_EVALUATED_FRACTION = 2/3` so corpus drift can't silently lower `n_evaluated` and trivially pass the strict floors. `--strict` now distinguishes COVERAGE REGRESSION (stale fragments) from RANKING REGRESSION (real recall degradation). Plus a contract docstring on `gold_fragments` per reviewer #4. 6/6 tests pass.

PR #404, opened 2026-05-03
feat(eval): sleep-cycle health monitor — found 3 silently-broken phases

tfatykhov:
## Summary
- New probe `nous_eval.probes.sleep_cycle_health` aggregates the past N `sleep_completed` events from prod and verdicts each phase against sanity bounds derived from observed baseline.
- **Discovered 3 silently-broken sleep phases** in prod on first live run (passive; no actions taken on prod).
- `--strict` for CI: exit 1 if any phase shows persistent silent-failure pattern.

## Findings on first live run

n=14 cycles, agent=`nous-default`, range 2026-04-26 → 2026-05-02:

| phase | field | min | max | mean | recent zero-streak | verdict |
|---|---|---:|---:|---:|---:|---|
| reflect | facts_created | 3 | 5 | 3.6 | 0 | **GREEN** |
| resolve_contradictions | contradictions_resolved | 0 | 2 | 0.6 | 4 | GREEN |
| **stale_scan** | stale_deactivated | 0 | 0 | 0.0 | **14** | **RED** |
| **cluster_consolidation** | clusters_merged | 0 | 0 | 0.0 | **14** | **RED** |
| graph_densification | orphan_edges_created | 0 | 58 | 21.9 | 0 | GREEN |
| graph_densification_ce | ce_backfill_survived | 17 | 93 | 46.5 | 0 | GREEN |
| bridge_linker | bridge_edges_created | 0 | 20 | 2.2 | 0 | GREEN |
| **procedures** | procedures_created | 0 | 0 | 0.0 | **14** | **RED** |

**Three silently-broken phases:**
- **stale_scan** never deactivates anything despite the agent being continuously active for weeks
- **cluster_consolidation** (F027) never merges clusters — the LLM rejects every candidate
- **procedures** (F012 K-Line) never synthesizes — the cluster-detection step probably never produces an eligible cluster

These are real prod issues. Per-cycle the sleep handler reports `phases_completed=[..., procedures]` and looks healthy, but aggregation across cycles reveals the truth.

## Why "passive aggregation" instead of the planned "trigger-on-snapshot"

Original design (greenlit yesterday) called for triggering a synthetic sleep cycle on the eval DB. Investigating the data model revealed prod already emits all the stats we need into `nous_system.events`. Reading 14 prod cycles is *cheaper* than triggering one, *catches the same failure modes* the trigger approach would catch, AND **catches real prod failures the trigger approach would miss** (a controlled snapshot is too small to surface a 14-cycle silent-failure pattern).

[...]

PR #405, opened 2026-05-03
fix(sleep): unblock 3 silently-broken sleep phases

tfatykhov:
## Summary
Fixes the 3 silently-broken sleep phases discovered by the sleep cycle health monitor (PR #404).

| Phase | Before | After (expected) |
|---|---|---|
| **stale_scan** | 0/14 cycles (filter structurally impossible) | Activates when corpus has >60-day never-recalled facts |
| **cluster_consolidation** | 0/14 cycles (always picks accumulating subjects too big to merge) | Picks small mergeable clusters (3-10 facts); ~1-3 merges per cycle expected |
| **procedures** (F012) | 0/14 cycles (7-day recency gate yields ~3 of 200 candidates) | 30-day window yields ~71 of 200; gate becomes meaningful filter not a block |

### stale_scan — old filter was empty by design
Filter: `active=true AND superseded_by IS NOT NULL AND confidence < 0.5`
- Supersede flow deactivates AND sets the link in the same transaction, so `active=true AND superseded_by IS NOT NULL` is always 0 rows.
- `confidence < 0.5` matches 0 rows in prod (min observed confidence is 0.70, mean 0.91).

New filter: never-recalled facts older than 60 days, excluding `rule` category (user directives shouldn't be auto-deactivated on recall stats).

### cluster_consolidation — picked the wrong clusters every time
Old: `HAVING count >= 3 ORDER BY count DESC LIMIT 5` always landed on accumulating subjects (`lesson_learned` 164 facts, `Tim` 36 facts). LLM correctly refused to merge those into one fact. The 11 actually-mergeable small clusters never got a chance.

New: `HAVING count BETWEEN 3 AND 10` — caps cluster size so accumulating subjects skip and small mergeable clusters surface.

### procedure_learner._check_recency — 7-day window too tight
Hardcoded `cutoff = now - 7 days`. Of 200 successful-reviewed-with-bridge prod decisions, only 3 fall in the last 7 days. For a cluster to satisfy the gate it had to contain one of those 3 — vanishingly rare.

Made configurable via `procedure_recency_days` (default 30). 30-day window yields ~71 of 200 candidates eligible.

## New settings (all safe defaults)
```
stale_scan_age_days: int = 60
stale_scan_excluded_categories: list[str] = ["rule"]
cluster_consolidation_min_facts: int = 3
cluster_consolidation_max_facts: int = 10
procedure_recency_days: int = 30
```

## Verification
After deploy, re-run `nous_eval.probes.sleep_cycle_health`. Expected:
- cluster_consolidation: 1-3 merges per cycle → GREEN
- procedures: 0-3 procedures per cycle → GREEN
- stale_scan: still 0 short-term (corpus too young; oldest fact 2026-02-24), will activate as it ages → YELLOW until then

## Test plan
- [x] 8 mock-based unit tests in `tests/test_sleep_phase_fixes.py`
- [x] Settings() parses cleanly with new fields
- [x] All test logic verified against prod-data analysis

## Related
Acts on findings from PR #404. Closes the loop on the eval-tooling-hardening series — the eval found the bugs; this PR fixes them.

### DIFF
--- a/nous/config.py
+++ b/nous/config.py
@@ -105,6 +105,30 @@ class Settings(BaseSettings):
     staleness_penalty_enabled: bool = True
     staleness_half_life_days: int = 30
 
+    # Sleep-cycle stale_scan phase: deactivate facts that are old AND
+    # never recalled. Prior filter (active=true AND superseded_by IS NOT NULL
+    # AND confidence < 0.5) was structurally impossible — the supersede flow
+    # already deactivates at the same time, and prod confidence distribution
+    # never goes below 0.7. New filter targets the actual stale-fact pattern.
+    stale_scan_age_days: int = 60
+    # Categories excluded from stale_scan. `rule` represents explicit user
+    # directives that may be infrequently exercised but still in force —
+    # deactivating them on recall stats alone is unsafe.
+    stale_scan_excluded_categories: list[str] = Field(
+        default_factory=lambda: ["rule"]
+    )
+
+    # Sleep-cycle cluster_consolidation phase (F027): merge near-duplicate
+    # facts under the same subject into one. Prior code picked top-5 clusters
+    # by size, but prod has accumulating subjects like `lesson_learned`
+    # (164 facts) and `Tim` (36 facts) at the top — the LLM correctly
+    # refuses to merge those, leaving the 11 actually-mergeable small
+    # clusters (3-5 facts each) untouched.
+    # Cap cluster size so accumulating subjects are skipped and small
+    # mergeable clusters get a chance.
+    cluster_consolidation_min_facts: int = 3
+    cluster_consolidation_max_facts: int = 10
+
     # F025 P2-C: Transcript truncation limit for episode summarization
     transcript_max_chars: int = 16000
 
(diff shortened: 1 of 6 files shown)


tfatykhov:
Addressed reviewer's important #1 finding in be07f19:

The new filter `last_recalled_at IS NULL AND recall_count = 0` had the same shape as the original `superseded_by IS NOT NULL` bug — a safe-looking AND that excludes the population we actually want. A fact recalled once two years ago and never since would be permanently exempt.

Replaced with OR semantics on the recall side:
```sql
active = true
AND created_at < cutoff
AND (last_recalled_at IS NULL OR last_recalled_at < cutoff)
AND category NOT IN excluded
```

Catches both:
- Old facts never recalled (obvious case)
- Old facts last recalled long ago (the reviewer's flagged case)

Verified on prod data — still returns 0 today (corpus too young, oldest fact 70 days, recent recall mass) but structurally correct now. Added a 9th test that asserts the source uses the OR shape and does NOT gate on `recall_count` (the buggy pattern).

Other reviewer notes accepted as-is per their guidance: max_facts=10 stays, recency_days=30 stays, no flag-gate (correct call — these are bug fixes not behavior changes), Heart.settings exposure already verified earlier.

tfatykhov:
Pushed test fixes in 78d9fb8: 3 pre-existing tests broke because their mocks didn't supply the new settings as ints/lists (so MagicMock attribute access returned MagicMock, which can't be subtracted via timedelta or used in SQL clauses). Added explicit numeric defaults to `_make_handler` (test_f027_supersession.py) and `_make_settings` (test_procedure_learner.py). 61 tests pass across the 3 affected files. Awaiting CI re-run.

PR #406, opened 2026-05-03
feat(eval): sleep-phase filter dry-run probe

tfatykhov:
## Summary
PR #2 of the sleep eval series. Sibling to PR #404 (passive health monitor) — runs PR #405's new SQL filters directly against the eval-DB snapshot and reports what each phase WOULD select, **without triggering sleep or making LLM calls**.

Catches regressions to the structurally-impossible-filter pattern that PR #404 surfaced and PR #405 fixed.

## Why both probes

| Probe | Reads from | Strength | Weakness |
|---|---|---|---|
| **sleep_cycle_health** (#404) | recent `sleep_completed` events | real prod outcomes | slow — needs nightly cycles |
| **sleep_filter_dryrun** (this PR) | direct SQL against snapshot | deterministic, instant | doesn't catch LLM-side failures |

Together: passive monitor catches "phase produced 0 in prod"; dry-run catches "filter would select 0 in any corpus shape." Different signal, complementary.

## First live run against nous-prod-snapshot

```
[OK]   stale_scan                  candidates=0
       no genuinely stale facts (all old facts recently recalled — healthy housekeeping)

[OK]   cluster_consolidation       candidates=13
       13 eligible clusters in [3, 10] range (2 skipped above cap)
       samples:
         'nous-repo-status' (6 facts)
         'agentic memory research' (6 facts)
         'nous repo' (5 facts)
         ...

[----] procedure_recency           candidates=0
       no successful-reviewed-with-bridge decisions in corpus
```

**Verifies PR #405**:
- stale_scan filter is structurally correct. The 119 old active facts in the snapshot all have recent recall, so genuinely 0 are stale (= healthy). The probe correctly distinguishes this from "filter broken."
- cluster_consolidation cap works exactly as intended. 13 small mergeable clusters surface; 2 mega-clusters (`lesson_learned` 164, `Tim` 36) correctly skipped.
- procedure_recency NO_DATA reveals a separate finding: `snapshot_prod_to_eval.py` doesn't copy `brain.decision_bridge`. Tracking as a follow-up; not a probe issue.

## Verdict logic (mid-PR fix)

First live run had over-strict logic — flagged stale_scan RED because it reasoned "corpus has old facts, filter selects 0, so filter must be broken." But the actual signal is **truly-stale facts exist AND filter selects 0**. Old facts that have been recently recalled are NOT stale; the filter correctly excludes them. Fixed before commit + tested explicitly.

## Test plan
- [x] 12 mock-based unit tests in `tests/test_sleep_filter_dryrun_probe.py`:
  - 4 stale_scan verdicts (YELLOW corpus-too-young / GREEN no-stale-facts / RED regression / GREEN candidates-found)
  - 3 cluster_consolidation verdicts (RED above-cap-only / GREEN eligible / NO_DATA empty corpus)
  - 3 procedure_recency verdicts (RED <5% / GREEN >5% / NO_DATA empty)
  - 2 overall_exit_code (RED blocks strict / YELLOW+NO_DATA don't)
- [x] Live `--strict` exits 0 against snapshot
- [x] `SET default_transaction_read_only = on` defense-in-depth (mirrors F058 + sleep_cycle_health probes)

## Related
- PR #404 — passive sleep cycle health monitor
- PR #405 — fixed the 3 silent failures the monitor found
- This PR — verifies the fixes against snapshot data deterministically

### DIFF
--- a/nous_eval/probes/sleep_filter_dryrun.py
+++ b/nous_eval/probes/sleep_filter_dryrun.py
@@ -1,0 +1,495 @@
+"""Sleep-phase filter dry-run probe.
+
+Sibling to ``sleep_cycle_health`` (PR #404). That probe reads recent
+``sleep_completed`` events and detects which phases produced zero
+output across N consecutive cycles. This probe runs the underlying
+SQL filters directly against the eval-DB snapshot and reports what
+each phase WOULD select — without triggering sleep or making LLM
+calls.
+
+Why a separate probe: the passive monitor measures real prod outcomes
+(slow feedback — needs nightly cycles to accumulate). This one gives
+deterministic synthetic regression coverage that catches the bug
+PR #405 fixed:
+
+  - stale_scan filter must select rows the supersede flow doesn't
+    already deactivate. Original ``active=true AND superseded_by IS
+    NOT NULL`` was structurally impossible. PR #405 replaced it with
+    ``never recalled OR not recalled in cutoff window``. This probe
+    asserts the new SQL produces a non-empty candidate set on
+    aged-enough corpora.
+
+  - cluster_consolidation filter must skip accumulating subjects
+    (lesson_learned 164, Tim 36) that the LLM correctly refuses to
+    merge. PR #405 capped at max_facts=10. This probe asserts the
+    new HAVING clause finds the small mergeable clusters.
+
+  - procedure_learner._check_recency must use a window wide enough
+    that real prod traffic produces eligible decisions. PR #405
+    bumped the default 7→30. This probe counts successful-reviewed
+    decisions in the configured window.
+
+Connects to the eval DB (default 127.0.0.1:5433) READ-ONLY.
+
+Run:
+    NOUS_EVAL_DB_NAME=nous_eval_scratch \\
+    NOUS_EVAL_AGENT_ID=nous-prod-snapshot \\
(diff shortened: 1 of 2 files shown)


tfatykhov:
Addressed all 5 review findings in 7f50a0c:

**#1 Settings drift**: Replaced module-level duplicated defaults with `_settings_defaults()` that reads from a live `Settings()` instance at module load. Drift impossible now — adding a category to `stale_scan_excluded_categories` propagates automatically.

**#2 Docstring lie**: Replaced 'We exec the SAME SQL the sleep handler emits' with an honest acknowledgement that the SQL is hand-ported, plus a pointer to sleep_cycle_health (#404) as the cross-check and a follow-up note about extracting a shared SQL helper.

**#3 Hardcoded 5% threshold**: Added `--procedure-recency-min-pct` CLI arg (default 5.0). Low-traffic corpora can tune without forking.

**#4 Test brittleness**: Replaced substring-matching dict (where 2 keys differed only by trailing whitespace) with a semantic `_classify()` function using reformat-stable predicates.

**#5 Polish**: removed unused `asynccontextmanager` import; fixed the f-string with no interpolation.

Low-severity now() pinning deferred with a code comment — the multi-step verdict reads consistently within milliseconds and the boundary edge doesn't change verdict outcomes (both subqueries shrink together).

12 tests pass; live `--strict` still exits 0 against snapshot.

PR #407, opened 2026-05-03
feat(eval): sleep action-quality audit + F027 per-action instrumentation

tfatykhov:
## Summary
PR #3 of the sleep eval series. Catches the failure mode #404 (passive monitor) and #406 (filter dry-run) both miss: **phase fired AND produced output, but the output was wrong.**

Two parts in one PR:
1. **Instrumentation**: F027 cluster_consolidation now emits per-action `f027_cluster_merge` events (F031 already did). Outcome enum: `merged | llm_refused | rejected_by_admission`.
2. **Audit probe**: samples N events from prod, fetches source-fact content, Sonnet-judges each against a phase-specific rubric, aggregates per-phase quality scores.

## First live run (prod, n=5 F031)

```
[OK]   f031_contradiction_resolution     n=5  correct=5  wrong=0  ambiguous=0  score=100%
```

The agent's applied F031 actions are sound. F027 has no events in prod yet — will start populating after deploy.

## A meaningful methodological finding

First-run version of the rubric scored 5/5 actions as **40%** (3 wrong) — but all 3 "wrong" verdicts were the judge complaining that "raw model output was MERGE but applied was KEEP_BOTH, these differ." Those are correct safety downgrades (the F031 confidence floor + missing-merged-content guard intentionally downgrades unsafe MERGEs to KEEP_BOTH).

Refined the F031 rubric to explicitly tell the judge about the safety floor and ask it to evaluate the **applied** action. Same 5 actions then scored 5/5 (100%). Lesson: LLM-judge rubrics need to encode the system's safety semantics, not just the action enum.

## Why three sleep probes

| Probe | Catches |
|---|---|
| #404 sleep_cycle_health | Phase produces 0 output across N cycles (silent failure) |
| #406 sleep_filter_dryrun | Phase filter would select 0 candidates (structural bug) |
| #407 (this PR) action_audit | Phase fired AND produced output, but output was wrong (LLM quality regression) |

The triangle covers structural / behavioral / quality failure classes.

## Test plan
- [x] 9 mock-based unit tests covering aggregate_quality + --strict gate behavior
- [x] Live audit run against prod produces sensible output (5/5 score)
- [x] Read-only transaction guard via `SET default_transaction_read_only = on`
- [x] No prod runtime impact (instrumentation is fire-and-forget; existing F031 pattern)

## Related
- PR #404 sleep_cycle_health — passive monitor
- PR #405 — fixed the 3 silent failures #404 found
- PR #406 sleep_filter_dryrun — deterministic filter verification
- this PR — action quality audit, completes the triangle

Sleep eval series PR #3 of 4. Remaining: PR #4 reversal-rate longitudinal tracker (low priority — needs weeks of data).

### DIFF
--- a/nous/handlers/sleep_handler.py
+++ b/nous/handlers/sleep_handler.py
@@ -131,6 +131,28 @@ _REFLECTION_SCHEMA: dict[str, Any] = {
     "required": ["patterns", "lessons", "connections", "gaps", "summary", "facts"],
 }
 
+# F031 SAFETY DOWNGRADE SEMANTICS — single source of truth.
+#
+# Used both by ``_phase_resolve_contradictions`` (the actual downgrade
+# logic) and by ``nous_eval/probes/sleep_action_audit.py`` (which
+# formats this into the F031 judge rubric so the judge correctly
+# scores downgraded actions as intentional safety, not mistakes).
+#
+# Keep the floor + the downgrade matrix in sync with the corresponding
+# ``if`` branches in ``_phase_resolve_contradictions`` below. A test in
+# ``tests/test_sleep_action_audit_probe.py`` asserts the audit rubric
+# references this constant so drift fails loud.
+F031_SAFETY_FLOOR_DESCRIPTION = (
+    "The agent has a safety floor: any action with confidence below "
+    "0.7 OR a MERGE without merged_content is automatically downgraded "
+    "to KEEP_BOTH before being applied. This is INTENTIONAL safety "
+    "behavior, not a bug — KEEP_BOTH preserves both facts and is "
+    "reversible. When you see 'applied=KEEP_BOTH (downgraded from "
+    "MERGE)', evaluate the APPLIED action (KEEP_BOTH) — i.e., would "
+    "keeping both facts be acceptable here? — not the raw model output."
+)
+
+
 # F031 prompt rewritten 2026-04-30 after the synthetic eval
 # (reports/f031_resolution_eval.md) measured 33% accuracy and a strong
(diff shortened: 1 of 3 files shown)


tfatykhov:
Addressed all 5 review findings in b4c2526 — both blockers + the 3 worth-flagging items.

**#1 Rubric drift (BLOCKING)**: extracted `F031_SAFETY_FLOOR_DESCRIPTION` constant in sleep_handler.py. Audit imports + .format()s it into the rubric instead of duplicating prose. New test `test_audit_imports_safety_floor_from_sleep_handler` asserts the wiring; future PR removing/renaming the constant fails loud.

**#5 Truncation hidden (BLOCKING)**: F027 prompt now emits an explicit "(NOTE: showing 6 of 12 source facts ... mark 'ambiguous' if correctness depends on facts not shown)" when truncated. Judge can opt out instead of scoring partial.

**#2 NaN guard masked sample-size regressions**: added `ambiguity_rate` to aggregate; `overall_exit_code` now skips gating when ambiguity_rate > 0.5 (not just NaN). The 3-correct/0-wrong/47-ambiguous case now skips strict gating instead of silently passing 100%.

**#3 Unbounded create_task**: added `_pending_emits: set` to SleepHandler. Helper does `.add(task) + add_done_callback(discard)` — standard idiom, prevents GC mid-flight + caps task pileup.

**#6 Two connections, no error path**: collapsed to single connection. Sequential queries; no pool needed; one less leak path.

5 new tests; total 13 (was 9). Live audit confirmed working end-to-end.

PR #410, opened 2026-05-03
fix(F027): rewrite cluster_merge prompt — fix the 20% audit verdict

tfatykhov:
## Summary
PR #407's action audit found F027 cluster_merge quality at **20% (1 correct / 4 wrong)** — the LLM refused 4 of 5 small mergeable clusters that the judge said *should* have been merged (coherent topics like \`claude-code-delegation\`, \`nous-repo-status\`, \`agentic memory research\`).

This PR rewrites the F027 prompt in F031's structured-decision-tree style (which scored 100% in the same audit).

## Diagnosis: prompt was bare

Original system: \`"You are a memory consolidation system. Merge related facts into one."\`
Original user: \`"Subject: X | Facts: ... | Create a single consolidated fact."\`

No decision tree, no examples, no permission to lose detail, no structured refusal path. The LLM defaulted to refusing on accumulating subjects (chronological updates, multi-aspect descriptions) because they aren't synonyms.

## Fix

**1. New \`_CLUSTER_MERGE_PROMPT\` — explicit decision tree**:
- Step 1 coherence test (same entity? → refuse with entity-collision reason)
- Step 2 merge feasibility (1-3 sentence summary possible? → refuse with contradiction reason)
- 4 worked examples (good merge: chronological / multi-aspect; good refuse: entity collision / irreconcilable contradiction)
- Default-lean: prefer merge over refuse for coherent topics — admission gate catches over-eager merges, but refused-but-mergeable clusters bloat memory forever

**2. Schema update**: explicit \`should_merge: bool\` + \`refuse_reason: str\` fields. Previous empty-content-as-refusal was implicit and brittle.

**3. Per-action event** now includes \`refuse_reason\` so the audit can judge the refusal's quality.

**4. Audit rubric** updated to surface \`refuse_reason\` to the judge and distinguish "justified refusals" from "abandoned coherent topics."

## Verification path (the eval framework)

After deploy:
1. Trigger sleep cycle (POST /sleep/trigger)
2. Re-run \`nous_eval.probes.sleep_action_audit\`
3. Expect F027 \`quality_score\` to rise from 20% toward F031's 100% baseline

LLM-prompt iteration is noisy on small samples — take median of 2-3 audit runs as the real signal.

## Test plan
- [x] 14 existing audit tests pass
- [x] Smoke imports clean (\`_CLUSTER_MERGE_PROMPT\` has 3 decision steps, schema has new fields, audit rubric mentions refuse_reason)
- [x] Will verify against live audit once deployed

## Related
[...]

PR #411, opened 2026-05-03
fix(admission): bypass cluster_consolidation + contradiction_resolution merges

tfatykhov:
## Summary
PR #410's F027 prompt rewrite worked (LLM merges 4 of 5 clusters, up from 0 of 5 baseline). But the action audit then revealed admission was rejecting 3 of those merges as "redundant" — composite scores 0.52-0.59 vs prod threshold 0.60.

Audit judge said 2 of 3 rejections were **wrong** — the merges were substantively useful consolidations that admission incorrectly tagged as redundant.

## Root cause

Cluster_consolidation by design **restates existing fact content** as a single compressed fact. Admission's novelty score is therefore naturally low for these — admission would always penalize correct consolidations as "you already have this content elsewhere", which is exactly the redundancy the merge is trying to eliminate.

Same logic applies to F031 contradiction_resolution MERGE actions (currently surviving admission only because their confidence is hardcoded to 0.8; if that ever drifts they hit the same trap).

## Fix

Add both `cluster_consolidation` and `contradiction_resolution` to the default `bypass_sources` list. These are post-LLM consolidation operations, not new content — admission's quality filter doesn't apply.

Safety preserved by the **supersede chain**: if a merge turns out wrong, the source facts remain reachable via `superseded_by` links and can be re-activated.

## Verification path (eval framework)

After deploy:
1. Trigger sleep cycle (POST /sleep/trigger)
2. Re-run `nous_eval.probes.sleep_action_audit`
3. Expect F027 quality_score to rise from 50% (today, 2 merged + 3 false-rejected) toward ~80-100% (admission stops rejecting valid merges)

## Test plan
- [x] 2 new tests in `tests/test_admission.py::TestBypass`:
  - `test_cluster_consolidation_bypasses` — verify bypass triggers even with `max_existing_similarity=0.95` (the case admission penalizes hardest)
  - `test_contradiction_resolution_bypasses` — same, for F031 MERGE
- [x] 7 bypass tests pass total
- [x] Existing admission behavior unchanged for all other sources

## Related
The eval-find-then-fix loop continuing into a third iteration:
- PR #404 found F027 produced 0 output
- PR #405 fixed the cap
- PR #410 fixed the LLM-refusal pattern (20% → 50%)
- THIS PR fixes the admission false-rejections (50% → expected 80%+)

### DIFF
--- a/nous/heart/admission.py
+++ b/nous/heart/admission.py
@@ -99,6 +99,20 @@ class AdmissionConfig:
             "censor",
             "supersede",
             "contradict",
+            # Post-LLM consolidation operations bypass admission. By
+            # design these merge EXISTING fact content into a single
+            # compressed restatement, so the novelty score is naturally
+            # low — admission would reject them as "redundant" when the
+            # whole point is to compress redundancy. The supersede chain
+            # preserves the originals if a merge turns out wrong, so
+            # bypassing admission here is safe.
+            #
+            # Caught 2026-05-03: action audit (PR #407) found
+            # admission rejected 3 of 5 F027 merges that the judge
+            # said were correct consolidations of coherent topics
+            # (composite scores 0.52-0.59 vs prod threshold 0.60).
+            "cluster_consolidation",
+            "contradiction_resolution",
         ]
     )
     utility_llm_enabled: bool = True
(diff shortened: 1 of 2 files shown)


tfatykhov:
Addressed both polish items in ceedc21:

- **Safety claim softened**: comment now says "originals remain reachable via superseded_by link" + acknowledges "no public reactivate API yet — recovery is currently manual SQL or re-learning." That's the honest story.
- **PR cite fixed**: was #407, should be #410 (the F027 prompt rewrite cycle whose audit revealed the false-rejections).

7 bypass tests still pass. Ready to merge once CI clears.

PR #412, opened 2026-05-03
fix(F031): preserve supersede chain on MERGE (Codex finding on PR #411)

tfatykhov:
## Summary
Codex review of PR #411 surfaced a **pre-existing data-integrity bug** in F031 MERGE: the merged fact's UUID was never captured and never linked to the originals via `superseded_by`. Every successful F031 MERGE has been silently orphaning the source facts since the code was written.

## The bug

```python
# sleep_handler.py — F031 MERGE branch (BEFORE this fix)
await self._heart.learn(FactInput(...))     # ← merged fact created
                                             #   but result DISCARDED
await self._heart.deactivate_fact(fact1_id)  # ← active=False, no link
await self._heart.deactivate_fact(fact2_id)  # ← active=False, no link
```

Result: originals end up `active=False` with `superseded_by=NULL` — chain broken, recovery only via manual SQL.

## Why this matters now

PR #411 added `contradiction_resolution` to `bypass_sources` with the safety justification *"the supersede chain preserves originals if a merge turns out wrong."* Codex correctly noted: that claim was false — the chain wasn't being set. So the bypass + bug stacked into a real risk.

## Fix

Mirror F027's pattern (which has always done this correctly):
1. Capture `merged_detail = await self._heart.learn(...)`
2. Defensive: skip if `FactRejected` (shouldn't happen with bypass, but guard)
3. Single session: set `superseded_by = merged_detail.id` AND `active = False` on **both** originals atomically
4. Drop the bare `deactivate_fact()` calls

Result: every successful F031 MERGE now leaves originals reachable via `superseded_by` — and PR #411's safety claim becomes literally true.

## Test

`test_f031_merge_preserves_supersede_chain` uses `inspect.getsource()` to pin the contract:
- MERGE branch must capture `merged_detail`
- MERGE branch must reference `superseded_by`
- MERGE branch must NOT use `deactivate_fact(fact1_id)` (the orphaning pattern)

A future PR can't silently re-introduce the bare-deactivate pattern. Mirrors the inspect-based assertions from PRs #408/#409.

## Test plan
- [x] 3 MERGE-related tests pass (1 new + 2 existing)
- [x] Smoke imports clean
- [x] Will verify in prod via next sleep cycle's audit

## Related
Codex review on PR #411 caught this. The eval-find-then-fix loop continuing into a fourth iteration on F027/F031:
- #404 found F027 silent failure
- #405 fixed cap (LLM still refusing)
- #410 fixed prompt (admission rejecting)
- #411 fixed admission bypass (chain not preserved)
- THIS PR fixes the F031 chain (data integrity restored)

### DIFF
--- a/nous/handlers/sleep_handler.py
+++ b/nous/handlers/sleep_handler.py
@@ -894,21 +894,56 @@ class SleepHandler:
                         # case); but keep a defensive check just in case.
                         if merged_content:
                             try:
-                                await self._heart.learn(FactInput(
-                                    subject=pair.get("subject", None),
-                                    content=merged_content,
-                                    source="contradiction_resolution",
-                                    confidence=0.8,
-                                    category=pair.get("category", None),
-                                ))
-                                await self._heart.deactivate_fact(fact1_id)
-                                await self._heart.deactivate_fact(fact2_id)
-                                sleep_stats["contradictions_resolved"] += 1
-                                sleep_stats["facts_created"] += 1
-                                logger.info(
-                                    "F031 resolve: MERGE — combined %s + %s into new fact (%.2f confidence)",
-                                    fact1_id, fact2_id, confidence,
+                                merged_detail = await self._heart.learn(
+                                    FactInput(
+                                        subject=pair.get("subject", None),
+                                        content=merged_content,
+                                        source="contradiction_resolution",
+                                        confidence=0.8,
+                                        category=pair.get("category", None),
+                                    )
                                 )
+                                # PR #411 follow-up (Codex P1): F031
+                                # MERGE used to discard the learn()
+                                # result and call deactivate_fact() on
+                                # both originals — leaving them
+                                # active=False with NO superseded_by
+                                # link. That orphaned every successful
+                                # MERGE's source facts (chain broken,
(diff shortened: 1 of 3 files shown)


tfatykhov:
Addressed both cheap items in 6ce7c83:

**Defensive self-supersede check** — added `if orig_id == merged_detail.id: continue` to both F031 (this PR's main change) AND F027 (which had the same shape). One comparison per iteration; removes a class of "what if learn ever changes" bugs.

**FactRejected.explanation** — verified the attribute exists in `nous/heart/schemas.py:148`. Existing code is correct as-is.

**Atomicity gap (two-transaction window)** — flagged but pre-existing in F027. Not introducing it; just propagating the existing pattern. Tracked as a separate concern per reviewer's note: a `supersede_within_session` helper would close for both. Filed mentally as follow-up.

**Async behavior test** — kept the source-pin test. The reviewer suggested pairing it with one async DB-backed test for the concrete attribute writes. That's a good idea but adds significant test infra; the source-pin catches the specific Codex-flagged regression and matches the pattern from PRs #408/#409 (also inspect-based). Filing as a follow-up if F031 ever sees more changes.

6 merge/contradiction tests pass.

PR #434, opened 2026-05-22
F065 Phase 2: pre_turn hub-shift autosurface + sleep prune

tfatykhov:
## Summary

Wires the F065 hub-shift autosurface into \`CognitiveLayer.pre_turn\` and adds the sleep-handler retention prune for \`brain.graph_hub_snapshots\`. Both behaviors are **gated dormant** by default — this PR adds the integration path; flipping the flag in prod is a separate ops PR.

## Files

- \`nous/cognitive/layer.py\` — new \`_compute_hub_shift_notice()\` + pre_turn integration (gated by \`graph_hub_autosurface_enabled\`)
- \`nous/handlers/sleep_handler.py\` — new \`_phase_prune_hub_snapshots()\` phase
- \`nous/config.py\` — \`graph_hub_snapshot_retention_days\` field relaxed to \`ge=0\` so \`0\` disables the prune
- \`tests/test_f065_pre_turn_autosurface.py\` — 6 new tests

## Design decisions

**Synchronous, not asyncio.create_task.** \`HubSnapshotManager.record_snapshot\` has its own try/except + WARN log; the fire-and-forget pattern was hedging against a blocking risk that doesn't materialize once the manager owns its error handling. Synchronous lets us inject the notice into the same-turn system prompt without one-turn lag. The whole computation is wrapped in a try/except at the call site too — pre_turn never fails on a diagnostic.

**Session plumbing.** \`_compute_hub_shift_notice\` takes an optional session, matching the existing pattern (\`Heart.recall\`, \`Brain.top_hubs\`). pre_turn passes its session in.

**Phase placement.** Prune runs between \`prune_dead_edges\` and \`generalize\` in the sleep cycle.

## Rollout

1. **This PR** — lands the integration with both flags dormant. No behavior change in prod.
2. **Ops PR (separate)** — flip \`NOUS_GRAPH_HUB_AUTOSURFACE_ENABLED=true\` after manual review of a handful of real notice outputs. Watch the rate (\< 2× per session is the spec target).
3. **F051 neutrality check** — should be byte-identical MRR between flag-on and flag-off (this PR's logic doesn't touch \`recall_deep\` ranking).

## Test plan

- [x] \`uv run pytest tests/test_f065_pre_turn_autosurface.py\` — 6/6 pass
- [x] All earlier F065 tests still pass (storage, writer-coverage, pipeline-penalty, top_hubs, hub_shift)
- [x] 7 pre-existing cognitive_layer test failures verified to reproduce on main (unrelated)

### DIFF
--- a/nous/brain/hub_snapshots.py
+++ b/nous/brain/hub_snapshots.py
@@ -55,6 +55,67 @@ class HubSnapshotManager:
                 return await self._get_latest_inner(node_ids, s)
         return await self._get_latest_inner(node_ids, session)
 
+    async def get_latest_top_n(
+        self,
+        top_n: int,
+        session: AsyncSession | None = None,
+    ) -> dict[UUID, GraphHubSnapshot]:
+        """F065: return the most-recent snapshot for each node that was
+        in the top-N at last capture (rank IS NOT NULL AND rank <= top_n).
+
+        Used by pre_turn alongside ``get_latest`` for the live hubs —
+        the combined set lets ``detect_rank_shifts`` fire 'left top-N'
+        notices for nodes that dropped out (Codex review P1, 2026-05-22).
+        Without this, the lookup is biased toward current live hubs and
+        we never emit the 'left' branch of the shift signal.
+        """
+        from sqlalchemy import desc, func, select as _select
+
+        if session is None:
+            async with self._db.session() as s:
+                return await self._get_latest_top_n_inner(top_n, s)
+        return await self._get_latest_top_n_inner(top_n, session)
+
+    async def _get_latest_top_n_inner(
+        self,
+        top_n: int,
+        session: AsyncSession,
+    ) -> dict[UUID, GraphHubSnapshot]:
+        from sqlalchemy import desc, func
+
+        subq = (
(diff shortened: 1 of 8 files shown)


tfatykhov:
@codex review

tfatykhov:
@codex review — addressed P1 (prior-snapshot now includes prior-top-N members) and P2 (entered resolved by node_id, not label). Added test_left_top_n_emits_notice.

tfatykhov:
@codex review — addressed P1 follow-up. `get_latest_top_n` now ranks against the absolute-latest snapshot, and `_compute_hub_shift_notice` writes a rank=NULL snapshot for every 'left' notice so the next pre_turn doesn't re-emit. New regression test `test_left_top_n_persisted_no_repeat_notice` covers the loop.

tfatykhov:
@codex review — addressed both P2 findings. record_snapshot now uses SAVEPOINT on shared sessions to prevent PendingRollbackError poisoning, and format_hub_shift_block takes top_n so the rendered text matches the configured threshold. Two new regression tests cover both paths.

PR #435, opened 2026-05-22
F066.1 Phase 1.5 — LLM-based fix-node dispatch (tool-use constraint)

tfatykhov:
## Summary

Adds the LLM-based fix-node dispatcher promised by the F066.1 spec's "Phase 1 free-form LLM dispatch" item. Phase 1 shipped a rule-based dispatcher as a deliberate scope cut; Phase 1.5 fulfills the spec while keeping the Phase 1 rule-based path as a fallback.

## Selection precedence

\`\`\`
settings.dag_fix_llm_dispatch_enabled AND llm_client wired
    → choose_action_llm (Phase 1.5)
        LLM failure (timeout, parse error, unsupported action)
            → fall back to choose_action (Phase 1)
flag off OR no llm_client
    → choose_action directly (Phase 1)
\`\`\`

## Tool-use constraint

The LLM is forced (via \`tool_choice={"type":"tool","name":"choose_fix_action"}\`) to return a tool_use block. The tool's input schema constrains \`action\` to an enum populated verbatim from \`fix_actions\`, so the model cannot pick an action the fix node didn't authorize. \`amended_prompt\` is required when \`action='retry_with_amended_prompt'\`.

## Files

- \`nous/dag/fix_executor.py\` — new \`choose_action_llm\` function + tool schema + prompt builder
- \`nous/dag/orchestrator.py\` — \`__init__\` accepts \`llm_client=None\`; \`_try_fix_failed_nodes\` tries LLM dispatch first when flag is on, falls back to rule-based on failure; \`retry_with_amended_prompt\` action now overwrites \`parent.instructions\` with \`outcome.amended_prompt\`
- \`nous/config.py\` — 3 new settings: \`dag_fix_llm_dispatch_enabled\` (default False), \`dag_fix_llm_model\` (default Haiku), \`dag_fix_llm_timeout_seconds\` (default 10s)
- \`tests/test_f066_1_llm_dispatch.py\` — 11 new tests

## Test plan

- [x] \`uv run pytest tests/test_f066_1_llm_dispatch.py\` — 11/11 pass
- [x] \`uv run pytest tests/test_f066_1_fix_stage.py\` — 25/25 still pass (Phase 1 contract intact)
- [x] \`uv run pytest tests/test_dag_orchestrator.py tests/test_dag_store.py\` — 69/69 pass

## Rollout

1. **This PR** — lands the integration dormant (flag default false).
[...]

PR #433, opened 2026-05-22
F066.1 Phase 1 implementation — fix-stage node type + skipped state

tfatykhov:
## Summary

Phase 1 of F066.1 per the spec at \`docs/features/F066.1-dag-fix-stage.md\` and the plan at \`docs/superpowers/plans/2026-05-23-f066.1-dag-fix-stage-impl.md\`.

## Ships

- New \`fix\` node type, \`skipped\` terminal state, \`on_failure\` edge type
- Migration 048: extends \`dag_nodes\` CHECK constraints + adds 5 fix-stage columns
- \`DAGCreateRequest\` validator: enforces fix-node structural invariants
- Orchestrator hook \`_try_fix_failed_nodes\` (runs BEFORE \`_propagate_failures\`)
- \`fix_executor.choose_action\` rule-based dispatcher
- 24 new tests, all green; 88 pre-existing DAG tests unchanged

## Deviation from spec

Phase 1 ships rule-based dispatch instead of LLM-based free-form dispatch (the spec's Phase 1 target). Rationale: the value is the state-machine surface; LLM dispatch requires standing up an LLM-call mechanism inside the orchestrator, which is a larger change. Rule-based delivers the F065-class repair scenario with ~30 lines and zero new infrastructure. LLM-based dispatch is queued as Phase 1.5 follow-up that swaps \`choose_action\`.

## Plan-review P0 fixes addressed

Internal multi-team review of the plan flagged 3 P0s; all addressed in this PR:

- **P0-1**: fix nodes auto-dispatched on tick 1 → fix nodes are now stored with status='pending' regardless of wave, AND \`_find_ready_nodes\` skips \`node_type == 'fix'\`.
- **P0-2**: \`'skipped'\` was not in \`_TERMINAL\` → added.
- **P0-3**: \`_find_ready_nodes\` used \`completed\` only as the resolved-predecessor set → introduced \`_RESOLVED = {completed, skipped}\` so \`skip_and_continue\` actually unblocks successors.

## Test plan

- [x] \`uv run pytest tests/test_f066_1_fix_stage.py -v\` — 24/24 green
- [x] \`uv run pytest tests/test_dag_orchestrator.py tests/test_dag_store.py tests/test_dag_concurrency_caps.py\` — 88/88 green (no regression)
- [ ] CI: Postgres-backed migration applies cleanly

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #439, opened 2026-05-23
fix(F066.1): expose fix nodes through dag_create tool surface

tfatykhov:
## Summary
Closes the silent-drop gap between F066.1's shipped orchestrator/schema (PRs #433 + #435) and the \`dag_create\` tool surface in \`nous/api/tools.py\`. Without this fix, fix-stage recovery is unauthorable via normal agent flow — the Anthropic tool-use JSON-schema validator rejects \`\"type\": \"fix\"\` at the API edge, and even bypassing the LLM, the handler silently drops the four fix-specific fields.

### 1. Tool schema enums missing values (rejected by Anthropic API)
\`\`\`diff
- "type": {"enum": ["subtask", "check", "gate", "callback"]}
+ "type": {"enum": ["subtask", "check", "gate", "callback", "fix"]}
- "edge_type": {"enum": ["dependency", "cancel_cascade", "context_flow"]}
+ "edge_type": {"enum": ["dependency", "cancel_cascade", "context_flow", "on_failure"]}
\`\`\`

### 2. Handler dropped fix-specific fields
\`DAGNodeSpec(...)\` construction at \`tools.py:2565-2578\` only forwarded a fixed set of fields. The schema-level fields \`parent_node\`, \`fix_actions\`, \`max_fix_attempts\`, \`expected_modes\` (F066.1) and \`stall_timeout_seconds\` (F064.1) were silently dropped — so even direct MCP calls bypassing the LLM produced a broken fix node.

Handler now threads every field DAGNodeSpec accepts, with \`max_fix_attempts\` conditionally forwarded so its pydantic default (1) applies when the caller omits it.

### 3. Schema didn't document the surface
Added per-field \`description\` blocks so the LLM knows when each fix-node field applies (only when \`type=='fix'\`), what \`fix_actions\` are allowed, and how Phase 1 dispatcher rules interact with them.

## Test plan
- [x] \`test_dag_create_authors_fix_node\` — end-to-end: create a DAG with a fix node, assert all four fix-specific fields landed in the persisted node.
- [x] \`test_dag_create_fix_node_missing_parent_node_errors\` — missing \`parent_node\` on a fix node now surfaces an error (vs the prior silent-success-with-broken-fix-node behavior).
- [x] \`test_dag_create_tool_schema_advertises_fix_surface\` — JSON-schema regression guard: \`fix\` in node.type.enum, \`on_failure\` in edge.edge_type.enum, and the four fix-node fields present in node.properties.
- [x] \`pytest tests/test_dag_tools.py tests/test_dag_schemas.py tests/test_dag_store.py tests/test_dag_orchestrator.py\` — 117/117 passing.

### DIFF
--- a/nous/api/tools.py
+++ b/nous/api/tools.py
@@ -2739,7 +2739,7 @@ def register_dag_tools(
                         # when type='fix'; the DAGCreateRequest validator
                         # enforces parent_node + non-empty fix_actions for
                         # fix nodes and rejects these fields on non-fix nodes.
-                        "parent_node": {"type": "string", "description": "F066.1 (type='fix' only): name of the node this fix attaches to. Fires when the parent transitions to 'failed'. The matching 'on_failure' edge must have this node's name as from_node."},
+                        "parent_node": {"type": "string", "description": "F066.1 (type='fix' only): name of the node this fix attaches to. Fires when the parent transitions to 'failed'. The matching 'on_failure' edge MUST point from the parent to this fix node — i.e. from_node = (this parent_node value), to_node = (the fix node's own name). Pointing the edge the other direction fails validation."},
                         "fix_actions": {
                             "type": "array",
                             "items": {"type": "string", "enum": ["retry_as_is", "retry_with_amended_prompt", "mark_unrecoverable", "skip_and_continue"]},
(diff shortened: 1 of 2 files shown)


tfatykhov:
@codex review

tfatykhov:
@codex review — addressed P2: parent_node description now explicitly states from_node = parent_node value, to_node = fix node's name, instead of the ambiguous "this node's name" phrasing.

tfatykhov:
@codex review (re-trigger — please review commit da50151)

PR #440, opened 2026-05-23
spec(F068): censor action tiers — steer/refuse/abort (renamed from F067)

tfatykhov:
## Summary

Proposes renaming the censor action vocabulary from `warn | block | absolute` to `redirect | refuse | absolute` and introduces a new **refuse** tier (LLM runs with refusal directive + `suggested_alternative` field).

## Provenance

⚠️ This spec was **reconstructed from memory** — specifically Heart decision `7614dff0` and repo grep validation against current schema. The original draft (referenced in fact `10496ec0`) was never committed and no longer exists on disk. See the *Provenance* section in the spec for details.

## What's in the spec

- **Tier rename rationale** — three failures of current `warn|block|absolute` vocabulary
- **Refuse mechanics** — LLM runs with refusal directive instead of being silently suppressed
- **Loop prevention** — `refusal_censor_ids` field on turn context
- **F031 interaction matrix** — confirms `trigger_action` / `action_instruction` / `unblock_pattern` stay tier-neutral
- **Schema migration SQL** + 3-phase rollout
- **Three open questions** flagged

## Status

- Spec-only — no code changes
- Requires second-opinion review (in flight) before treating as authoritative
- Three OPEN questions need product input

## Related

- Builds on F031 (PR #209) censor middleware
- Replaces never-committed F052/F053 draft


tfatykhov:
## 🔍 Second-Opinion Review — F067 Censor Action Tiers Spec

**Reviewer:** Claude Code (automated second-opinion, job `job-20260523-190859-27ee88e5`)  
**PR:** #440 — F067 censor action tiers spec  
**Date:** 2026-05-23  
**Cost:** $0.667 (claude-opus-4-5)

---

### Verdict: ✅ PASS — fix-then-merge

The spec is coherent, aligns with current code reality (all line references verified against live source), and proposes a reasonable vocabulary upgrade. No P1 blockers. Five P2 items require attention before implementation — one of them (the loop-prevention rationale) is particularly important as it will create implementation ambiguity if unaddressed.

---

### P1 Issues

_None_

---

### P2 Issues (5)

**[P2] Lines 103–113 — Loop prevention rationale describes a loop that cannot occur**

The spec warns of a censor→refusal→re-match→∞ loop. This is architecturally impossible:
- `post_turn` (`layer.py:1061`) runs *after* the model replies — it cannot issue a new prompt.
- `post_turn` is explicitly logging-only (`layer.py:1058`).
- `pre_turn` (`layer.py:754`) checks only `user_input`, not prior model replies.

The `refusal_censor_ids` mechanism solves a non-existent problem.

**Fix:** Either (a) remove `refusal_censor_ids` as unnecessary, or (b) explicitly state this is forward-looking for when post-turn gains enforcement power, and document that precondition.

---

**[P2] Line 145 — "byte-identical behavior" claim contradicts the semantic upgrade**

`"behavior [that] is byte-identical after migration (modulo the block → refuse semantic upgrade)"` — the parenthetical negates the claim entirely. Phase 1 renames `block→refuse` (mechanical, no behavior change); Phase 2 changes behavior. The statement conflates these phases.

**Fix:** Reword to: *"Phase 1 is a rename with no behavior change; Phase 2 introduces the new refuse semantics."* Remove the "modulo" clause.

---

**[P2] Lines 165–166 — Open question 3 (`unblock_pattern` + downgrade stamping) must be resolved before impl**

The spec "leans no" on whether `refusal_censor_ids` gets stamped when `unblock_pattern` downgrades `refuse→redirect`. Deferring this to implementation risks inconsistency.

**Fix:** Commit to "no" and add one sentence: *"Downgrade means the refusal did not occur; no ID to record."*

---

**[P2] Lines 88–92 — Refuse-tier directive is vague on tool access during refusal turn**

Spec says "LLM runs normally with full tools, full memory" but doesn't clarify whether the model can execute destructive tools (e.g. `bash`, `write_file`) while composing a refusal. If `refuse` is meant to prevent action, action tools should be stripped for that turn.

**Fix:** Specify whether `refuse` strips action-tools from the tool list, or whether the directive alone is sufficient (and note the audit path at `layer.py:1078`).

---

**[P2] Lines 108–109 — "`refusal_censor_ids` per turn only" is redundant; data flow is missing**

`TurnContext` is already per-turn by definition — stating "the skip is per-turn, not session-wide" is redundant. More critically: the spec doesn't describe how `refusal_censor_ids` propagates from `pre_turn` into `post_turn` within the same turn.

**Fix:** Add `refusal_censor_ids: list[UUID] = Field(default_factory=list)` to `TurnContext` in the design section, and document the data flow explicitly.

---

### Agreement Hooks (Rebuttals Considered)

1. **On the non-existent loop (P2 #1):** Counter-argument is defensive forward design. Counter fails because the forward-looking intent isn't documented — the spec should make it explicit if that's the rationale.

2. **On tool access during refuse (P2 #4):** Counter-argument is that keeping tools is intentional (e.g. `recall_deep` for policy lookup). Counter *succeeds* if the directive is empirically reliable — the spec should note the audit path at `layer.py:1078`.

---

### Summary

All 5 findings are P2. The most important is the loop-prevention rationale (P2 #1) — 2 sentences of edit would clarify whether `refusal_censor_ids` is dead code or forward-looking design, preventing implementation confusion. The remaining 4 are documentation precision issues.

**Recommendation: fix-then-merge** ✅

tfatykhov:
Pushed a rewrite addressing review feedback. Summary of changes from the original F067 v1 (commit \`e99ce9e\` → \`0ddd590\`):

**Rename: F067 → F068.** F067 was reassigned today to Episode Chunks + Parent-Episode Recall (PRs #441/#443/#444/#445). This spec now lives at \`docs/features/F068-censor-action-tiers.md\`.

**Renamed actions: \`redirect | refuse | absolute\` → \`steer | refuse | abort\`.**
- \`redirect\` overpromised in the no-payload case (steers nothing if no \`trigger_action\` / \`action_instruction\`)
- \`absolute\` was vague (absolute *what*?)
- New names are active verbs that describe what the system *does*

**\`refuse\` is now a real safety boundary, not just a directive.** Action-tools (\`bash\`, \`write_file\`, \`send_file\`, \`spawn_task\`, \`schedule_task\`, \`record_decision\`, \`learn_fact\`, \`create_censor\`) are stripped when a \`refuse\` censor matches, with per-censor \`refuse_keep_tools\` opt-out. The directive alone is a soft constraint; the tool strip is the hard one. Addresses second-opinion P2 #4.

**\`suggested_alternative\` promoted from F068.2 deferred → Phase 1 schema.** Without it, the refuse-tier UX promise has no structured way to surface alternatives.

**Dropped \`refusal_censor_ids\` from Phase 1.** Second-opinion P2 #1 was right — the loop it prevents is architecturally impossible in current code (\`post_turn\` is logging-only). Deferred to F068.4 for when post-turn enforcement lands.

**Fixed codex P1.** Migration constraint name corrected to \`ck_censors_action\` (was \`censors_action_check\`).

**Resolved all 3 open questions inline.**

@codex please re-review — major rewrite addressing all P1 + P2 feedback.

tfatykhov:
Round 2 codex fixes pushed (commit \`06ad999\`). Both P1s addressed at root cause rather than per-tool:

1. **spawn_sync was missing from the denylist** — typed inline counterpart to \`spawn_task\` at \`tools.py:1849\`, would have let a refused turn launch a subtask that performs the blocked action.

2. **\`run_python\` was misclassified as read-only** — already marked state-modifying in \`execution_ledger.py:47\` with comment \"Can call learn_fact() and modify state\". My denylist had it on the wrong side.

**Root-cause fix:** replaced the hand-enumerated denylist with a reference to the canonical \`STATE_MODIFYING_TOOLS\` set in \`nous/cognitive/execution_ledger.py\` — the same list F026's execution ledger maintains. Implementation reuses it directly.

The benefit beyond fixing these two P1s: future state-modifying tools added to the ledger automatically participate in the refuse strip without needing a spec update. F026's contract becomes the safety boundary for F068's refuse tier.

Also added explicit circumvention examples (\`spawn_sync\` subtask launch + \`run_python\` state writes) to the \"why strip tools\" section.

@codex please re-review.

tfatykhov:
Round 3 fixes pushed (\`94d93ef\`). Codex P1 was right — I invented \`STATE_MODIFYING_TOOLS\`; the actual sets are \`WRITE_TOOLS\`, \`EXTERNAL_TOOLS\`, \`IRREVERSIBLE_TOOLS\`.

Two corrections:

1. **Spec now references the actual symbols** — union of \`WRITE_TOOLS ∪ EXTERNAL_TOOLS ∪ IRREVERSIBLE_TOOLS ∪ {\"bash\"}\` (bash treated as always denied since per-command classification isn't viable at strip time).

2. **\`spawn_sync\` is genuinely missing from \`WRITE_TOOLS\`** — pre-existing F026 audit gap from when F062 (typed-spawn-sync) shipped. The spec now requires adding \`spawn_sync\` to \`WRITE_TOOLS\` as a **precondition fix** in the same implementation PR (or a precursor) — first item in the code touch-points list. Without it the canonical-set denylist still has a hole.

The implementation snippet now reads:
\`\`\`python
from nous.cognitive.execution_ledger import (
    WRITE_TOOLS, EXTERNAL_TOOLS, IRREVERSIBLE_TOOLS
)
REFUSE_DENYLIST = WRITE_TOOLS | EXTERNAL_TOOLS | IRREVERSIBLE_TOOLS | {\"bash\"}
\`\`\`

@codex please re-review.

tfatykhov:
Round 4 fix pushed (`3e7cdf9`). Codex was right — one stale `STATE_MODIFYING_TOOLS` reference still lived in the "Why strip tools" rationale paragraph, contradicting the implementation snippet a few lines above. Replaced with `WRITE_TOOLS ∪ EXTERNAL_TOOLS ∪ IRREVERSIBLE_TOOLS`.

The two remaining mentions in the revision history are intentional — they document the bug as part of the spec's own audit trail ("the symbol I referenced...does not exist"). Left in place.

@codex please re-review.

PR #443, opened 2026-05-24
fix(F067): score-rerank when chunks enabled so they reach top-K

tfatykhov:
## Summary

The F067 chunk-recall leg has been silently dead in production since PR #441 was merged.

**The bug**: ``run_recall_pipeline`` appends chunks AFTER the fact stage (preserving legacy stage-order text output). A top-K=10 consumer (the agent) only ever sees the first 10 results = facts. Chunks sit at positions 11+ regardless of their cosine scores.

**Discovery**: while running a prod-shape A/B (chunks ON vs OFF) on cloned prod data, the retrieved snippets were BYTE-IDENTICAL between configs. Chunks were being retrieved (the stage fired, ``acc.chunk_results`` populated) but the appended results never reached the consumer.

**Fix**: pass ``rerank_by_score=getattr(settings, "episode_chunks_enabled", False)`` to ``run_recall_pipeline``. When chunks are enabled, the merged result list re-sorts by score so chunks compete fairly with facts/episodes for top-K slots. When chunks are disabled (current prod default), the flag stays False and ``recall_deep`` text output remains byte-identical to the legacy stage-order snapshot.

## Test plan
- [x] 3 new unit tests pin both invariants (legacy preserved when off, rerank fires when on) + missing-attribute safety
- [x] 30 existing F067 tests still passing (test_retrieval_pipeline, test_recall_parent_episodes, test_fetch_parent_episodes, test_chunking)
- [x] Validated on prod-shape A/B harness: ``chunks_on`` win share went from undetectable (0% A/B signal pre-fix, byte-identical retrieval) to 100% of decisive verdicts (2/2 decisive, 28 ties on N=30 — chunks help when they differ, never hurt)

### DIFF
--- a/nous/api/retrieval_pipeline.py
+++ b/nous/api/retrieval_pipeline.py
@@ -617,6 +617,13 @@ def _heart_graph_to_pipeline(
             score=_f065_provenance_penalty(n, n.edge_weight, decay, settings),
             source="graph_expanded",
             edge_relation=n.edge_relation,
+            # Stage origin tag — the formatter uses this to bucket
+            # graph_expanded results into the "Graph-Connected Decisions"
+            # (heart-side) vs "Brain Decisions" (brain-side) sections
+            # without relying on positional inference. Required because
+            # ``rerank_by_score`` can globally re-sort the result list,
+            # which would otherwise break the position-based heuristic.
+            metadata={"stage_origin": "heart_graph"},
         )
         for n in heart_graph
     ]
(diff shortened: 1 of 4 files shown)


tfatykhov:
@codex

tfatykhov:
@codex please re-check — addressed both P2 issues:

1. **First P2** (rerank gating): rerank_by_score now mirrors the exact chunk-fetch condition (`episode_chunks_enabled AND (search_all OR "fact" in search_types)`), so non-chunk recall paths (e.g. memory_types=["decision"]) keep legacy stage-order output even with the flag on.

2. **Second P2** (formatter positional invariant): graph-expanded results are now tagged with `metadata["stage_origin"]` ("heart_graph" / "brain_graph") at the to_pipeline conversion site. The formatter buckets via the metadata tag instead of inferring origin from position, so bucketing survives the global score-rerank.

Regression test added that constructs a result list where the heart-side graph decision sits AFTER a brain decision after a score sort, and pins that it still lands in the correct section. All 20 F067 tests pass, including the byte-identical snapshot.

PR #449, opened 2026-05-26
fix(F055): residual record_surfaced wrote loaded_at=None, 500'd /status

tfatykhov:
## Summary

Prod hotfix for a pre-existing F055 bug (unrelated to F070). Tim observed the prod nous container 500-ing on \`/status?dashboard=true\` and WARN-then-error on \`pre_turn\` working-memory init.

\`\`\`
pydantic_core._pydantic_core.ValidationError: 1 validation error
for WorkingMemoryItem
loaded_at
  Input should be a valid datetime
  [type=datetime_type, input_value=None, input_type=NoneType]
\`\`\`

## Root cause

\`ResidualActivator.record_surfaced\` (nous/heart/residual_activation.py:248) writes residual items into \`WorkingMemory.items\` JSONB with \`"loaded_at": None\` — but \`WorkingMemoryItem.loaded_at\` in \`heart/schemas.py:305\` is a required \`datetime\` (non-optional). Every subsequent \`_to_state\` call rejects the row.

\`/status?dashboard=true\` returns working_memory state in its response, so it 500s. \`pre_turn\` also calls \`get_or_create_working_memory\` and fails the same way. Both stuck in a tight 500 loop until F049's TTL sweep eventually evicts the bad rows (~24h).

## Fix (two parts)

**1. Writer** — \`residual_activation.py:248\` now uses \`datetime.now(UTC).isoformat()\`. Residual surfacing IS a load event into WM, so this is semantically correct AND pydantic-compatible.

**2. Reader defense** — \`working_memory.py::_to_state\` pre-processes \`raw_items\` and coerces \`loaded_at=None\` to \`now()\` with a WARN log naming session/agent/ref_id. Existing prod JSONB rows self-heal on read. No migration needed.

## Test

\`tests/test_working_memory.py::test_to_state_tolerates_null_loaded_at_residual_row\` injects a row mimicking the historical F055 writer and asserts: (a) no exception, (b) loaded_at coerced to real datetime, (c) other fields preserved, (d) WARN log emitted.

## Test plan

- [ ] @codex review
- [ ] Merge + restart prod nous
- [ ] Confirm \`/status?dashboard=true\` returns 200
- [ ] Confirm \`pre_turn\` warnings stop in nous-agent logs

### DIFF
--- a/nous/heart/residual_activation.py
+++ b/nous/heart/residual_activation.py
@@ -40,6 +40,7 @@ from __future__ import annotations
 import logging
 import math
 from dataclasses import dataclass
+from datetime import UTC, datetime
 from typing import TYPE_CHECKING
 from uuid import UUID
 
@@ -238,6 +239,14 @@ class ResidualActivator:
             max_score = max((s for _id, _t, s in surfaced), default=0.0)
             if max_score <= 0:
                 return
+            # ``loaded_at`` is a required ``datetime`` on WorkingMemoryItem
+            # (heart/schemas.py:305). Previously this was written as ``None``,
+            # which the JSONB roundtrip via ``_to_state`` rejected with a
+            # pydantic ValidationError, breaking /status?dashboard=true and
+            # the pre_turn working-memory init in prod. Use the surface time
+            # as the load time — semantically correct since residual surfacing
+            # IS a load event into WM.
+            now_iso = datetime.now(UTC).isoformat()
             entries: list[dict] = []
             for node_id, node_type, score in surfaced[:top_k]:
                 entries.append({
@@ -245,7 +254,7 @@ class ResidualActivator:
                     "ref_id": str(node_id),
                     "summary": f"residual {node_type}",
                     "relevance": float(score) / max_score,
-                    "loaded_at": None,
+                    "loaded_at": now_iso,
                     "activation": float(score) / max_score,
                     "last_surfaced_turn": current_turn,
                 })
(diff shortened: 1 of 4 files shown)


tfatykhov:
@codex please review

tfatykhov:
Added a second related fix while auditing WM-vs-recent-changes interactions.

**`MemoryType` was missing `chunk`** — F055's `record_surfaced` passes `r.type` through verbatim from every `PipelineResult`, and since F067 those can be `chunk`. But `MemoryType` in `heart/schemas.py:15` was still the pre-F067 5-element set, so any recall_deep where top-K included a chunk wrote a row `_to_state` then rejected with the same 500-loop symptom — different error string (`Input should be 'fact', 'procedure', 'decision', 'censor' or 'episode'`) but same root cause family.

Fix: add `"chunk"` to `MemoryType`. Audited the other consumer (`RecallResult.type`) — heart.recall never emits chunks (chunks come via the separate F067 leg into `acc.chunk_results`), so widening the Literal is a pure superset relaxation. Audited downstream consumers (rest.py:273 dashboard render, cognitive/context.py branches) — no closed-set assumptions.

Regression test: `test_to_state_accepts_chunk_type_residual_row`.

PR now covers both WM-vs-F055 bugs Tim's prod 500s exposed:
1. `loaded_at = None` (commit 1)
2. `type = "chunk"` rejected by MemoryType Literal (commit 2)

Plus reader-side defensive coercion for #1 so existing bad JSONB rows self-heal. Re-review please.

tfatykhov:
@Codex - review

PR #448, opened 2026-05-25
feat(F070): chunk-aware sleep consolidation (v1: edges only)

tfatykhov:
## Summary

Closes the orphan-chunks gap. 2026-05-25 audit found 1,775 graph edges in
`brain.graph_edges`, **all** fact↔fact / procedure↔procedure — zero
involving chunks. Adjacency boost and F022 spreading activation can't
reach chunks until edges exist.

This PR makes chunks a first-class graph node and lets the sleep cycle
densify chunk neighborhoods. Spec: `docs/features/F070-chunk-aware-sleep.md` (#447).

## What this changes

**New node type & relations** (`nous/brain/schemas.py`):
- `NodeType` admits `"chunk"`
- `RelationType` admits `"part_of"` and `"summarized_by"`

**Wired into densifier** (`nous/brain/graph_densifier.py`, `_entity_config.py`):
- `chunk → episode` `[part_of]` — always created (provenance FK as edge)
- `chunk → fact` `[summarized_by]` — same-episode, gated on `graph_threshold_chunk_fact`
- `chunk → chunk` `[related_to]` — sequential adjacency always linked; non-adjacent gated on `graph_threshold_chunk_chunk_intra` (intra-episode) / `..._cross` (cross-episode)
- `backfill_orphan_chunks` slotted into `run_backfill_cycle` alongside facts/decisions/episodes/procedures

**Schema migration 051**: extends `ck_edges_{source_type,target_type,relation}` to admit the new values. Drops both the auto-named `graph_edges_*_check` constraints (from `init.sql`) and the `ck_edges_*` names added by migration 016 — fresh-DB compose-up tested.

**Settings** (`nous/config.py`) — all opt-in:
- `NOUS_CHUNK_CONSOLIDATION_ENABLED` (default `False`)
- `NOUS_GRAPH_BACKFILL_MAX_CHUNKS` (default `100`)
- `NOUS_GRAPH_THRESHOLD_CHUNK_FACT` (default `0.55`)
- `NOUS_GRAPH_THRESHOLD_CHUNK_CHUNK_INTRA` (default `0.70`)
- `NOUS_GRAPH_THRESHOLD_CHUNK_CHUNK_CROSS` (default `0.85`)

## Validation

**Smoke on eval DB** (`nous_eval_scratch`, agent `nous-lme-corpus`, 35,804 chunks):
Ran one densifier cycle with the flag on. Result on 30 orphan chunks:

| Relation | Count |
|----------|-------|
[...]

PR #450, opened 2026-05-26
spec(F070.1): cross-episode chunk graph edges

tfatykhov:
## Summary

Spec for the next chunk-graph layer after F070 v1.

F070 v1 (PR #448, merged 2026-05-26) shipped the chunk-edge data layer + Path A retrieval consumer. Empirical result on LME N=60 K=5 per_haystack: F070's chunk edges are **consumed correctly** but the metric doesn't move — the same-episode constraint means within-session corroboration, and session-level recall is structurally insensitive to that.

F070.1 adds the missing piece: **cross-episode edges**, which let a fact-seed surface chunks from *other* sessions via Path A's already-shipped fan-out.

## Scope

- `chunk → fact summarized_by` ACROSS episodes (cosine ≥ 0.75)
- `chunk ↔ chunk related_to` ACROSS episodes (cosine ≥ 0.85, reuses `graph_threshold_chunk_chunk_cross` env var that v1 reserved-but-didn't-use)
- 3 new densifier methods + new orphan finder
- 3 new settings (`graph_threshold_chunk_fact_cross`, `graph_backfill_max_chunks_cross_episode`, `chunk_cross_episode_top_k`)
- Backfill script gets `--mode {same-episode, cross-episode, all}` flag
- No retrieval-side changes — Path A consumes these automatically
- No schema migration — v1's migration 051 already admits the relations
- No CE rerank in v1 (matches F070 v1's pattern; defer to F070.1.1 if eval shows noise)

## Hypothesis

Baseline hit@5 lifts from 0.950 toward 0.96-0.97, concentrated in multi-session + knowledge-update categories. chunks_off probably stays at 0.917 (it doesn't traverse chunks).

If lift fails to materialize, the structural hypothesis was wrong and F070.2 needs to layer something else (multi-hop walk, CE rerank, etc.).

## Review focus

- Threshold defaults (0.75 chunk→fact_cross, reuse 0.85 chunk↔chunk_cross from v1)
- Whether to use CE rerank in v1 (spec says no — explain why)
- The new finder approach (`find_chunks_lacking_cross_episode_edges` vs reusing `find_orphans`)
- Whether cross-episode backfill should be added to `run_backfill_cycle` or stay operator-triggered

## Test plan

- [ ] @codex review the spec
- [ ] If approved, implement on `feat/F070.1-cross-episode-chunk-edges`

### DIFF
--- a/docs/features/F070.1-cross-episode-chunk-edges.md
+++ b/docs/features/F070.1-cross-episode-chunk-edges.md
@@ -1,0 +1,280 @@
+# F070.1 — Cross-episode chunk graph edges
+
+**Status**: spec
+**Depends on**: F070 v1 (merged 2026-05-26, PR #448)
+**Spec author**: Claude + Tim, 2026-05-25
+**Branch when implementing**: `feat/F070.1-cross-episode-chunk-edges`
+
+
+## Problem
+
+F070 v1 wrote 229K chunk graph edges on the LME eval corpus and a smaller
+number on prod, but **every edge is same-episode**: `chunk → fact summarized_by`
+requires `f.source_episode_id = chunk.episode_id`; `chunk → chunk related_to`
+uses sequential adjacency + intra-episode cosine.
+
+That decision was load-bearing for v1's scope (no embedding fan-out across the
+full chunk table). It also means F070's retrieval consumers are
+**structurally inert on session-level recall**. We proved this empirically:
+
+| Eval | chunks_off | baseline |
+|---|---|---|
+| BGE only (no chunk edges consulted) | 0.917 | 0.950 |
+| BGE + Path A (Stage 2b consumes chunk edges) | 0.917 | 0.950 |
+| BGE + adjacency_boost (in-batch reweighting) | 0.917 | 0.950 |
+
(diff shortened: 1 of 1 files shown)


tfatykhov:
@codex please review the spec

tfatykhov:
@codex addressed all three findings:

**P1 — finder needed both edge directions.** Spec now enumerates 3 cross-episode paths (chunk→fact, chunk→chunk where current is source, chunk→chunk where current is target) and the pseudo-SQL covers all three via LEFT JOINs on `heart.episode_chunks` from both `e.source_id` and `e.target_id` sides. Idempotency restored.

**P2 #1 — rollout cmd missing the env flag.** Step 3 now shows `NOUS_CHUNK_CONSOLIDATION_ENABLED=true` inline so an operator following the spec verbatim doesn't no-op the backfill.

**P2 #2 — verification query couldn't prove cross-episode existence.** Step 4 replaced with two episode-aware queries (chunk→fact and chunk→chunk) that JOIN `heart.facts` / `heart.episode_chunks` and filter on `episode_id` mismatch. Both must return > 0 for rollout to be healthy.

Re-review please.

PR #451, opened 2026-05-26
feat(F070.1): cross-episode chunk graph edges

tfatykhov:
## Summary

Implementation of [F070.1 spec](https://github.com/tfatykhov/nous/pull/450) (still under codex review). Closes the structural gap F070 v1 had on session-level recall: v1's same-episode chunk edges don't add new sessions to top-K. Cross-episode edges let a fact-seed surface chunks from OTHER sessions via Path A's existing fan-out.

## What ships

**Code** (+696 LOC across 4 files):
- `find_chunks_lacking_cross_episode_edges` — idempotency finder. Three correlated NOT EXISTS clauses (codex P1: covers chunk→fact AND both directions of chunk↔chunk). Index-friendly: returns count on 35K chunks / 229K edges in <1s.
- `_link_chunk_to_cross_episode_facts` / `_link_chunk_to_cross_episode_chunks` — HNSW LIMIT-K scans excluding the source's own episode, threshold-gated, `extraction_method='inferred'`.
- `backfill_orphan_chunks_cross_episode(max_count)` — top-level entry, idempotent.
- 3 new settings: `graph_threshold_chunk_fact_cross=0.75`, `graph_backfill_max_chunks_cross_episode=2000`, `chunk_cross_episode_top_k=20`. Reuses `graph_threshold_chunk_chunk_cross` from v1.
- `--mode {same-episode, cross-episode, all}` flag on the backfill script.

**Tests** (5 new):
- chunk→fact cross-episode happy path
- chunk↔chunk cross-episode happy path
- threshold gate (high threshold = 0 edges)
- **finder covers both edge directions** (codex P1 regression test)
- idempotent re-run (second call returns 0)

All 5 pass on real Postgres alongside the 8 existing F070 v1 chunk tests (13 total).

## Performance journey

The finder query went through 3 rewrites:
1. `NOT EXISTS` with multi-table `LEFT JOIN` — stalled on 35K chunks × 229K edges
2. `WITH ... UNION CTE` + `NOT IN` — also stalled (materialized intermediate set)
3. **Three separate `NOT EXISTS` clauses correlated on `c.id`** — shipped. Planner short-circuits per row as soon as any one fires.

Verified on real eval-scale data: dry-run on `nous-lme-corpus` (35,804 chunks) returns the cross-episode candidate count in <1 sec.

## Hypothesis

Per the spec, baseline hit@5 lifts 0.950 → 0.96-0.97 on LME per_haystack K=5, concentrated in multi-session + knowledge-update categories. `chunks_off` stays at 0.917 (no chunk traversal). Path A flag (`NOUS_HEART_GRAPH_ALL_TYPES_ENABLED=true`) must be on at eval time to actually consume the new edges.

If the lift doesn't materialize, the F070.1 spec calls out the next levers (multi-hop expansion, graph-feature-aware re-ranker, threshold sweep).

## Test plan

- [ ] @codex review
- [ ] Run cross-episode backfill on eval DB: `... --mode cross-episode`
- [ ] Re-run `scripts/eval/eval_session_level_recall.py --mode per_haystack --k 5` with Path A on
- [ ] Verify on prod (run backfill, sanity-check via the episode-aware verification queries in the F070.1 spec)

### DIFF
--- a/nous/brain/graph_densifier.py
+++ b/nous/brain/graph_densifier.py
@@ -732,6 +732,246 @@ class GraphDensifier:
                 created += 1
         return created
 
+    # ------------------------------------------------------------------
+    # F070.1 — Cross-episode chunk graph edges
+    # ------------------------------------------------------------------
+
+    async def find_chunks_lacking_cross_episode_edges(
+        self,
+        limit: int,
+        session: AsyncSession,
+    ) -> list[tuple[UUID, str, UUID]]:
+        """F070.1: chunks that have at least one edge but no cross-episode one.
+
+        Returns ``(chunk_id, content, episode_id)`` tuples.
+
+        Cross-episode detection covers THREE paths (codex P1 review): chunk
+        may be source OR target of a chunk↔chunk cross-edge, plus chunk→fact.
+        Missing any path would re-process already-linked chunks on every
+        run, defeating idempotency.
+
+        Implementation note: uses three separate ``NOT EXISTS`` clauses
+        correlated on ``c.id``. This lets the Postgres planner short-circuit
+        per row — as soon as ANY one cross-episode path matches for a given
+        chunk, the row is excluded without evaluating the others. The
+        earlier attempts (UNION + NOT IN, big NOT EXISTS with multi-table
+        LEFT JOIN) materialized intermediate sets and stalled on the
+        real eval corpus (35K chunks × 229K edges).
+        """
+        sql = text(
+            """
+            SELECT c.id, c.content, c.episode_id
+            FROM heart.episode_chunks c
+            WHERE c.agent_id = :a
+              AND EXISTS (
+                  SELECT 1 FROM brain.graph_edges e
+                  WHERE e.agent_id = :a
+                    AND ((e.source_id = c.id AND e.source_type = 'chunk')
+                         OR (e.target_id = c.id AND e.target_type = 'chunk'))
+              )
+              -- Path 1: chunk → fact in another episode
+              AND NOT EXISTS (
+                  SELECT 1 FROM brain.graph_edges e
+                  JOIN heart.facts f
+                      ON e.target_id = f.id AND f.agent_id = :a
+                  WHERE e.agent_id = :a
+                    AND e.source_id = c.id
+                    AND e.source_type = 'chunk'
+                    AND e.target_type = 'fact'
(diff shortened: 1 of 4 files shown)


tfatykhov:
@codex please review

tfatykhov:
@codex round-2 P1 + P2 addressed:

**P2 (L770) — chunks without embeddings now excluded.** Added `AND c.embedding IS NOT NULL` to both the densifier finder and the script's count helper. NULL-embedded chunks could never produce cross-episode edges and would otherwise sit in the LIMIT window forever. Regression test `test_cross_episode_finder_excludes_null_embedding_chunks` confirms.

**P1 (L323) — offset pagination prevents early termination.** Threaded `offset` through `find_chunks_lacking_cross_episode_edges` → `backfill_orphan_chunks_cross_episode` → the script loop. The loop now advances `offset += effective_batch` per iteration regardless of progress. New break condition is `offset >= candidates_before` (walked the whole pool). Regression test `test_cross_episode_finder_honors_offset` inserts 3 chunks and verifies offsets 0/1/2 each expose a distinct leading row, with offset=3 empty.

7/7 F070.1 unit tests pass. Eval-scale dry-run still returns count in <1s. Re-review please.

tfatykhov:
@codex round-3 P1 addressed:

You're right — `offset` was the wrong abstraction. When a batch links some chunks, they drop out of the ordered result set, and `offset += effective_batch` jumps past their (still-unlinked) neighbors. The 3-candidate scenario you cited (batch=2, 1 success, 1 hard negative) would exit early leaving C unattempted.

Replaced `offset` with **exclusion tracking**:
- `find_chunks_lacking_cross_episode_edges(..., exclude_ids: set[UUID] | None)` — adds `c.id::text != ALL(CAST(:excl AS text[]))` to the SQL when non-empty
- `backfill_orphan_chunks_cross_episode(..., exclude_ids) -> tuple[int, list[UUID]]` — returns (edges_created, attempted_ids); caller extends `attempted` across batches
- Script loop tracks `attempted: set` across batches, no longer uses offset

New regression test `test_cross_episode_backfill_visits_every_chunk_even_when_some_link` builds the exact scenario you described and asserts every chunk lands in the attempted set.

The previous offset param is removed entirely; it was the wrong shape.

All 8 F070.1 tests pass on real Postgres. Re-review please.

PR #452, opened 2026-05-26
fix(brain): auto_link referenced a constraint name that never existed

tfatykhov:
## Summary

Tim noticed prod logs spamming:
\`\`\`
nous.brain.brain WARNING auto_link failed for decision <uuid>, continuing
\`\`\`

## Root cause

\`Brain._auto_link\` at \`brain.py:1567\` used:
\`\`\`python
.on_conflict_do_nothing(constraint=\"uq_edges_src_tgt_rel\")
\`\`\`

But the actual unique constraint on \`brain.graph_edges\` is auto-named by Postgres as \`graph_edges_source_id_target_id_relation_key\` (init.sql:205 declares \`UNIQUE(source_id, target_id, relation)\` with no explicit name). The referenced name \`uq_edges_src_tgt_rel\` **never existed in any migration**.

Every \`_auto_link\` call therefore raised \`psycopg2.errors.UndefinedObject\` and was silently caught by the \`except Exception\` at \`brain.py:420\` — so auto-linking has been **a 100% no-op since this code was written**. \`brain.auto_link()\` returns the in-memory edges list as if writes succeeded; the WARN was the only operator signal. Existing \`test_record_auto_links\` accepted \`isinstance(edges, list)\` as success → didn't catch it.

## Fix

1. **Replace \`constraint=\"...\"\` with \`index_elements=[\"source_id\", \"target_id\", \"relation\"]\`** — matches by columns. Mirrors \`GraphLinker.create_edge\` which uses this pattern and works.
2. **Add \`exc_info=True\` to the WARN log** so future failures show the actual exception in the trace.

## Test

Regression: \`test_auto_link_uses_existing_constraint_not_bogus_name\` constructs two cosine=1.0 decisions and verifies an edge IS inserted. Pre-fix raises UndefinedObject; post-fix passes.

7/7 auto-link / neighbor tests pass on real Postgres.

## Scope

Unrelated to F070/F070.1. Pure pre-existing bug in main. Recommend merging ahead of those PRs since it's small + clear.

## Test plan

- [ ] @codex review
- [ ] Merge + restart prod nous
- [ ] Confirm \`auto_link failed for decision\` WARN stops appearing
- [ ] Check graph_edges count grows: similar decisions should now produce \`auto_linked=true related_to\` edges (it hasn't been working all this time)

### DIFF
--- a/nous/brain/brain.py
+++ b/nous/brain/brain.py
@@ -418,7 +418,15 @@ class Brain:
             async with session.begin_nested():
                 await self._auto_link(decision.id, session)
         except Exception:
-            logger.warning("auto_link failed for decision %s, continuing", decision.id)
+            # exc_info=True so the actual failure mode is visible in prod
+            # logs. The prior log line gave no clue why auto-linking was
+            # silently a no-op (it was the missing constraint name; see
+            # _auto_link). Keep the WARN-level since auto-link is best-
+            # effort and we don't want to fail the parent record() call.
+            logger.warning(
+                "auto_link failed for decision %s, continuing",
+                decision.id, exc_info=True,
+            )
 
         # Re-fetch with eager loading to avoid lazy-load issues
         decision = await self._get_decision_orm(decision.id, session)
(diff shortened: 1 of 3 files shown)


tfatykhov:
@codex please review

tfatykhov:
@codex round-2 P1 addressed:

**Brain was constructed with the wrong agent_id** — fixed. Settings.agent_id defaults to whatever's in the env (`NOUS_AGENT_ID` or `nous-default`), and Brain reads its agent_id from settings at construction time. `auto_link` then queries/inserts under `brain.agent_id`. So passing `--agent-id alice` while the env was `nous-default` would have:
- Loaded decision IDs from `alice`
- Queried similar decisions in `nous-default` (the wrong tenant)
- Written graph_edges rows under `nous-default` (cross-tenant pollution)

Now `settings.model_copy(update={"agent_id": agent_id})` before constructing Brain, with an assertion to fail loudly if the scoping doesn't take. Verified by running the script for `nous-prod-2026-05-24` on the eval DB: all 7 new `auto_linked=true` decision-decision edges land under that agent_id, zero under `nous-default`.

Also cleaned up two non-ASCII characters in print output that broke on Windows cp1252 (≈, ↔) — script now runs cleanly without PYTHONIOENCODING override.

Re-review please.

PR #453, opened 2026-05-26
feat(F050): CJK-aware word counting + first unit tests for query expansion

tfatykhov:
## Summary

Direct answer to your question \"does our spec is similar to gbrain for that?\" — yes, F050 was implemented earlier inspired by gbrain's \`src/core/search/expansion.ts\`, but:
1. Shipped without unit tests
2. Used Latin-only \`len(query.split())\` gate — rejected every CJK query

This PR adds:

**\`nous/heart/cjk.py\` (new)** — shared CJK detection + word counting ported from gbrain's \`src/core/cjk.ts\`. BMP-only Unicode ranges (Han + Hiragana + Katakana + Hangul Syllables). The \`count_cjk_aware_words\` helper switches between whitespace tokens (Latin-dominant) and non-whitespace chars (CJK-dominant) at the 0.30 density threshold. Single source of truth for any future CJK-aware caller (chunker, slug grammar).

**F050 gate now uses CJK-aware count** — pre-fix, a 9-character Chinese query had 1 whitespace token and \`min_words=3\` rejected it. Post-fix, the same query passes via char counting.

**34 unit tests** — CJK detection + count, gate behavior, input/output sanitization, fuse semantics, fail-open paths. Tests adapted to the actual Nous sanitizer (the \`_INJECTION_PREFIXES\` tuple) rather than gbrain-shaped assumptions.

## Behavior change

- Latin queries: **no change** (helper falls back to whitespace tokens below 0.30 CJK density)
- Pure-CJK queries: **now pass the gate** when their char count meets \`min_words\`
- Mixed-Latin-dominant queries (e.g. \"What 北京 like\"): **no change**, density below threshold keeps whitespace tokens

## Test plan

- [ ] @codex review
- [x] 34 unit tests pass locally
- [ ] No production wire-in changes — module + gate fix only
- [ ] Existing F050 callers unaffected when \`query_expansion_enabled=false\` (default)

### DIFF
--- a/nous/heart/cjk.py
+++ b/nous/heart/cjk.py
@@ -1,0 +1,74 @@
+"""Shared CJK (Chinese / Japanese / Korean) detection + word counting.
+
+Ported from gbrain's ``src/core/cjk.ts`` (github.com/garrytan/gbrain) so
+downstream callers (F050 query expansion gate, future chunker support)
+share one source of truth instead of duplicating regex literals.
+
+Scope: BMP-only Unicode ranges that cover ~99% of real CJK content:
+  - Han (CJK Unified Ideographs): U+4E00-U+9FFF
+  - Hiragana: U+3040-U+309F
+  - Katakana: U+30A0-U+30FF
+  - Hangul Syllables: U+AC00-U+D7AF
+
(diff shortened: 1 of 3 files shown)


tfatykhov:
@codex please review

PR #454, opened 2026-05-26
fix(eval): wire F050 QueryExpander so harness reflects prod retrieval

tfatykhov:
## Summary

Tim spotted a methodology gap: \`.env\` has \`NOUS_QUERY_EXPANSION_ENABLED=true\` and \`main.py\` wires a QueryExpander into Heart at startup, but \`scripts/eval/eval_session_level_recall.py\` constructs Heart directly and **never calls \`heart.set_query_expander\`**.

Result: every LME eval we ran tonight measured a stripped-down pipeline that doesn't reflect what prod actually runs. The F070.1 baseline of 0.917/0.950 was F050-OFF; prod is F050-ON.

## Changes

1. **Eval script wires F050 the way main.py does** — constructs \`AnthropicClient\` if the flag is on, builds \`QueryExpander\`, calls \`heart.set_query_expander\` on both heart variants, cleans up api_client in finally.

2. **Skipped F055 / F023 / F027 / F047**: F055 needs ConversationState across turns (stateless eval can't exercise it); F023/F027/F047 are fact-write-path wirings, irrelevant for retrieval-only eval.

3. **Sample-rate INFO log** (\`_maybe_log_success\`) in \`nous/heart/query_expansion.py\` — every 100th successful expansion emits one INFO line with source / variant count / elapsed ms. Operators tailing at INFO can now see F050 is alive without DEBUG noise. Query text intentionally omitted for privacy (matches gbrain).

## Deferred (in memory)

- Periodic F050 stats summary log
- \`/status\` endpoint exposing F050 counters (cache hit rate, budget consumed, fail-open events)

Saved to memory as \`project_f050_observability_followups\` so they don't get lost.

## Next

After this lands, re-run the LME eval (currently in progress in background) is the actual prod-parity baseline. If F050-on lifts above 0.917/0.950, our F070.1 conclusion needs re-examination.

## Test plan

- [ ] @codex review
- [ ] Verify startup log shows \`F050: QueryExpander wired\`
- [ ] Run eval; confirm no Haiku errors at INFO level
- [ ] Compare result vs prior 0.917/0.950 baseline

### DIFF
--- a/nous/heart/query_expansion.py
+++ b/nous/heart/query_expansion.py
@@ -46,6 +46,12 @@ logger = logging.getLogger(__name__)
 
 _MAX_QUERY_LEN = 500
 _MAX_VARIANT_LEN = 200
+# Observability: emit an INFO line every Nth successful expansion so a
+# normal log tail confirms F050 is alive in prod (the per-call DEBUG
+# would be a firehose). 100 picked as a balance between visibility and
+# noise — at a few recall_deep calls per minute, this lands ~one log
+# every 30-60 min during normal use.
+_SUCCESS_LOG_SAMPLE = 100
 
 _INJECTION_PREFIXES: tuple[str, ...] = (
     "ignore ",
@@ -138,6 +144,11 @@ class QueryExpander:
         self._budget_lock = asyncio.Lock()
         self._bucket_count: dict[int, int] = {}
         self._budget_warned_bucket: int | None = None
+        # Observability sample counter — one INFO log per N successful
+        # expansions so operators tailing logs see F050 is alive without
+        # the per-call DEBUG firehose. Query text is NOT logged (privacy
+        # — matches gbrain's expansion.ts pattern).
+        self._success_count: int = 0
 
     # ------------------------------------------------------------------
     # Public entrypoint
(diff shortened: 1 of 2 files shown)


tfatykhov:
@codex please review

tfatykhov:
@codex round-1 P2 findings addressed:

**P2 #1 (eval script chunks_off ranking)** — fixed. Both `chunks_off` and `baseline` now call `run_recall_pipeline(..., rerank_by_score=True)`. Pre-fix the comparison conflated chunk-state AND ranking-strategy; post-fix the variant delta is a clean chunks-on-vs-off A/B. Caveat: this changes prior chunks_off numbers (0.917/0.733/0.933 baselines were stage-ordered). All future runs are apples-to-apples within a run.

**P2 #2 (F050 success log on Haiku failure)** — fixed. `_maybe_log_success(source="haiku")` now gated on `len(cleaned) > 0` (i.e., Haiku actually returned variants that survived output sanitization). When Haiku returns `[]` (auth/timeout/network), `final` is `[query]` and we no longer log "expansion success" — which would have lied to operators after the one-time 401 WARN was suppressed.

Cache-hit success log unchanged (cache hits are by definition real prior successes).

Re-review please.

PR #457, opened 2026-05-26
feat(F069): document-aware ingestion path Phase 1

tfatykhov:
## Summary

Closes the doc-ingestion gap (F069 Phase 1). Two parts:

**(a)** `web_fetch` limit raised: default 10K → 50K, hard ceiling 50K → 200K so explicit doc-ingest callers can pull a full arxiv paper in one fetch.

**(b)** New `ingest_document(content, source_ref)` tool that chunks document bodies at 1500-char structure-aware boundaries and persists them to `heart.episode_chunks` with `source_kind='document'`. Bypasses soft-trim by writing directly to DB at the handler level. Nous parses PDF/.docx himself via `run_python` (`pypdf` / `python-docx` client-side) and calls `ingest_document` to persist.

## The bug, measured

| Path | Pre-fix retention | Why |
|---|---|---|
| arxiv via `web_fetch` | ~1% (1.5KB of a 300KB paper) | 10K fetch cap → 3.5K soft-trim → 6× 600-char dialogue chunks |
| Word doc via `read_file` | 0% (binary garbage) | `read_text(utf-8, errors='replace')` on a ZIP archive |
| PDF via `read_file` | 0% (binary garbage) | same |
| Long markdown via `read_file` | ~2% | 1MB read OK, but soft-trim + dialogue chunker |

## Architecture

Tool args flow through agent context **once** at emission and get persisted by the handler at full fidelity **before** soft-trim degrades the conversation copy. So even though Nous's context briefly holds the huge text, the chunks land in `heart.episode_chunks` at full fidelity and survive forever.

No auto-classification heuristic in v1 — the tool is explicit and opt-in. The agent decides what's a document.

### Backend

- **`sql/migrations/052_f069_document_source_kind.sql`**: adds `source_kind VARCHAR(32) NOT NULL DEFAULT 'dialogue'` (check constraint: dialogue|document|code) + `source_ref TEXT NULL` + index on `(agent_id, source_kind)`. Existing F067 rows default to `'dialogue'`.
- **`nous/heart/document_chunker.py`**: recursive structure-aware splitter at 1500-char target / 200-char overlap. Walks `\n\n` → `\n` → `. ` → ` `; overlap snaps to whitespace.
- **`nous/api/tools.py`**: new `ingest_document` tool + JSON schema + dispatcher registration + `_session_id` injection.
- **`nous/api/web_tools.py`**: hard ceiling 50K → 200K.
- **`nous/config.py`**: `web_fetch_max_chars` 10K → 50K; new `document_ingest_enabled` / `document_chunk_size` / `document_chunk_overlap` / `document_chunk_min_chars`.
- **`nous/storage/models.py`**: `EpisodeChunk` gets `source_kind` + `source_ref`.

### Tests

- **`tests/test_document_chunker.py`** (new): 15 unit tests covering empty/short passthrough, parameter validation, paragraph-boundary preference, oversized-atom passthrough, overlap word-boundary snapping, deterministic output, realistic arxiv-shape document. **15/15 pass.**
- **`tests/test_ingest_document_integration.py`** (new): 3 end-to-end tests against real Postgres covering happy path, missing-episode error, short-content rejection. **3/3 pass.**
- **`tests/test_web_tools.py`**: updated factory default to 50K; renamed `test_max_chars_capped_at_50000` → `test_max_chars_capped_at_hard_ceiling` and pointed at the new 200K ceiling.

### Docs

- **`CLAUDE.md`**: `ingest_document` row added to Agent Tools table.

## Verification

- **85/85** dashboard + density + chunker + web_tools tests pass against local eval Postgres (`NOUS_TEST_DB=postgres`, :5433).
- **3/3** integration tests pass: real DB roundtrip writes `source_kind='document'` + `source_ref` rows under the target episode.
- Migration 052 applied cleanly to the eval container; existing F067 rows default to `'dialogue'` as expected.

## Apples-to-apples

All changes additive:
- New column has DEFAULT so existing F067 rows unchanged.
- New tool isolated from existing `recall_deep` / `web_fetch` flow.
- Soft-trim continues working identically for dialogue messages — only the explicitly-ingested content goes to chunks at full fidelity.
- F067 dialogue chunker (`chunk_text`) and its tests untouched.

## Deferred (separate decisions, not lost)

- **F069 Phase 2**: tool-dispatch auto-classification of web_fetch results (`classify_tool_result`) + soft-trim bypass heuristic.
- **F069 Phase 3**: per-source-kind retrieval weighting + display differentiation in `_format_pipeline_text`.

Both need a doc-QA eval to validate (LME is dialogue-shaped); spec deliberately defers them.

## Forge decision

Decision `4390e625` finalized in forge MCP after self-architecture review (pattern reference: F042 gated-optional-retrieval-feature, decision `8629d0e3`).

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -587,6 +587,7 @@ DB connection vars are **unprefixed** (shared with docker-compose). All others u
 | `recall_recent` | all | Retrieve recent memory items |
 | `learn_fact` | conversation, question, creative, task | Store a new fact |
 | `learn_skill` | conversation, question, task | Register a skill from URL, local path, or inline markdown |
+| `ingest_document` | conversation, question, task, debug | F069: chunk & persist a full document body (arxiv, PDF/.docx text, long markdown) to heart.episode_chunks with source_kind='document'. Use after extracting text yourself via run_python / web_fetch. |
 | `get_procedure` | all | Retrieve a specific procedure by ID |
 | `create_censor` | all | Create a guardrail censor |
 | `cache_retrieve` | all | Retrieve original content from SmartCompressed results |
--- a/nous/api/runner.py
+++ b/nous/api/runner.py
@@ -50,12 +50,12 @@ _OPTIONAL_DECISION_FRAMES = frozenset({"task", "debug"})
 
 # Frame-gated tool access (D5)
 FRAME_TOOLS: dict[str, list[str]] = {
-    "conversation": ["record_decision", "learn_fact", "learn_skill", "recall_deep", "recall_recent", "recall_hubs", "get_procedure", "create_censor", "bash", "read_file", "write_file", "web_search", "web_fetch", "cache_retrieve", "spawn_task", "spawn_sync", "schedule_task", "list_tasks", "cancel_task", "run_python", "send_file", "heartbeat_check_create", "heartbeat_check_manage", "dag_create", "dag_manage"],
-    "question": ["recall_deep", "recall_recent", "recall_hubs", "get_procedure", "bash", "read_file", "write_file", "record_decision", "learn_fact", "learn_skill", "create_censor", "web_search", "web_fetch", "cache_retrieve", "list_tasks", "cancel_task", "run_python", "dag_manage"],
+    "conversation": ["record_decision", "learn_fact", "learn_skill", "recall_deep", "recall_recent", "recall_hubs", "get_procedure", "create_censor", "bash", "read_file", "write_file", "web_search", "web_fetch", "cache_retrieve", "spawn_task", "spawn_sync", "schedule_task", "list_tasks", "cancel_task", "run_python", "send_file", "heartbeat_check_create", "heartbeat_check_manage", "dag_create", "dag_manage", "ingest_document"],
+    "question": ["recall_deep", "recall_recent", "recall_hubs", "get_procedure", "bash", "read_file", "write_file", "record_decision", "learn_fact", "learn_skill", "create_censor", "web_search", "web_fetch", "cache_retrieve", "list_tasks", "cancel_task", "run_python", "dag_manage", "ingest_document"],
     "decision": ["record_decision", "recall_deep", "recall_recent", "recall_hubs", "get_procedure", "create_censor", "bash", "read_file", "web_search", "web_fetch", "cache_retrieve", "list_tasks", "cancel_task", "dag_manage"],
     "creative": ["learn_fact", "recall_deep", "recall_recent", "recall_hubs", "get_procedure", "write_file", "web_search", "cache_retrieve"],
     "task": ["*"],  # All tools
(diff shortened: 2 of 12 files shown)


tfatykhov:
@codex addressed all three findings in commit 0673812:

- **P1-1** (FRAME_TOOLS): added `ingest_document` to conversation/question/debug frame allowlists in `nous/api/runner.py:52-58`. `task` already had `*`.
- **P1-2** (web_fetch schema): bumped `max_chars.maximum` from 50000 → 200000 in `_WEB_FETCH_SCHEMA` to match the new runtime ceiling.
- **P2** (chunker size contract): `_pack` now re-checks size after overlap-tail seeding; if the seeded buffer would exceed target_size, drops the tail and starts the new buf with just the piece. Regression test `test_overlap_tail_does_not_violate_size_contract` covers the 92+30 vs 100 case.

Also fixed the CI test failure root cause: `test_create_nous_tools` needed `ingest_document` in its expected key set.

tfatykhov:
@codex round 2 addressed in commit 978cfbe:

- **P2** (delimiter preservation in chunker): `_recursive_split` now right-attaches the delimiter to each piece except the last, so periods, paragraph breaks, and sentence punctuation survive chunking. `_pack` joins with no extra separator since the delimiter is already in the piece content. New `TestDelimiterPreservation` class covers periods, `. ` boundaries, and paragraph breaks (`\n\n`).

- **P1** (idempotency): pre-check rejects re-ingest if any chunks already exist under (agent_id, episode_id, source_ref). Returns 'Already ingested: N chunks exist for source_ref=X' with guidance to use a different source_ref or delete first. Integration test `test_ingest_document_idempotent_for_same_source_ref` covers all three cases: same source_ref → reject + row count stable; different source_ref → succeeds; baseline first call → succeeds.

73/73 pass locally against eval Postgres.

tfatykhov:
@codex round 3 addressed in commit fc7a8c9:

- **P1** (atomic idempotency): collapsed COUNT + MAX + INSERTs into a single transaction with `pg_advisory_xact_lock(hashtextextended('ingest_document:{episode}:{source_ref}', 0))`. The lock auto-releases on COMMIT/ROLLBACK, so concurrent calls for the same key serialize at the DB; the second waiter sees the first's rows and returns 'Already ingested'.

- **P2** (embed-batch length guard): mirrored the F067 writer pattern from `handlers/episode_summarizer.py:205-211`. On `len(embeddings) != len(chunks)`, log WARNING and return an error response without writing — better to refuse than write a partial ingest. New test `test_ingest_document_rejects_partial_embed_batch` verifies 0 rows written on mismatch. Pre-existing mocks updated to dynamic side_effect (one vector per input chunk).

74/74 pass on eval Postgres.

tfatykhov:
@codex round 4 addressed in commit ba24b96:

- **P1** (lock scope): dropped `source_ref` from the advisory lock key. Lock is now keyed on `episode_id` alone, so concurrent ingests targeting the same episode (even with different source_refs) serialize their `MAX(chunk_index)+1` allocations. Idempotency COUNT still filters by source_ref inside the serialized transaction.

- The other comments in this round are repeats of earlier findings already resolved in round 3 (atomic transaction + embed-batch length guard).

New `test_ingest_document_concurrent_different_source_refs_no_collision` uses `asyncio.gather` to launch two concurrent ingests against the same episode with different source_refs and verifies disjoint chunk_index ranges + zero silently-dropped rows. 6/6 integration tests pass.

PR #456, opened 2026-05-26
feat(dashboard): surface chunks + F070 edges + F065 provenance

tfatykhov:
## Summary

Closes the dashboard side of the F067 / F070 / F070.1 / F065 graph series. Two commits:

**1. `feat(dashboard): surface chunks + F070 edges + F065 provenance`** — Knowledge Graph + Graph Health + Density now see chunks as a first-class node type, plus F065 `extraction_method` per edge.

**2. `feat(dashboard): chunks tab in Memory Browser + Total Chunks stat card`** — Memory Browser gets a Chunks tab; Overview gets a Total Chunks stat card; new `GET /chunks` REST route.

## The bug, in prod data

Probed prod (`http://192.168.1.141:8383`) before this fix:

| Endpoint | Pre-fix | What was wrong |
|---|---|---|
| `/dashboard/graph?limit=500` | 2000 edges returned, 1986 with `source_type='chunk'`, **zero chunk nodes** in the nodes array | `type_queries` only resolved 4 source tables — chunk source IDs were dropped, force graph rendered phantom edges |
| `/dashboard/density` | `density_by_type` has 4 types, but `edge_distribution` shows `part_of: 1276` + `summarized_by: 690` | Per-type breakdown silently missed the largest node population |
| `/dashboard/health` | `orphan_trend` for 2026-05-26 reads `0` | Quiet correctness bug — `daily_new_connected` counted chunks, `daily_new_total` didn't, diff went negative, `GREATEST(...,0)` clamped to zero |
| Memory Browser | 5 hardcoded tabs (facts/episodes/decisions/procedures/censors) | No chunk surface anywhere |
| Overview stat cards | facts/episodes/decisions/procedures/censors counts only | No total_chunks card |

### Commit 1: graph / health / density (4 files, +47 / −8)

**Backend — `nous/api/dashboard_queries.py`**
1. `get_graph_data.type_queries`: + `chunk` → `heart.episode_chunks` (`chunk_index::text` as category)
2. `get_graph_data` edge SELECT: + `extraction_method`, propagated to each edge dict
3. `get_graph_data.orphan_queries`: + chunks
4. `get_health_data.orphan_queries`: + chunks
5. `get_health_data` orphan-trend UNION: + `heart.episode_chunks` *(quiet correctness fix)*
6. `get_density_data.type_configs`: + chunk

**Frontend**
7. `static/dashboard/js/app.js` `typeColor`: chunk = `#06b6d4`
[...]

PR #458, opened 2026-05-26
feat(dashboard): swap D3 for Cytoscape + centrality + cluster coloring

tfatykhov:
## Summary

Replaces the D3 force-directed Knowledge Graph view with Cytoscape.js 3.30.4. Phase 1 (library swap with parity) + Phase 3 (centrality sizing + community detection). Bundles the long-standing cache-busting fix for dashboard static assets.

## Why now

Three pain points converged:

1. **Console errors**: D3's drag handlers race against \`rebuildGraph()\` filter changes, throwing \`Cannot read properties of null (reading 'alphaTarget')\` at \`dragStart\`/\`dragEnd\`. The simulation reference goes stale between filter toggles. Cytoscape has a single instance lifecycle and eliminates this entire class of bug.

2. **Performance ceiling**: D3 forceSimulation degrades past ~500 nodes. We're at 815 today and growing ~1K/week as F070 chunk backfill runs land more chunks. Cytoscape handles 5K-10K comfortably.

3. **Hairball topology**: hub-and-spoke graphs (chunks around episodes, decisions around their evidence_for facts) produce a hairball in pure force-directed view. No clustering or centrality cues meant most nodes looked equally important.

### Phase 1 — library swap with parity

| | Before (D3) | After (Cytoscape) |
|---|---|---|
| Library | d3.min.js | cytoscape.min.js 3.30.4 vendored |
| Layout | \`d3.forceSimulation\` | \`cose\` (built-in, tuned for hub-and-spoke) |
| Type filters | ✓ | ✓ (incl. chunk) |
| Min-edges slider | ✓ | ✓ |
| Search highlight | ✓ | ✓ (neighborhood fade) |
| Detail panel | ✓ | ✓ (same fetch logic) |
| F065 inferred-edge dash | ✓ (post PR #456) | ✓ |
| F070 relation colors | ✓ (post PR #456) | ✓ |

### Phase 3 — centrality + community

- **Degree-based sizing**: \`mapData(edge_count, 0, 30, 14, 44)\`. Hubs visible immediately; isolated chunks shrink. No more "everything looks the same size" problem.
- **"Highlight clusters" toggle**: runs \`cy.elements().components()\` to find connected components; each cluster gets a colored 4px border (type color stays as fill). 12-hue palette cycled across components. Reveals the dec↔dec cluster, fact/episode bridges, and chunk constellations as visually distinct islands.

### Cache-busting fix (bundled because every dashboard PR was hitting this)

\`_NoCacheStaticFiles\` wraps Starlette's \`StaticFiles\` to set \`Cache-Control: no-cache, must-revalidate\` on every dashboard asset response. ETag is still set by the base class, so re-validation returns 304 (zero body) when nothing changed. Eliminates the post-deploy stale-bundle problem.

## What's NOT in this PR

Per user direction, **Phase 2 deferred**:
- Compound nodes (collapse chunks under their parent episode)
- Layout switcher dropdown (force vs hierarchical vs concentric)

If connected components proves insufficient as a clustering signal, a follow-up could add Louvain modularity via the cytoscape-louvain extension. Started conservative.

## Verification

- API response shape verified against prod \`/dashboard/graph?limit=10\` — node {id,type,label,category} and edge {id,source,target,source_type,target_type,relation,weight,extraction_method} both match what graph.js expects.
- Python syntax check on rest.py passes.
- Cytoscape vendored file (374,376 bytes) starts with the official Cytoscape Consortium copyright header.
- 6 tasks tracked + completed.

## Files

- \`static/dashboard/lib/cytoscape.min.js\` (new, 374 KB vendored)
- \`static/dashboard/js/graph.js\` — full rewrite (~590 LOC)
- \`static/dashboard/index.html\` — \`<script src="lib/d3.min.js">\` → \`<script src="lib/cytoscape.min.js">\`
- \`nous/api/rest.py\` — \`_NoCacheStaticFiles\` wrapper at the Mount


tfatykhov:
@codex both findings addressed in commit 35066bc:

- **P1** (DAG view regression): restored `<script src="lib/d3.min.js">` alongside cytoscape. Confirmed dag.js uses `d3.select()` at lines 413 + 474. Cytoscape replaces D3 only for the Knowledge Graph view; DAG remains on D3. Comment in index.html notes why both libraries load.

- **P2** (edge fade order-dependence): replaced the per-node toggle loop with a two-pass approach. Pass 1 classifies nodes and records matches in a Set; pass 2 visits each edge exactly once and fades only if NEITHER endpoint matches. Order-independent.

tfatykhov:
@codex addressed in commit 798a687:

- **P2** (Cytoscape instance leak): two-place teardown. (1) `app.js` loadView leave-graph-view branch now calls `Dashboard._cyInstance.destroy()` alongside the legacy `_graphSimulation.stop()`. (2) `graph.js` belt-and-suspenders: destroys any pre-existing `Dashboard._cyInstance` at view-entry to cover the `reloadView('graph')` path where the leave-view branch doesn't fire (currentView === target). New cy registered on `Dashboard._cyInstance` after construction so the leave-view teardown can find it.

The other two comments on this round are repeats from earlier commits (P1 d3-for-DAG and P2 edge-fade), both resolved in 35066bc.

tfatykhov:
@codex addressed in commit d858fbf:

- **P2** (clusters from visible subgraph): three-place fix. (1) `computeClusters()` clears stale `clusterColor` then derives components from `cy.elements().filter(:visible)`. (2) `applyClusterColors()` re-runs `computeClusters()` when cluster mode turns on, so toggling on always reflects current filters. (3) `applyFilters()` invokes `applyClusterColors()` while clusterMode is on, so type/min-edges filter changes re-color live without re-toggling. Removed the unused `clusterIdByNodeId` map.

PR #463, opened 2026-05-29
fix(#462): background turns no longer starve the sleep cycle

### LINKED ISSUE #462: Sleep cycle starved indefinitely by background heartbeat turns (autopilot_tick resets idle timer)
## Summary

The sleep cycle is being indefinitely starved by internal heartbeat activity. Despite `NOUS_SLEEP_TIMEOUT=7200s` (2 h), the last `sleep_started` event in the deployed instance fired **22.7 hours ago** (2026-05-28 02:06 UTC). Over that window, the maximum activity gap was only **~32 min** — every gap is closed by a `dynamic-check-autopilot_tick-*` session.

## Root cause

`SessionMonitor.on_activity` is subscribed to the `turn_completed` event and unconditionally resets the global sleep timer:

```python
# nous/handlers/session_monitor.py
self._global_last_activity = now
self._sleep_emitted = False
```

`CognitiveLayer.run_turn` emits `turn_completed` for **every** turn — including background heartbeat ticks — and does not propagate the existing `is_background` flag into `event_data`:

```python
# nous/cognitive/layer.py  (~L1135)
await self.event_bus.emit(Event(
    type="turn_completed",
    data={...},   # is_background NOT included
))
```

Meanwhile `DynamicCheck.run` calls `runner.run_turn(..., is_background=True)` every `interval_seconds`. The built-in `autopilot_tick` check runs every **1800 s**, which is strictly less than `NOUS_SLEEP_TIMEOUT`, so the idle timer can never expire.

Net effect: a 30-min internal heartbeat permanently resets a 2 h sleep timer → **sleep never fires** in any normally-configured deployment with autopilot enabled.

## Evidence

- Last `sleep_started`: `2026-05-28 02:06:38 UTC` (22.7 h before report)
- Activity-gap analysis over the last 22 h: max gap **1938 s** (~32 min); all gap-ending events are `session_id LIKE 'dynamic-check-autopilot_tick-%'`
- `is_background` is plumbed through `AgentRunner.run_turn` and used by F048 streaming routing, but never reaches `SessionMonitor`

## Proposed fix (minimal, surgical)

1. **`nous/cognitive/layer.py`** — include `is_background` in the `turn_completed` event payload:
   ```python
   data={
       ...,
       "is_background": is_background,
   }
   ```
2. **`nous/handlers/session_monitor.py`** — in `on_activity`, early-return when `event.data.get("is_background") is True`. Do NOT touch `_global_last_activity` or `_sleep_emitted`.
3. Audit any direct callers of `SessionMonitor.touch()` for the same leak (e.g. heartbeat schedulers, autopilot loops) and gate them on the same flag.

## Hardening (follow-up, optional)

- Regression test: 5 consecutive `is_background=True` turns spaced over `> sleep_timeout` must still produce a `sleep_started` event.
- Expose `seconds_since_last_sleep` on `/status` so this class of regression is loudly visible.
- Consider a defensive cap in `SessionMonitor`: if `now - last_sleep > 2 * sleep_timeout`, force a sleep regardless of activity (catches future leaks).

## Impact

- Sleep consolidation (densification, supersession, decay) hasn't run for >22 h in the live instance.
- Any deployment with `autopilot_tick` or any sub-`sleep_timeout` dynamic check is silently affected.
- No user-visible error — the bug is invisible without explicit `/sleep/status` checks.

## Workaround until fix lands

`curl -X POST http://localhost:8000/sleep/trigger` to force a cycle manually.

---
_Diagnosed in conversation with Nous, 2026-05-29._

tfatykhov:
## Problem

Heartbeat and subtask turns emit `turn_completed` exactly like foreground turns. `SessionTimeoutMonitor.on_activity` unconditionally reset `_global_last_activity` and cleared `_sleep_emitted` on every such event. So any background turn firing inside the `sleep_timeout` window deferred the sleep cycle **indefinitely** — under a normal heartbeat tick interval (30s) vs sleep_timeout (default 7200s), sleep never fired at all once heartbeat was active.

Closes #462.

Knock-on effect this also unblocks: the F075 `happened_before` edge build runs during the sleep cycle. No sleep → no temporal edge densification in prod.

## Fix

Thread `is_background` through the existing F048 wire (the same flag heartbeat/subtask callers already pass into `run_turn`):

- **`CognitiveLayer.post_turn`** gains `is_background: bool = False`; included in the `turn_completed` `event_data`.
- **`runner.run_turn`** passes `is_background` to both `post_turn` calls and to `monitor.touch(is_background=...)`.
- **`SessionTimeoutMonitor`** keeps a background session **tracked** in `_last_activity` (recorded in a new `_background_sessions` set), but for background turns it does **not** reset `_global_last_activity` / `_sleep_emitted`, and the `all_sessions_sleeping` gate **excludes** `_background_sessions`.

`stream_chat` (foreground) has no `is_background` param; its `post_turn` calls correctly default to `False`.

### Why tracked-but-excluded (not invisible)

The first cut made background sessions fully invisible to the monitor. Codex correctly flagged that inline `await_result` subtasks (`tools.py`) run `run_turn(is_background=True)` with their own `session_id` and never call `end_conversation` — the monitor's idle-close was their **only** cleanup, so invisibility leaks runner `_conversations`/ledger/persisted state after each inline subtask. Keeping them tracked preserves that cleanup. The exclusion-set is independently necessary: a recurring background session (heartbeat reusing one `session_id`) keeps a "recent" `_last_activity` entry forever, which would block the gate — the same starvation in a second guise. `_background_sessions` is discarded on session end and on idle-close, so it can't grow unbounded.

## Verification

- **Producer stamps it** — every background `run_turn` caller already passes `is_background=True`: heartbeat triage + on_complete, subtask_worker, DynamicCheck, subtask_executor, inline spawn_task. Foreground emitters (rest/mcp) omit it.
- **Sole emitter** — `turn_completed` is emitted only inside `post_turn` (`layer.py`).
- **Re-arm semantics** — `_sleep_emitted` re-arms on foreground activity, matching the global-timer gating. Sleep fires once per genuine-idle period.

## Tests

`TestSessionTimeoutMonitor`:
- `test_background_turn_does_not_reset_global_timer` — background turn leaves `_global_last_activity` + `_sleep_emitted` untouched, but the session **is** tracked and flagged background.
- `test_sleep_fires_despite_repeated_background_turns` — 5 background turns leave a fresh `_last_activity` entry, yet sleep still fires (gate exclusion).
- `test_foreground_session_still_blocks_sleep` — guard: a recent foreground session still blocks sleep (exclusion must not leak).

Existing monitor + runner + `test_runner_background` suites green (the `touch()` call-signature assertion was updated). Pre-existing `test_cognitive_layer` DB-dependent failures are unrelated — identical on the clean baseline.

### DIFF
--- a/nous/api/runner.py
+++ b/nous/api/runner.py
@@ -354,7 +354,12 @@ class AgentRunner:
         # otherwise be closed mid-stream by a monitor tick. The bus is
         # async-queued so emitting message_received here would leave a
         # residual race; the synchronous touch eliminates it.
-        if self._session_monitor is not None:
+        #
+        # #462: skip the refresh for background turns (heartbeat/subtask).
+        # touch() resets _global_last_activity + _sleep_emitted; a background
+        # turn firing inside the sleep_timeout window would otherwise defer
+        # the sleep cycle forever. Foreground turns still close the race.
+        if self._session_monitor is not None and not is_background:
             try:
                 self._session_monitor.touch(session_id, _agent_id)
             except Exception:
@@ -394,7 +399,10 @@ class AgentRunner:
                 response_text = turn_context.censor_block_reason or "I can't process that request."
                 conversation.messages.append(Message(role="assistant", content=response_text))
                 turn_result = TurnResult(response_text=response_text)
-                await self._cognitive.post_turn(_agent_id, session_id, turn_result, turn_context)
+                await self._cognitive.post_turn(
+                    _agent_id, session_id, turn_result, turn_context,
+                    is_background=is_background,
+                )
                 return response_text, turn_context, usage
 
             # F026: Get/create execution ledger and set turn
@@ -525,7 +533,10 @@ class AgentRunner:
                 error=error,
                 thinking_blocks=thinking_blocks,
             )
-            await self._cognitive.post_turn(_agent_id, session_id, turn_result, turn_context)
+            await self._cognitive.post_turn(
+                _agent_id, session_id, turn_result, turn_context,
+                is_background=is_background,
+            )
 
             # 8. Safety net: warn if decision frame but record_decision not called
             self._check_safety_net(turn_context, tool_results)
--- a/nous/api/runner.py
+++ b/nous/api/runner.py
@@ -355,13 +355,15 @@ class AgentRunner:
         # async-queued so emitting message_received here would leave a
         # residual race; the synchronous touch eliminates it.
         #
-        # #462: skip the refresh for background turns (heartbeat/subtask).
-        # touch() resets _global_last_activity + _sleep_emitted; a background
-        # turn firing inside the sleep_timeout window would otherwise defer
-        # the sleep cycle forever. Foreground turns still close the race.
-        if self._session_monitor is not None and not is_background:
+        # #462: a background turn (heartbeat/subtask) still touches its own
+        # session entry — so the monitor won't close it mid-turn and will
+        # later reclaim its runner state — but touch(is_background=True) does
+        # not reset the global sleep timer and flags the session for exclusion
+        # from the sleep gate. Foreground turns still close the race and re-arm
+        # the timer.
+        if self._session_monitor is not None:
             try:
-                self._session_monitor.touch(session_id, _agent_id)
+                self._session_monitor.touch(session_id, _agent_id, is_background=is_background)
             except Exception:
                 logger.debug("session_monitor.touch failed (suppressed)", exc_info=True)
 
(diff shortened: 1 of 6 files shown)


tfatykhov:
@codex the P2 you flagged (test_runner_background.py _MockCognitive.post_turn missing is_background) is fixed in 83d5f9d — widened the mock signature. Please re-review.

tfatykhov:
@codex addressed the P2 (background subtask cleanup leak) in 4b97544 — switched from invisible to tracked-but-excluded: background sessions stay in _last_activity (idle-close still reclaims their runner state) via a new _background_sessions set, but are excluded from the sleep gate and don't reset the global timer. Added test_foreground_session_still_blocks_sleep so the exclusion can't leak. Please re-review.

PR #461, opened 2026-05-28
feat(F075): temporal fact extraction live-path

tfatykhov:
## Summary

Live-path implementation of F075 (temporal fact extraction + date-aware retrieval). Spec was merged in #460 (squash `0115568`) after 17 codex review rounds. This PR adds the code per impl plan v4 (`docs/superpowers/plans/2026-05-28-f075-temporal-fact-extraction.md`), which split the original 12-phase scope to ship Phases 1-7 + 9-10 here. **Phase 8 (backfill script) is deferred to F075.1** — too large + risky for single-PR review focus per devil's-advocate review.

## What ships

- **Layer 1 — extraction**: episode summarizer prompt + threading `Episode.started_at` for relative-date resolution (5 hops); FactExtractor threads `event_date` through both candidate-facts and LLM-fallback paths; dedup-bypass at all 3 sites (pre-learn × 2 + cosine in `Heart._learn`); subject-supersession bypass at `_supersede_by_subject`. All flag-gated on `NOUS_TEMPORAL_EXTRACTION_ENABLED` (default OFF).
- **Layer 2 — graph edges**: `GraphDensifier._build_happened_before_edges` builds chronological chains within episodes via LATERAL `LIMIT 1` (no quadratic on same-date clusters). Wired into `run_backfill_cycle`.
- **Layer 4 — retrieval surface**: `event_date` propagated through `FactSummary`, `FactDetail`, `Heart._to_recall_result`, and `_heart_results_to_pipeline` so it reaches `PipelineResult.metadata` for downstream consumers + future Layer 3.
- **Schema**: migration 053 (event_date + event_date_classified_at columns + 3 partial indexes including the composite for the LATERAL probe). ORM brought current with F070 drift (chunk source_type/target_type + part_of/summarized_by/happened_before relations).
- **Settings**: 6 new fields in `nous/config.py`, all dark-launch defaults.
- **Tests**: 8 pure-Python tests (validator, schema passthrough, `_merge_summaries` partition, dedup polarity) + 5 Postgres-gated roundtrip tests (skip-decorator with `NOUS_F075_POSTGRES_TESTS=1`).

## Review history

Three plan reviews + three impl reviews:

| Stage | Reviewers | Findings | Status |
|---|---|---|---|
| Plan v1 → v3 | arch + code-reviewer + devil's-advocate | 5 P1s + ~10 P2/P3s | All incorporated |
[...]

PR #460, opened 2026-05-27
spec(F075): temporal fact extraction + date-aware retrieval

tfatykhov:
## Summary

Spec proposes extracting date-anchored events as discrete facts during episode summarization. Adds `event_date` column to `heart.facts`, `happened_before` edges (reranking reinforcer), and a retrofit script (F047 pattern). Date-aware retrieval boost (Layer 3) deferred pending Layer 1+2+4 measurement.

## Motivation

BEAM-100K prod-v3 measures temporal_reasoning at **0.417**. Diagnostic chain across 3 falsified easy levers (staleness, K-bump, inline date markers) traced the root cause to `EpisodeSummarizer.candidate_facts` schema lacking a slot for dates — date-anchored events survive in `heart.episode_chunks` but never become discrete facts in `heart.facts`. Chunks containing the dates rank too low on date-arithmetic queries because surrounding content (code, prose) dominates the embedding.

**Empirical validation (~$0.05):** hand-injecting `"Christina obtained the OpenWeather API key on March 10, 2024."` into `heart.facts` for conv 2 caused the synthetic fact to rank #3 of 39 in retrieval, and the LLM produced *"...That's 2 days between the two events."* — moving conv 2 Q0's BEAM score from 0.000 → 1.000.

Failure-mode classification (verified empirically across 4 failed Qs):
- 2 PATTERN_MATCH: date in chunks, missing from facts, chunk not retrieved → extractor directly fixes
- 2 PARTIAL_MATCH: date in chunks, retrieved at rank 3, LLM doesn't compose prose → extractor helps via discrete dated facts
- 1 source ambiguity: unrelated to this feature

## Review history

Three parallel reviews (architecture, python-pro, devil's-advocate). Surfaced 4 P1s + 7 P2s + 2 P3s, all incorporated into v2:

| Finding | v2 fix |
|---|---|
| **Arch P1-1**: v1 augmented `FactExtractor._EXTRACT_PROMPT` — the fallback path | Retargets to `EpisodeSummarizer.candidate_facts` (primary prod path) |
| **Arch P1-2**: Summarizer LLM operates on lossy 150-word prose | Directs summarizer to extract dates from raw transcript it already holds |
| **Arch P1-3 / Devil's #3**: Layer 2 overstated as retrieval-surfacing lever | Reframed as reranking reinforcer; impl plan flips `NOUS_GRAPH_ADJACENCY_BOOST_ENABLED=true` |
[...]

PR #464, opened 2026-05-29
feat(F075.1): one-time temporal-fact backfill script

tfatykhov:
## What

Standalone one-time script (`scripts/backfill_temporal_facts.py`) to retrofit `event_date` onto `heart.facts` rows that were created **before** F075's live extraction (PR #461) landed. Without it, all pre-F075 facts stay `event_date_classified_at IS NULL` and never participate in `happened_before` edges or date-aware retrieval.

It is **not** a startup handler — it runs once per agent, manually, then you're done.

## How it works

For each fact with `event_date_classified_at IS NULL`:
1. Pull the nearest source-episode chunk by cosine to `fact.embedding` (context for the LLM — the verbatim chunk, not the lossy summary).
2. Ask a background LLM for the single calendar date the fact's event happened on; normalize through `FactInput`'s validator (identical drop rules to the live path — no parse-repair).
3. Write `event_date` (date | NULL) and **always** stamp `event_date_classified_at` so the row is never re-processed.
4. After the batch loop, trigger `GraphDensifier._build_happened_before_edges()` over the freshly-dated facts.

## Hardened patterns (carried verbatim from F075 spec §Layer 4, the 17-round codex baseline)

- **Two-connection advisory lock**: session-scoped `pg_try_advisory_lock` on a *dedicated* `engine.connect()` that never commits / never returns to the pool, so the lock survives every per-batch commit. Per-batch work runs on separate pooled sessions.
- **F047 collision-safe key**: lock key salted with `"f075-temporal:"` so it never collides with F047's actionability lock on the same agent (both would otherwise hash the bare `agent_id`).
- **Eligibility = `event_date_classified_at IS NULL`** (not `event_date IS NULL`): a stable fact correctly classified "no date" stays classified.
- **Commit-before-early-return** on budget exhaustion so spent LLM calls aren't rolled back.
- **Token budget → hard LLM-call cap** (call-count `BudgetTracker`).
- NULL-guarded chunk context (legacy rows with no source episode / no embedding fall back to `fact.content`).

## Usage

```bash
uv run python scripts/backfill_temporal_facts.py --agent-id nous-default            # run
uv run python scripts/backfill_temporal_facts.py --agent-id nous-default --dry-run  # estimate only, no LLM calls
uv run python scripts/backfill_temporal_facts.py --agent-id X --batch-size 100 --token-budget 50000
```

Validated end-to-end on the eval DB: `"March 10, 2024"` → `2024-03-10`, `"2024-04-09"` → `2024-04-09`, undated facts → `NULL` but still stamped, idempotent (0 eligible on re-run), lock acquired + released cleanly.

## Deploy ordering

Requires **migration 053** (shipped in #460) applied on the target DB first. In the F075 rollout it runs **after** code deploy, **before** flipping `NOUS_TEMPORAL_EXTRACTION_ENABLED`. Note: the sleep cycle (where the live edge build runs) is only unblocked once #463 merges.

## Tests

2 pure-unit tests pinning the two bits codex flagged as easy to silently break:
- `test_advisory_lock_key_namespaced_off_f047` — F075 key ≠ F047's bare-agent key.
- `test_budget_tracker_call_count_semantics` — `consume()` decrements by 1; `ok()` gates the next call.

Adds `pythonpath = ["."]` to the pytest config so the test can import the repo-root `scripts/` module under plain `uv run pytest` (CI doesn't set `PYTHONPATH`). Verified: full suite still collects 4311 tests cleanly with this added (the lone `test_f036_cache_dashboard` collection error is a pre-existing missing-`starlette` issue in the local env, unrelated).

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -56,6 +56,7 @@ dev = [
 include = ["nous*", "nous_eval*"]
 
 [tool.pytest.ini_options]
+pythonpath = ["."]
 asyncio_mode = "auto"
 asyncio_default_fixture_loop_scope = "session"
 asyncio_default_test_loop_scope = "session"
--- a/scripts/backfill_temporal_facts.py
+++ b/scripts/backfill_temporal_facts.py
@@ -114,12 +114,23 @@ _CLASSIFY_SYSTEM = (
 )
 
 
-async def _classify_event_date(client, model: str, row, chunk_context: str | None) -> date | None:
-    """LLM-extract the event date for one fact. Returns a validated date | None.
+async def _classify_event_date(
+    client, model: str, row, chunk_context: str | None
+) -> tuple[date | None, bool]:
+    """LLM-extract the event date for one fact.
+
+    Returns ``(event_date, should_stamp)``:
+      - valid date string        -> (date,  True)   record + terminal
+      - explicit null            -> (None,  True)   "no date", terminal
+      - malformed non-null date  -> (None,  False)  leave NULL, retry next run
+      - no structured result     -> (None,  False)  transient failure, retry
 
     Normalizes the raw LLM string through FactInput's validator so malformed
     shapes ('2024-3-10', ISO week, bad calendar dates) are dropped exactly as
-    on the live path — type-safe, no parse-repair here.
+    on the live path. A non-null raw the validator drops is malformed and is
+    NOT stamped terminal — otherwise one bad model response would permanently
+    lose a recoverable date on a legacy fact (mirrors the F075 live-path
+    "malformed stays eligible" rule). The row is simply revisited next run.
     """
     content = row["content"]
     context = (chunk_context or content)[:1500]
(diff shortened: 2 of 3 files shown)


tfatykhov:
@codex addressed both findings in 2839969:
- P1 (malformed dates): _classify_event_date now returns (event_date, should_stamp); malformed/transient results stay NULL and are retried next run. Switched the batch loop to keyset pagination on (learned_at, id) so un-stamped rows aren't re-fetched within a run.
- P2 (zero budget): dropped max(1,...) → max(0,...) so --token-budget 0 does zero LLM calls, matching the dry-run.
Please re-review.

tfatykhov:
@codex addressed the token-budget P2 in 2776139 — --token-budget now defaults to None and resolves from Settings.temporal_backfill_default_token_budget (NOUS_TEMPORAL_BACKFILL_DEFAULT_TOKEN_BUDGET) when omitted, so lowering or zeroing that env default is honored by the flagless command. Please re-review.

tfatykhov:
@codex fixed the P1 (pgvector text serialization) in 255001c. _fetch_chunk_context now computes the cosine entirely in SQL via a correlated subquery on the fact's embedding (matching GraphDensifier's c.embedding <=> f.embedding), so the vector never round-trips through Python. Verified on the eval scratch DB against a real embedded fact with a chunked source episode — returns the nearest chunk, no ValueError. Please re-review.

PR #465, opened 2026-05-29
fix(F071): streaming contextvar crash — Token reset across async-gen contexts

tfatykhov:
## Symptom

After **every Telegram (streaming) response**, the bot appends:

> ❌ `<Token var=<ContextVar name='CURRENT_TURN_EXCLUDE_IDS' …> …> was created in a different Context`

## Root cause

`stream_chat` is an async **generator**. `rest.py` pumps it via `asyncio.ensure_future(gen.__anext__())` once per SSE chunk (the keepalive race from the `/chat/stream` robustness work), so **each resume runs in a fresh copied context**. The F071 `CURRENT_TURN_EXCLUDE_IDS.set()` runs in one `__anext__` context; the `finally`'s `CURRENT_TURN_EXCLUDE_IDS.reset(token)` runs in a later one → CPython raises *"Token was created in a different Context"*.

Two aggravating facts:
1. That `reset()` was the **first statement** in the `finally`, so its raise **skipped the rest of the finally** — `post_turn`, the safety-net check, and `turn_context` bookkeeping — on every streaming turn.
2. The set/reset ran **unconditionally even with the flag off** (`_build_exclude_ids` returns `None` → `set(None)` still mints a token), so it crashed regardless of `NOUS_RECALL_EXCLUDE_CONTEXT_IDS` (prod default = off).

## Blast radius (what actually broke)

On every Telegram turn, `post_turn` was skipped → lost **per-turn monitoring/assessment, working-memory thread resolution, and the `turn_completed` event**.

Not broken: the user still gets the reply (streamed before the `finally`); the conversation transcript is intact (assistant message persisted before the `finally`), so **session-level memory still forms** (episode summary + fact extraction run on `session_ended`); sleep/idle timing is fine (start-of-turn `touch()` runs before the `finally`); and **non-streaming paths (`/chat`, subtasks, heartbeat) are unaffected** — they use a coroutine with a stable context, so `reset(token)` works there.

## Fix

Streaming path only (the coroutine `run_turn` path is correct by construction):
- **Engage the contextvar only when the feature is on** (`None` = flag off → no churn, no crash).
- **Restore by value (`set(None)`) instead of by token (`reset`)**, so the `finally` can never raise across the generator's per-chunk context boundaries.

## Tests

2 regression tests pump `stream_chat` via `ensure_future(__anext__())` per chunk (reproducing rest.py's context fragmentation), flag **off** and **on**; both assert no Token error and that `post_turn` still runs. **Verified they fail pre-fix** (`ValueError` at the `reset` line) **and pass post-fix**. Full `test_streaming` + dedup + runner suites green (103 passed).

### DIFF
--- a/nous/api/runner.py
+++ b/nous/api/runner.py
@@ -1053,9 +1053,19 @@ class AgentRunner:
             # loop — recall_deep can be dispatched inside the for-turn
             # iteration. Reset lives in the existing `finally:` below so
             # we don't add new nesting (single source of cleanup).
-            _f071_token = CURRENT_TURN_EXCLUDE_IDS.set(
-                _build_exclude_ids(self._settings, turn_context)
-            )
+            #
+            # stream_chat is an async GENERATOR. rest.py pumps it via
+            # asyncio.ensure_future(gen.__anext__()) once per SSE chunk (the
+            # keepalive race), so each resume runs in a FRESH copied context.
+            # A Token from set() in one __anext__ cannot be reset() in a later
+            # one — CPython raises "Token was created in a different Context".
+            # So: only engage the contextvar when the feature is actually on
+            # (None = flag off, the prod default → no churn, no crash), and
+            # restore by value (set) rather than by token (reset) so the
+            # finally can never raise across the generator's context boundaries.
+            _f071_exclude = _build_exclude_ids(self._settings, turn_context)
+            if _f071_exclude is not None:
+                CURRENT_TURN_EXCLUDE_IDS.set(_f071_exclude)
             try:
                 for turn in range(self._settings.max_turns):
                     # Unified block accumulator: keyed by block_index, preserves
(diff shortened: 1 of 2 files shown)


tfatykhov:
@codex acknowledged — you're right that the set() is invisible after the first yield, so F071 exclusions don't apply on the SSE path. That's a separate concern from this PR, which is scoped to the **crash** (the prod-visible Token error skipping post_turn on every Telegram turn). `recall_exclude_context_ids` is off-by-default/dark and was only ever validated on the non-streaming run_turn path. I've documented the streaming no-op loudly at the set site (badb9ff) and made the flag-on test's scope explicit (crash regression, not exclusion-applied) rather than silently shipping a broken-when-enabled path. Making streaming exclusions actually apply needs the contextvar set in rest.py's parent context (or exclude-ids threaded explicitly) — tracked as a follow-up. Please re-review the crash fix.

Issue #381, opened 2026-04-29
Dashboard orphan_trend SQL counts nodes-per-day instead of first-connected-at, causing health card to show 0 orphans while density card shows real counts


tfatykhov:
## Symptom

\`/dashboard/health\` orphan trend reports 0 orphans (latest day) while \`/dashboard/density\` correctly reports 1042 total orphans across types:

| Type | Total | Orphans | Rate |
|---|---|---|---|
| fact | 1,630 | 620 | 38.0% |
| decision | 443 | 110 | 24.8% |
| episode | 489 | 297 | 60.7% |
| procedure | 57 | 15 | 26.3% |
| **Total** | **2,619** | **1,042** | **39.8%** |

## Root cause

\`nous/api/dashboard_queries.py:791-800\` — the \`daily_new_connected\` CTE name (\`first_seen\`) is misleading; it actually counts \`(day, node_id)\` pairs where ANY edge for that node was created on that day:

\`\`\`sql
daily_new_connected AS (
    SELECT day, COUNT(DISTINCT node_id) AS cnt
    FROM (
        SELECT CAST(created_at AS date) AS day, source_id AS node_id
        FROM brain.graph_edges WHERE agent_id = :agent_id
        UNION
        SELECT CAST(created_at AS date) AS day, target_id AS node_id
        FROM brain.graph_edges WHERE agent_id = :agent_id
    ) first_seen
    GROUP BY day
)
\`\`\`

When a node accumulates edges across multiple days (e.g., sleep F040 backfill on accumulated orphans, plus subsequent cluster_consolidation, plus subsequent F022 cross-type linking), it's counted **once per day**.

The downstream \`SUM(new_connected) OVER (ORDER BY day)\` then overcounts; eventually \`cumulative_connected > cumulative_total\` and the \`GREATEST(..., 0)\` at \`dashboard_queries.py:822-826\` clamps to 0 — masking the real orphan count.

The clamp is silent: nothing logs the negative-clamp event, so the dashboard appears \"healthy\" even though density card concurrently shows 39.8% of nodes are orphans.

## Reproducer

\`\`\`bash
# Connect to prod DB
psql -h 192.168.1.141 -U nous -d nous -c \"
SELECT count(*) FROM heart.facts WHERE active=true
   AND NOT EXISTS (
     SELECT 1 FROM brain.graph_edges e
     WHERE e.agent_id = heart.facts.agent_id
[...]

PR #467, opened 2026-05-29
fix(#381): orphan_trend SQL counts each node once at first-connected-at

tfatykhov:
## Problem
`get_health_data`'s `daily_new_connected` CTE counted a node once **per day it had any edge**, so a node that gained edges across multiple days inflated the cumulative connected-node sum past the cumulative total. `GREATEST(..., 0)` then clamped `orphan_count` to 0 — the `/dashboard/health` card showed **0 orphans** while `/dashboard/density` reported the real ~40% orphan rate (#381).

## Fix
Replace the CTE with a `MIN(created_at)`-per-node **first-connected-day** calc so each node is counted exactly once, on the date its first edge appeared. `pre_period.connected_base` already counts distinct nodes connected before the window, so the two compose without double-counting. Single-CTE change.

## Test
Adds `postgres_only` regression `test_orphan_trend_counts_each_node_once_across_multiple_days` asserting `orphan_trend[-1].count == total_orphans`. Verified RED (`assert 0 == 1`) on the old CTE → GREEN after the fix. All 22 dashboard/density tests pass against Postgres. Registers the `postgres_only` pytest marker.

### DIFF
--- a/nous/api/dashboard_queries.py
+++ b/nous/api/dashboard_queries.py
@@ -809,15 +809,23 @@ async def get_health_data(session: AsyncSession, agent_id: str) -> dict:
                 GROUP BY CAST(created_at AS date)
             ),
             daily_new_connected AS (
-                SELECT day, COUNT(DISTINCT node_id) AS cnt
+                -- #381: count each node ONCE, on the date its FIRST edge appeared.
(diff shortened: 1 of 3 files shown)


tfatykhov:
@codex review

PR #468, opened 2026-05-29
feat(#377): Leg-1 dedup tiebreaker to fix RRF over-dedup of opposites

tfatykhov:
## Problem
The hybrid-search (RRF) pre-check in `FactExtractor` over-dedups high-lexical-overlap semantic **opposites** ("MRR -5%" vs "+5%"). The dedup eval showed Leg 1 at precision 0.667 / **10 FP**, and a threshold sweep (0.92/0.95/0.97) was **byte-identical** — RRF can't separate opposites from true paraphrases, and a cosine gate just collapses Leg 1 into Leg 2 (combined F1 0.782 → 0.765). The only separating signal is semantic.

## Fix
Flag-gated tiebreaker — `NOUS_FACT_DEDUP_TIEBREAKER_ENABLED` (**default off**, land dark):
- `FactManager.is_distinct_fact()` — purpose-built same-vs-distinct Haiku classifier (reuses `contradiction_model` + `call_background_llm_structured`), runs **only** when the RRF pre-check already flagged a dup. Fails open (None/error/flag-off → prior dedup behaviour) so a tiebreaker outage never blocks a learn.
- Both `FactExtractor` dedup sites (LLM-extraction **and** `candidate_facts` — the path the prod summarizer uses) route through a shared `_confirm_dedup` helper so they can't drift (cf. #354).
- Eval-harness wires the LLM client onto `heart.facts` so the dedup eval can measure it.

> Not the supersession classifier (`_classify_fact_pair`): it outputs UPDATE/CONTRADICTION/REFINEMENT/UNRELATED with no SAME label, so paraphrases map ambiguously.

## Result (nous_eval.handlers.dedup)
| | precision | recall | F1 | FP |
|---|---|---|---|---|
| Leg 1 baseline | 0.667 | 1.000 | 0.800 | 10 |
| **Leg 1 + tiebreaker** | **0.950** | **0.950** | **0.950** | **1** |

Combined 0.782 → **0.857**. Prod runs `dedup_via_search=True` = Leg 1, so the production path is fixed.

## Tests
5 unit tests (`tests/test_f377_dedup_tiebreaker.py`) covering verdict mapping + both fail-open paths; fact tests pass on Postgres.

## Rollout
Land dark; flip `NOUS_FACT_DEDUP_TIEBREAKER_ENABLED=true` in prod `.env` after deploy.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #470, opened 2026-05-29
fix(#469): dedup top-hit decided at limit=1 (RRF score-coupling P2-5)

### LINKED ISSUE #469: F377 follow-up: dedup-search widening (limit 1→5) shifts RRF scores, can miss single-channel dups
## Context
Follow-up from the codex review of #468 (F377 dedup tiebreaker, merged). Non-blocking P2; the feature is **flag-gated and default off**, so this only matters **before flipping `NOUS_FACT_DEDUP_TIEBREAKER_ENABLED=true`**.

## P2-5 (valid) — RRF scores are limit-dependent
`nous/handlers/fact_extractor.py::_resolve_dedup` widens `search_facts(limit=1 → 5)` when the tiebreaker is on (to examine lower-ranked hits, fixing the original P2-1). But `nous/heart/search.py:91` uses `penalty_rank = limit + 1` for single-channel hits, and the normalization divides by a limit-independent `1/k` — so a **vector-only** (single-channel) hit's normalized score *drops* as `limit` grows:

- k=30, vector-only hit: limit=1 → `30/32 ≈ 0.9375`; limit=5 → `30/36 ≈ 0.833`.
- With `fact_dedup_threshold=0.92`, that hit is **above** threshold at limit=1 (deduped) but **below** at limit=5 (stored) → the multi-hit path can miss a single-channel duplicate that legacy dedup would have caught.

Dual-channel dups (high keyword + vector overlap) are unaffected, which is why the dedup eval recall held at 0.95. Net combined F1 still rose 0.782 → 0.870, so flag-on is net-positive — but this edge should be closed before the flag is the default.

### Fix options
1. Decouple the threshold decision from the search width: resolve the top-hit dedup verdict at `limit=1` scoring (legacy), and only *additionally* scan widened hits for the tiebreaker, OR
2. Score candidates with a limit-independent measure for the threshold comparison (e.g. native cosine of the candidate vs each hit) instead of the limit-sensitive RRF score, OR
3. Use a fixed search width regardless of flag and re-derive a stable score.

## P2-4 (verify — likely intended)
Codex noted `_supersede_by_subject` (facts.py:747) runs before contradiction detection and doesn't honor `exclude_ids`, so a tiebreaker-DISTINCT fact could still have its predecessor deactivated as superseded. On inspection that path has its own UPDATE-vs-CONTRADICTION gate (facts.py:814) — it supersedes only on a genuine *update* (correct) and lets contradictions coexist. Likely intended F027 behavior, not a regression. Confirm with a targeted test (UPDATE → old superseded; CONTRADICTION → both active) when addressing P2-5.

## Acceptance
- [ ] `_resolve_dedup` multi-hit check no longer misses a single-channel above-threshold dup vs the `limit=1` baseline (add a regression test with a vector-only hit near threshold).
- [ ] Supersession behavior under a DISTINCT verdict covered by a test (UPDATE vs CONTRADICTION).
- [ ] Re-run `python -m nous_eval.handlers.dedup` (flag on) — Leg-1 F1 ≥ 0.974 retained.

tfatykhov:
## Problem (#469 / codex P2-5 on #468)
`_resolve_dedup` widened `search_facts` from `limit=1` to `limit=5` when the tiebreaker was on (to scan lower hits). But `nous/heart/search.py:91` uses `penalty_rank = limit + 1`, so a **single-channel** (vector-only) hit's normalized RRF score *drops* as `limit` grows — e.g. at k=30: `30/32 ≈ 0.9375` (limit=1) → `30/36 ≈ 0.833` (limit=5). A rank-1 duplicate above `fact_dedup_threshold=0.92` at limit=1 could fall **below** it at limit=5 and be stored as new — a dup the legacy path would have caught.

## Fix
Two-phase `_resolve_dedup`:
1. **Phase 1** — resolve the **top hit at `limit=1`** (each hit's highest-possible score). Widening later can never regress this rank-1 verdict.
2. **Phase 2** — only when the tiebreaker is on *and* the top was DISTINCT / a different event, widen to `limit=5` to find a lower true-duplicate. Lower hits have no limit=1 baseline, so any RRF penalty there is lost *added* coverage, not a regression.

Flag-off path = a single `limit=1` search → byte-identical legacy behaviour. Extracted `_event_date_differs` helper.

**P2-4 (supersession after DISTINCT):** verified pre-existing F027 behaviour — `_supersede_by_subject` (facts.py:747) has its own UPDATE-vs-CONTRADICTION gate, so it deactivates the predecessor only on a genuine update (correct) and lets contradictions coexist; independent of the tiebreaker, covered by `test_f027_supersession`.

## Validation
- Flag-on: Leg-1 F1 **0.974** (FP 0, recall 0.95) — win held.
- Flag-off: **byte-identical** to the 0.782 baseline (Leg-1 FP 10).
- New regression test mocks limit-dependent RRF scores (0.94@limit=1 vs 0.83@limit=5) and asserts the dup is still caught. 18 tests pass in postgres mode.

Closes #469

### DIFF
--- a/nous/handlers/fact_extractor.py
+++ b/nous/handlers/fact_extractor.py
@@ -162,27 +162,51 @@ class FactExtractor:
         # `is True` so a truthy Mock/duck-typed setting can't widen the search
         # or trip the await-a-mock path (codex P2 / MagicMock footgun).
         tiebreaker = getattr(self._settings, "fact_dedup_tiebreaker_enabled", False) is True
-        # Only widen the search when the tiebreaker can re-judge lower hits;
-        # flag-off keeps the historical single-top-hit behaviour exactly.
-        limit = 5 if tiebreaker else 1
-        existing = await self._heart.search_facts(content, limit=limit)
-        for cand in existing:
-            # search_facts is score-desc; stop at the first sub-threshold hit.
-            if cand.score is None or cand.score <= self._settings.fact_dedup_threshold:
-                break
-            # F075: distinct event_dates = distinct events, never a duplicate.
-            if (
-                candidate_event_date is not None
-                and cand.event_date is not None
-                and candidate_event_date != cand.event_date
-            ):
+        threshold = self._settings.fact_dedup_threshold
+
+        # Phase 1 — decide the TOP hit at limit=1. RRF penalises single-channel
+        # hits by `limit + 1` (search.py), so limit=1 yields each hit's HIGHEST
+        # possible normalised score. Resolving the top hit here means widening
+        # the search later can never drop the rank-1 hit below threshold and
+        # miss a duplicate the legacy (limit=1) path would have caught (#469).
+        top_hits = await self._heart.search_facts(content, limit=1)
+        if not top_hits or top_hits[0].score is None or top_hits[0].score <= threshold:
+            return None, distinct_ids  # no above-threshold candidate → store
+        top = top_hits[0]
+        if not self._event_date_differs(top, candidate_event_date):
+            if await self._confirm_dedup(top.content, content):
+                return top.id, distinct_ids
+            distinct_ids.append(top.id)  # tiebreaker judged the top DISTINCT
+
+        # Legacy (flag off) examines only the top hit — byte-identical behaviour.
+        if not tiebreaker:
+            return None, distinct_ids
+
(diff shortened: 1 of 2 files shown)


tfatykhov:
@codex review

tfatykhov:
@codex review

tfatykhov:
@codex review

PR #472, opened 2026-05-31
feat(F076): co-mention / shared-entity associative graph linking

tfatykhov:
## What

Build graph edges between memories that **name the same entity**, independent of embedding cosine — the associative edge that cosine-only sleep linking structurally misses (the disjoint-bridge "Steve Hillage" orphan case: two facts both mention *Steve Hillage* but embed at cosine 0.43 ≪ the 0.82 fact-fact threshold → no edge). This is the cheap, data-derived version of the association class tinyHippo would build via replay.

## Changes

- **`nous/brain/entity_extraction.py`** (new) — deterministic, Unicode-aware, sentence-bounded proper-noun extractor (possessive + `de'`-apostrophe normalized; token scanner, no regex backtracking risk).
- **`graph_densifier.build_comention_edges()`** — sleep-cycle pass: raw INSERT (`extraction_method='co_mention'`, weight direct so no `related_to` multiplier discount), hub degree-cap + per-node fan-out cap, rarer-entity-first fill, reverse-dup/existing-edge guard, batched, interrupt-aware. Wired into `run_backfill_cycle` (non-fatal: a failure can't mask the sleep phase).
- **migration 054** — extends `graph_edges.extraction_method` CHECK to allow `co_mention`; `edge_provenance` gains the tier (escapes the F065 inferred penalty — it's a direct match, not an inference).
- **Retrieval consumption** — fix-B **seed-score scoring** (`graph_neighbor_seed_score_enabled`, eval-gated default OFF) lets a recovered neighbor inherit its seed's score and clear the vector top-k cutline; **line-403 guard** now lets Path A (Stage 2b) expand from **chunk seeds** (previously blocked on chunk-only retrieval).
- **Config** — `comention_linking_enabled` (default **ON**) + `max_degree` / `max_edges_per_node` / `weight` / `min_entity_chars` / `max_facts_per_cycle` (all `Field`-bounded).
- **Tests** — `tests/test_comention_linking.py` (12: entity extraction edge cases + the seed-score scoring composition incl. co_mention-escapes-penalty + None→fallback); `test_f065_storage.py` updated for the new tier.

## Review

3-agent **spec** review (architect / devil's-advocate / python-pro) + 3-agent **code** review (code-reviewer / silent-failure-hunter / test-analyzer). All must-fixes applied: the red-suite regression, the `seed_score or 0.0` None-coercion that defeated the fallback, the densifier try/except, and the missing pure tests. Feature tests green.

## Validation status (honest)

- ✅ **Mechanism proven** — bare-pipeline recall@k recovers a genuinely vector-missed bridge into top-k via Path A.
- ✅ **Cheap** — ~674 edges / 1.4s on a 7,898-fact corpus; idempotent; caps respected.
- ⚠️ **Whole-system answer-level VALUE not yet cleanly measured.** The public-fact A/B (MuSiQue/2Wiki) is confounded by the LLM's parametric knowledge (Opus answers many bridges without retrieval). A clean verdict needs a **private-fact, retrieval-necessary** harness (tracked as follow-up; it's also the instrument needed to measure tinyHippo).

## Safety / rollout

Co-mention **edge-building** is default-ON, but the retrieval **consumers** (`heart_graph_all_types_enabled`, `graph_adjacency_boost_enabled`, `graph_neighbor_seed_score_enabled`) are default-OFF, so **prod ranking is unchanged** until those are flipped. Every knob is config-reversible. v1 is **fact-only** (chunk co-mention deferred). Note: at scale the ~N edges raise global graph density, which can influence density-gated spreading activation — flagged for the consumer-flag-flip decision.

### DIFF
--- a/docs/features/F076-comention-entity-linking.md
+++ b/docs/features/F076-comention-entity-linking.md
@@ -46,7 +46,7 @@ This feature builds that edge as a real, default-on **sleep** step.
 - `comention_linking_enabled: bool = True`  (NOUS_COMENTION_LINKING_ENABLED) — **default ON**
 - `comention_max_degree: int = 10`          — skip hub entities (> N nodes) to bound noise
 - `comention_max_edges_per_node: int = 20`  — fan-out cap
-- `comention_weight: float = 0.80`          — edge weight
+- `comention_weight: float = 0.90`          — edge weight (was 0.80; the private-fact value harness showed 0.80 lands a fully-vector-missed disjoint bridge at rank 11, 0.90 clears the cutline at rank 7 with zero displacement — Path-A score = seed_score × comention_weight)
 - `comention_min_entity_chars: int = 6`     — entity length floor
 
 ### 5. Retrieval consumption (already built — no new work, but note for the A/B)
--- a/docs/features/F076-comention-entity-linking.md
+++ b/docs/features/F076-comention-entity-linking.md
@@ -1,0 +1,77 @@
+# F076 — Co-mention / Shared-Entity Linking (associative graph layer)
+
+**Status:** spec → review → implement
+**Default:** ON (core capability — the associative graph's whole purpose)
+
+## Problem
+
+Nous builds graph edges **only by embedding cosine similarity** (`graph_densifier.py`
+fact↔fact ≥0.82, cross-type ≥0.55–0.72). Two memories that **name the same entity** but
+aren't embedding-similar enough stay **orphans** — e.g. "Green is Steve Hillage's 4th album"
+and "Miquette Giraudy … with Steve Hillage" share *Steve Hillage* yet have cosine 0.43 ≪ 0.82,
+so no edge forms. This is the root cause of unbridgeable multi-hop retrieval (formation sweep
+finding #1). Cosine links things vector search already finds; the **associative** edge — link
+by *shared mention*, independent of embedding distance — is the HippoRAG mechanism Nous lacks.
+
+Proven on real data (2026-05-30): an injected co-mention edge + Path A (line-403 fix) + fix-B
+seed-score recovered a **vector-missed disjoint bridge** into top-3 (`stage_origin=heart_graph_memory`).
+This feature builds that edge as a real, default-on **sleep** step.
+
+## Design
+
+### 1. Entity extraction — `nous/brain/entity_extraction.py` (new, deterministic, no deps)
+- `extract_entities(text) -> set[str]`: proper-noun phrases (≥2 capitalized tokens), normalized:
+  strip possessive `'s`/`’s`, trailing punctuation, leading stopwords (The/A/In/…). Lowercased.
+  Fixes the `Marie de' Medici` apostrophe break (connector allows `de'`/`de`/`of`/`the`).
+- Pure, unit-tested, reusable. (LLM/NER precision upgrade deferred — note in code.)
+
+### 2. Densifier co-mention pass — `nous/brain/graph_densifier.py`
+- New `build_comention_edges(agent_id)` invoked from `run_backfill_cycle` (runs in **sleep**).
+- For active **facts** and **chunks** (same-type within each): extract entities from content,
+  index entity→{node_ids}, drop hub entities mentioned in > `max_degree` nodes, link node
+  pairs sharing an entity with per-node fan-out cap `max_edges_per_node`.
+- Insert `graph_edges(relation='related_to', extraction_method='co_mention', weight=settings
+  value, auto_linked=true)`, `ON CONFLICT (source_id,target_id,relation) DO NOTHING`.
+  `related_to` ⇒ traversable by Path A / spreading / adjacency; not `contradicts` so spreading
+  won't filter it. Idempotent: a co-mention edge is skipped if any edge already connects the pair.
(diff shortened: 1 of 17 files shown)


tfatykhov:
@codex review

Addressed the P2: synced the `GraphEdge` ORM `ck_edges_extraction_method` CHECK constraint in `nous/storage/models.py` to include `co_mention` (matching migration 054 / `edge_provenance.VALID_METHODS`), so schemas built from SQLAlchemy metadata (test fixtures) no longer reject co-mention edges.

tfatykhov:
@codex review

Addressed P2-B (`compute_graph_density` counting co_mention edges): `nous/brain/spreading_activation.py` now filters `extraction_method IS DISTINCT FROM 'co_mention'` in all three `brain.graph_edges` references of the density CTE (NULL/legacy `extraction_method` rows stay counted). The default-on co-mention builder can no longer push an agent over `spreading_activation_density_threshold` or flip decision retrieval into spreading activation before the documented consumer flags are flipped — so "ranking unchanged by default" now holds. Co-mention edges remain available to their intended consumers (Path A / adjacency / seed-score) once those flags are on. Added `test_density_excludes_comention_edges`.

P2-A (ORM constraint) remains addressed in 98e99a6.

tfatykhov:
@codex review

New commit 48a35c4 raises the `comention_weight` default 0.80 → 0.90. Rationale: F076 co-mention edges only affect retrieval through Path A's seed-score composition (`score = seed_score × comention_weight`), and a private-fact value harness showed 0.80 lands a fully-vector-missed disjoint bridge at rank 11 (one short of top-10) while 0.90 clears the cutline (rank 7) with zero control displacement across a 0.80/0.90/0.95/1.0 sweep. Edges remain inert by default until the consumer flags (`heart_graph_all_types_enabled`, `graph_neighbor_seed_score_enabled`) are flipped. Prior P2-A (ORM constraint, 98e99a6) and P2-B (density-gate co_mention exclusion, 66fccea) remain addressed.

tfatykhov:
@codex review

Addressed both new P2s in fc42fb9:

- **P2-C** (`find_orphans`): co_mention edges no longer mask orphan status — added `extraction_method IS DISTINCT FROM 'co_mention'` to the orphan NOT EXISTS test, so a fact linked only by co_mention still receives the F040 cross-type backfill. NULL/legacy edges still mask (control test added).
- **P2-D** (existing-pair guard): the guard now loads ANY prior fact-fact edge (not just `related_to`), so a co_mention `related_to` edge is never added over a `contradicts`/`supersedes` pair — preventing adjacency-boost / spreading-activation (which filter only `!= contradicts`) from reinforcing contradictory facts.

3 new postgres tests cover both fixes + a positive happy-path control (50 densifier tests green). Prior P2-A (98e99a6), P2-B (66fccea) remain addressed.

tfatykhov:
@codex review

Addressed both new P2s in 3e81ec5:

- **P2-E (chunk linking):** `build_comention_edges` now links chunk↔chunk as well as fact↔fact (same-type only; never fact↔chunk), via a type-generic inner pass — facts from `heart.facts` (active, learned_at order), chunks from `heart.episode_chunks` (id order). Chunks use raw transcript text, so a shared entity the fact extractor dropped still forms a bridge for the chunk-seed Path A expansion. Validated on the eval corpus: 6 fact-fact + 12 chunk-chunk edges built.
- **P2-F (prior-edge lookup):** scoped the existing-edge query to `source_id/target_id = ANY(scanned ids)` for both node types, so it's bounded by the per-cycle scan rather than the agent's total historical edge count.

dry_run sums both types; +1 test (chunk same-type-only); 64 densifier/comention tests green. Prior P2-A..D remain addressed.

tfatykhov:
@codex review

Latest round addressed in e6ba370:

- **P2-G (spreading traversal):** the density-gate fix (66fccea) stopped co_mention edges from *tripping* spreading on; this stops the activation CTE from *traversing* them once spreading is enabled. Added `extraction_method IS DISTINCT FROM 'co_mention'` to the recursive join in `spreading_activation_search` (NULL/legacy rows still traverse). So the default-on builder no longer changes decision retrieval before a documented co_mention consumer (Path A / adjacency / seed-score) is rolled out.
- **P2-H (strongest-seed-wins):** Stage 2b now tracks the appended neighbor and updates it to the strongest non-null seed score when a later seed reaches the same node, instead of first-seed-wins.

Two earlier comments are now **moot**: "size the backfill cap to include chunks" and "use recency when scanning chunks" — chunk↔chunk co-mention was **removed** in ba7d7ee (reconsidered as a noisy proxy; documents will get a distilled connector *fact* via the new F077 spec instead). The builder is fact-only.

+2 tests; 130 green. Prior P2-A..F addressed.

tfatykhov:
@codex review

Addressed in 6b52e2b: the duplicate-neighbor path selection now compares the FULL composed score (extracted into a shared `_score_memory_neighbor()` so dedup selection and final scoring use the identical formula) and, when a later path wins, replaces the stored neighbor's `seed_score`, `edge_weight`, `edge_relation`, and `extraction_method` together — no more combining a later seed with the first path's edge (a path that does not exist). Added a test where weak-seed+strong-edge (0.36) correctly beats strong-seed+weak-edge (0.27) rather than producing the phantom 0.81.

PR #474, opened 2026-05-31
fix(runner): dispatch opus-4.8 tool calls returned with stop_reason='end_turn'

tfatykhov:
## Problem

opus-4.8 turns that want a tool silently do nothing on the **non-streaming** path (subtasks, heartbeat, `/chat`) — the agent appears to "not call tools / 0 tools this turn", returning an empty answer.

Root cause: the tool loop terminates on `stop_reason`. But opus-4.8 with adaptive/interleaved thinking returns `tool_use` blocks while reporting **`stop_reason='end_turn'`** (the message ends with a `thinking` block *after* the tool_use). The loop sees `end_turn`, declares "done", and returns **without dispatching the tool calls**.

```
RTDIAG terminal: stop=end_turn blocks=['thinking','tool_use','tool_use','tool_use','thinking'] text_len=0 turns=0 toolcalls=0
```

This is the deeper cause behind the original symptom; the obsolete-beta-header fix (#473) only covered the **streaming** manifestation. This one is backend-independent (reproduced on both the SDK and httpx clients).

## Fix

Dispatch whenever `tool_use` blocks are present; terminate only when there are genuinely none.
- `_tool_loop` (non-streaming): `has_tool_use = any(b.type=='tool_use'); if not has_tool_use:` (was `if stop_reason != 'tool_use':`)
- streaming loop: `if not tool_calls:` (dropped the `stop_reason == 'end_turn' or`)
- regression test: `stop_reason='end_turn'` + a `tool_use` block must dispatch the tool
- downgraded the RTDIAG `warning` diagnostics to `debug`

## Verification

- 9 tool-loop + 107 runner tests green.
- End-to-end on the **SDK / prod** backend with opus-4.8: pre-fix `turns=0 toolcalls=0` (dropped) → post-fix `turns=1 toolcalls=3` (dispatched, real answers).

Found by the associative-memory faculty eval (`scripts/diag/faculty/`), whose agentic lens was returning empty answers until this was fixed.

### DIFF
--- a/nous/api/runner.py
+++ b/nous/api/runner.py
@@ -1202,7 +1202,11 @@ class AgentRunner:
                             if thinking_text:
                                 all_thinking_blocks.append(thinking_text)
 
-                    if stop_reason == "end_turn" or not tool_calls:
+                    # Dispatch whenever there are tool calls, even with stop_reason='end_turn'
+                    # (opus-4.8 interleaved thinking can end the message with a thinking block
+                    # after the tool_use, flipping the stop_reason). Terminal only when there
+                    # are no tool calls to dispatch.
+                    if not tool_calls:
                         response_text = "".join(text_parts)
                         break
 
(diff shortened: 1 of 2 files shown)


tfatykhov:
@codex review

PR #475, opened 2026-05-31
Associative-memory faculty: experiential co-occurrence + temporal resolution

tfatykhov:
## Associative-memory faculty: experiential co-occurrence + temporal resolution

Builds out the two memory mechanisms a 18-cell capability baseline proved were missing, then validates them — and is honest about where they don't yet generalize. **All new behaviour is behind off-by-default flags; default `recall_deep` output is byte-identical.**

### The arc
We asked: *is the agent's memory an associative faculty, or just similarity search with a graph?* A discriminating 18-cell baseline (private invented entities, two lenses, mechanism attribution) showed: **surfacing is similarity, selection is the LLM, and the graph carried none of the wins** — isolating two real gaps:

- **Gap 1 — Formation:** links are never created from co-experience.
- **Gap 2 — Resolution:** the current value among stale competitors isn't chosen.

### What ships
| Commit | Change |
|---|---|
| formatter | `recall_deep` now renders Path-A **graph-memory neighbours** (were computed but never shown to the agent); `/chat` debug exposes `recalled_*_ids` for mechanism attribution |
| formation | **`build_cooccurrence_edges`** — facts from the same source episode get a `co_occurred` edge (migration 055: new relation + provenance); flag `cooccurrence_linking_enabled` (default off), noise gate, wired into sleep |
| prominence | **interleave graph-memory neighbours into the ranked Heart Memory list** by score with a `[via]` marker, instead of a buried trailing section — the fix that made formation actually reach the agent |
| Gap-2 | **`ContextEngine._resolve_recency`** — same-subject + differing `event_date` → demote+tag the stale one; brings the recall_deep resolver into the **pre-turn injection path** |
| backfill | `scripts/backfill_cooccurrence_edges.py` (one-time full-corpus retrofit, dry-run default, mirrors F076) |
| docs | `docs/research/017` (thesis+audit), `018` (baseline pre-registration+results), `019` (self-contained whitepaper + .docx) |

### Validation
- **18-cell agentic baseline on a live instance, all mechanisms on: c11 no-handle 0/3→3/3, c12 contradiction FAIL→PASS, every strength cell holds, no regression.**
- **Tests:** SQLite (0 regressions vs base), real-Postgres (DB-integrated paths green — `recall_deep`, `build_cooccurrence_edges`, `_resolve_recency`).
- Default-config output byte-identical (11 formatter tests pass).

### Honest limitation (BEAM step-0)
A ~$0 structural check on **BEAM** (independent prod-shaped data) found both new mechanisms **structurally inert as-built**: BEAM episodes hold 7–9 facts (above the formation noise gate → 0 edges), and its contradictions are diffuse/implicit (not same-subject value-swaps → resolver can't fire). **The mechanisms are validated on clean discrete data but are shape-specific; generalizing to prod-shaped long sessions (finer co-occurrence segmentation; diffuse-contradiction handling) is the documented next step.** This is called out in `docs/research/019`.

### Deploy
Flip `cooccurrence_linking_enabled` (sleep handles new conversations) + run the backfill once per agent (retrofits history). `recency_resolver_enabled` for Gap-2. Consumers: `heart_graph_all_types_enabled`, `graph_neighbor_seed_score_enabled`, `graph_adjacency_boost_enabled`.


tfatykhov:
@codex I've addressed both P2 comments in the latest commit:
- **Sticky-superseded recency** (context.py): made `superseded` sticky in `_resolve_recency` so an intermediate version among 3+ dated facts can't be re-tagged `current`. Verified with a 2024/2025/2026 case — only the newest is `current`.
- **Co-occurrence orphan totals** (sleep_handler.py): pop `_co_occurrence` before `sum(result.values())` and store as `cooccurrence_edges`, matching the underscore convention.

Please re-review.

tfatykhov:
@codex Both P2s are already addressed in commit `6cdcfdf` (current PR head) — the re-review appears to describe the pre-fix code. Evidence:

**Recency (context.py):** the fix makes `superseded` sticky and lets `current` be overwritten by `superseded` on the older side — exactly your suggestion. Verified across **all 6 orderings** of a 2024/2025/2026 group, including the non-newest-first case you flagged (`B,A,C`):
```
order=['B','A','C']: current=['bank C'] OK   # 2025 compared with 2024 first, then 2026 → 2025 ends superseded
```
All 6 permutations → exactly one `current` (the 2026 fact). No two-current case remains.

**Sleep metrics (sleep_handler.py):** `_co_occurrence` is now popped before `sum(result.values())` and stored as `cooccurrence_edges` (lines 1368/1373), matching the `_ce_stats`/`_happened_before`/`_co_mention` convention.

Please re-review against `6cdcfdf`.

PR #480, opened 2026-06-05
F078.1: Guarded send_email tool + recipient allowlist

tfatykhov:
## What & why
F078 kept the email censor advisory (`steer`) because email's exfil discriminator is the **recipient**, not the message text — and refuse/abort would break the daily email senders. F078.1 adds the missing chokepoint: a guarded `send_email` tool that validates the recipient at the tool boundary.

## The tool (`nous/api/email_tools.py`, mirrors `send_file`)
`send_email(to, subject, body, cc=None)`:
- **Recipient allowlist** — every `to`+`cc` address must be in `NOUS_EMAIL_ALLOWLIST` (CSV, case-insensitive). **Empty allowlist ⇒ reject all** (safe default — ships inert until the operator configures it).
- **Secret scan** of subject+body (`sk-…`, `AKIA…`, `password:`, `Bearer …`) → reject.
- **Rate limit** (`NOUS_EMAIL_MAX_PER_HOUR`, default 5, monotonic window).
- Sends via `smtplib` in `asyncio.to_thread` (non-blocking).
- **Exception sanitization** — smtplib/login errors are logged server-side but the model only sees a generic `email send failed: <Type>` (never raw creds/host).

## Additive — BC invariant held
The existing agent-authored `bash + python3 smtplib` path is **untouched** (daily email tasks keep working). The `send_email` procedure + frames now steer the agent to the guarded tool, with bash+smtplib documented as the fallback. Fully closing the hole = migrating tasks to the tool (operational).

## Validation
11 tests (smtplib monkeypatched — nothing sent): allowlist accept/reject incl. a bad cc address, empty-allowlist-rejects-all, secret-scan, rate-limit, creds-absent/disabled, and exception-sanitization (raw smtplib text does not leak).

## Rollout (safe by default)
Empty allowlist = can't send. Operator: set `NOUS_EMAIL_ALLOWLIST` (Tim + Maya + daily-task recipients), update the prod `send_email` procedure + the email steer directive to point at the tool, migrate the daily tasks.

Spec: `docs/features/F078.1-email-recipient-allowlist.md`


tfatykhov:
@codex review — CI is green. Latest commit `b6c4262` is a 2-line test fix (asyncio.get_event_loop().run_until_complete → asyncio.run in 2 TestSubtaskConfigDefaults tests; the new email tests' asyncio.run() cleared the thread loop the deprecated pattern relied on). Feature commit 667f4c9 already reviewed clean.

PR #483, opened 2026-06-06
F078.1.3: optional html_body for guarded send_email tool

tfatykhov:
## Summary

- Adds an optional `html_body: str | None = None` parameter to `_build_message` and `html_body: Any = None` to the `send_email` closure, plus the corresponding `html_body` property in `_SEND_EMAIL_SCHEMA`.
- When `html_body` is provided, the body part is constructed as `multipart/alternative` (plain text first, HTML second per RFC 2046 — clients render the last supported part). With attachments, this alternative part is nested inside a `multipart/mixed` envelope exactly as before.
- **BC guarantee**: when `html_body` is `None`/absent the MIME structure is byte-for-byte identical to today — plain `MIMEText` with no attachments, `MIMEMultipart("mixed")` with attachments. All 19 pre-existing tests pass unchanged.
- The secret scan now also covers `html_body` content (`_scan_secrets(f"{subject}\n{body}\n{html_body or ''}")`).

## Unblocks

Scheduled jobs (e.g. `premarket-daily`) that generate styled HTML newsletters currently bypass the guarded `send_email` tool and use raw `bash+smtplib` (no allowlist protection). This change lets those jobs migrate onto the protected path while keeping their rich HTML output.

## Test plan

- [x] `test_html_body_none_plain_text` — BC: no html_body → plain `MIMEText`, `content_type=text/plain`
- [x] `test_html_body_set_no_attachments` — top-level `multipart/alternative`, two parts in correct order (plain → html)
- [x] `test_html_body_set_with_attachment` — top-level `multipart/mixed`, first part is `multipart/alternative`, second is the attachment
- [x] `test_html_body_secret_scan_rejected` — secret in html_body triggers the scan and refuses send
- [x] All 19 existing tests pass (allowlist, secret scan, rate limit, attachments, exception sanitization, hot-reload)

```
23 passed in 2.03s
```

### DIFF
--- a/nous/api/email_tools.py
+++ b/nous/api/email_tools.py
@@ -203,9 +203,9 @@ def create_send_email_tool(settings: Settings):
         to: Any,
         subject: str,
         body: str,
-        html_body: Any = None,
         cc: Any = None,
         attachments: Any = None,
+        html_body: Any = None,
     ) -> dict[str, Any]:
         """Send an email to allowlisted recipient(s).
 
@@ -213,13 +213,13 @@ def create_send_email_tool(settings: Settings):
             to: Recipient address(es) — string or list. Each must be allowlisted.
             subject: Email subject line.
             body: Plain-text email body.
-            html_body: Optional HTML body. When provided, the email is sent as
-                multipart/alternative with the plain ``body`` as the fallback part.
-                Use this for styled newsletter-format emails.
             cc: Optional CC address(es) — string or list. Each must be allowlisted.
             attachments: Optional file path(s) — string or list — to attach (e.g. a
                 generated .docx/.pdf report). Each must be a readable file; the total
                 size must be within NOUS_EMAIL_MAX_ATTACHMENT_MB.
+            html_body: Optional HTML body. When provided, the email is sent as
+                multipart/alternative with the plain ``body`` as the fallback part.
+                Use this for styled newsletter-format emails.
 
         Returns:
             MCP-compliant response confirming the send or naming the rejection reason.
--- a/nous/api/email_tools.py
+++ b/nous/api/email_tools.py
@@ -121,20 +121,30 @@ def _build_message(
     to_list: list[str],
     cc_list: list[str],
     attachments: list[str],
+    html_body: str | None = None,
 ) -> tuple[Any | None, str | None]:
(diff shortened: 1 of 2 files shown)


tfatykhov:
Fixed in 0bf3157. Moved `html_body` to the end of the public `send_email` signature — restores the documented v1 positional contract `(to, subject, body, cc, attachments)` so positional cc/attachments are no longer reinterpreted. Chose append-to-end over keyword-only to keep it byte-for-byte BC with the v1 contract. Docstring Args reordered to match; all 23 tests pass (call sites are keyword-only). Thanks @codex.

tfatykhov:
@codex review

tfatykhov:
## Nous Review — F078.1.3: html_body for guarded send_email

**Verdict: ✅ Clean — merge-ready**

### What this does
Adds optional `html_body: str | None = None` to the guarded `send_email` tool, enabling styled newsletter-format emails (premarket-daily, etc.) to migrate off the unguarded bash+smtplib path. MIME structure:
- No html_body → plain `MIMEText` (byte-for-byte BC with v1)
- html_body only → `multipart/alternative` (plain first, HTML second per RFC 2046 §5.1.4)
- html_body + attachments → `multipart/mixed` wrapping `multipart/alternative` + file parts

### Verified
- ✅ BC preserved: `test_html_body_none_plain_text` confirms plain MIMEText when html_body absent
- ✅ RFC 2046 part order correct (plain before HTML — clients render last supported part)
- ✅ Nesting correct for html+attachment case (alt inside mixed)
- ✅ Secret scan extended to cover html_body content
- ✅ Schema updated with documented html_body field
- ✅ 23/23 tests pass, CI green on both Lint and Tests jobs

### Open item — non-blocking, track separately
**Codex comment (line 253) is valid:** `_scan_secrets()` runs against raw HTML source. An HTML-entity-encoded secret (`sk-&#97;bcdEFGH...`) won't match `_SECRET_PATTERNS` but renders in mail clients. Not a regression (bash+smtplib had zero scanning; allowlist is the primary guard), but a real defense-in-depth gap.

→ Filing follow-up issue before migrating premarket-daily onto this path.

**No P1 blockers. Merging now.**

PR #485, opened 2026-06-07
feat(observability): F079 Phase 0 — populate procedure/episode delivery counters

tfatykhov:
## What
First phase of **F079 — Memory Delivery Overhaul** (`docs/features/F079-memory-delivery-overhaul.md`): fix the context-log telemetry that was **structurally blind to procedure/episode delivery**.

Two defects (both code-verified):
- `loaded_procedures` / `loaded_episodes` / `recent_conversations` were **never assigned** in `ContextLogEntry.from_payload` (only facts + decisions were) → the columns (which already exist in `to_dict` + the INSERT) were always `0`.
- `parse_system_sections` matched a hardcoded `SECTION_MARKERS` with **no `## Known Procedures` / `## Past Episodes` entry** → those keys could never appear in `sections_present`, even when injected.

Together these hid that procedures/episodes are gated out of the prompt, and caused two wrong measurements during the F079 investigation.

## Change (pure observability — no schema / flag / migration; reversible)
- add the two missing markers to `SECTION_MARKERS`;
- populate the three counters in `from_payload` via the existing `count("\n-")` per-item heuristic.

The system prompt sent to the LLM is unchanged; this is a telemetry parse-bucket correction.

## Validation
- **Unit:** exact-count tests (3 procedures / 2 episodes / 4 recent) + absent-section zero case + `to_dict` round-trip — `tests/test_context_logger.py`, 37 pass.
- **Live (real eval instance):** a `task`-frame turn now logs `loaded_procedures=4` with `known_procedures` present in `sections_present` (previously always `0` / absent).

## Why first
This is the measurement substrate the later F079 phases depend on — you can't tell whether any delivery fix works while these signals read 0.

### DIFF
--- a/nous/api/runner.py
+++ b/nous/api/runner.py
@@ -88,23 +88,6 @@ def _build_exclude_ids(
         "procedure": set(turn_context.recalled_procedure_ids or []),
     }
 
-
-def _recalled_counts(turn_context) -> dict[str, int]:
-    """F079 Phase 0: exact per-section item counts from the turn's selected objects.
-
-    Authoritative source for the context-log delivery counters — derived from the
-    `recalled_*_ids` that `ContextEngine.build` populated at injection time, so the
-    telemetry never relies on parsing the rendered prompt (codex PR #485).
-    """
-    if turn_context is None:
-        return {}
-    return {
-        "facts":      len(turn_context.recalled_fact_ids or []),
-        "decisions":  len(turn_context.recalled_decision_ids or []),
-        "episodes":   len(turn_context.recalled_episode_ids or []),
-        "procedures": len(turn_context.recalled_procedure_ids or []),
-    }
-
 # Frame-gated tool access (D5)
 FRAME_TOOLS: dict[str, list[str]] = {
     "conversation": ["record_decision", "learn_fact", "learn_skill", "recall_deep", "recall_recent", "recall_hubs", "get_procedure", "create_censor", "bash", "read_file", "write_file", "web_search", "web_fetch", "cache_retrieve", "spawn_task", "spawn_sync", "schedule_task", "list_tasks", "cancel_task", "run_python", "send_file", "send_email", "heartbeat_check_create", "heartbeat_check_manage", "dag_create", "dag_manage", "ingest_document"],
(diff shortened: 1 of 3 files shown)


tfatykhov:
@codex review

tfatykhov:
@codex review — addressed the P2 (count only top-level item bullets via type-specific markers; facts/decisions unchanged; added a regression test for embedded `\n- ` in descriptions/summaries).

tfatykhov:
@codex review — addressed the P2 properly this time: counts are now derived from the turn's selected objects (recalled_*_ids threaded into the logger), not parsed from the rendered prompt. Text parse is only a fallback for no-TurnContext calls, and is gated on the section being present in that prompt. Added tests for the marker-shaped-embedded case + the stale-count guard; live-verified on the eval instance.

tfatykhov:
@codex review — addressed the P1: reverted the recalled_*_ids threading (you're right — they're collected pre-`_truncate_to_budget` and include `_format_episodes`-skipped abandoned episodes, so they over-state delivery; it also broke a Mock-based test). Counts now come from the RENDERED (post-truncation) section, which is exactly what the model received. The residual embedded marker-shaped-line case is an accepted, documented telemetry approximation — identical to the limitation the pre-existing facts/decisions counters have always had — out of scope for a pure observability fix. runner.py is back to no-diff-vs-main.

tfatykhov:
Merging on green CI + resolution of all substantive codex findings. **P1 (pre-truncation `recalled_ids` over-count)** and the **'count after truncation' P2** are both resolved by `ade1995`, which counts the **rendered, post-truncation** section — exactly what the model received (the approach codex's P1 recommended). The remaining embedded marker-shaped-line P2 is an **accepted, documented telemetry approximation, identical to the pre-existing facts/decisions counters** in the same function — out of scope for a pure observability fix (chasing it would require either over-counting via IDs or changing the LLM-facing formatter). Net change: +2 SECTION_MARKERS, +3 previously-unpopulated counters; runner.py no-diff-vs-main. 24 logger tests + the 2 previously-failing subtask tests pass; live-validated on the eval instance (loaded_procedures=4).

PR #488, opened 2026-06-07
feat(procedures): dedup Phase 0 + create/import fixes (consolidation, loop-close, learn_skill, parser)

tfatykhov:
Executes the procedure-subsystem audit (`docs/reviews/procedure-subsystem-audit-2026-06-06.md`, §6 authoritative; §7 = execution). Collapses the duplicate clusters the F079 catalog would faithfully list **and** fixes the create/import paths that let duplicates/broken rows accumulate. Self-contained and deploy-safe — the only piece that must wait for the prod consolidation is the unique index, shipped separately in #489.

## Dedup execution
- **migration 057** — nullable `superseded_by` + `archived_at` (+ redirect index). Additive, splitter-safe for the auto-migrator (no `DO $$`, no `BEGIN/COMMIT`), plain UUID.
- **B1 restart-loop close** — bootstrap skips re-importing a consolidated name; `reactivate_skills` no longer un-archives superseded rows. Keyed on `superseded_by`.
- **`scripts/diag/proc_consolidate.py`** (operator-gated `--dry-run`/`--commit`) — per cluster, one tx: UPSERT-SUM affinity into canonical → **fold procedure-level outcome counters** → delete redundant sibling affinity → enrich canonical → delete incident `graph_edges` → archive siblings. Idempotent; refuses non-null outbound arrays. Canonicals: email→`send_email`, action_gate→`recovery`, compaction→`Management`; context trio kept. **Net 62 → 55.**

## Create/import algorithm fixes (audit §6 PR2)
- `learn_skill` **updates in place** on a name collision (`Heart.update_procedure_body`) instead of retire+store, which reset counts and orphaned task affinity (bug 9).
- **SkillParser handles YAML block scalars** (`description: |` / `>`) — root cause of the empty one-char catalog descriptions on heavily-used skills (`deep-researcher`, `investigate`, `cso`, `office-hours`). `backfill_proc_descriptions.py` fixes the 4 existing rows (idempotent, gated).
- case-insensitive `get_procedure_by_name`/`is_name_superseded`; reactivate name-collision precheck; embedding-failure retry that logs ERROR + stores NULL (store **and** update paths).

## Validation
Through the **real auto-migrator splitter** on a throwaway DB (5433) seeded from prod cluster data: 057 applies; consolidation sums affinity (388) **and** counters (send_email 112+655=767) correctly and is re-run **idempotent**; siblings archived w/ `superseded_by`; edges/affinity gone. 77 tests green. Addresses codex P1 (fold counters; route `learn_skill` through `update_body`) + P2 (NULL stale embedding).

## ⚠ Deploy ordering
Deploy this PR → run `proc_consolidate.py --commit` + `backfill_proc_descriptions.py --commit` on prod → then merge/deploy **#489** (the unique index, which fails while active dups exist).

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -19,3 +19,10 @@ nous.db
 # gen_nous_procedure_qrels.py against your own nous-prod-snapshot.
 tests/fixtures/qrels_nous_prod*.jsonl
 tests/fixtures/qrels_longmemeval.jsonl
+
+# F074: BEAM harness — sibling clone of mohammadtavakoli78/BEAM
+tools/beam/
+# F074: per-run reports
+reports/beam/
+# F074: HuggingFace cache (~/.cache/nous-eval/beam) is outside repo
+scripts/diag/_proc_*.json
--- a/nous/api/tools.py
+++ b/nous/api/tools.py
@@ -958,18 +958,19 @@ def create_nous_tools(brain: Brain, heart: Heart, settings: Settings | None = No
             missing_requires = [var for var in manifest.requires if not _os.environ.get(var)]
             skill_active = len(missing_requires) == 0
 
-            # 3. Check for existing procedure with same name (dedup)
+            # 3. Check for existing procedure with same name (dedup, case-insensitive)
             existing = await heart.get_procedure_by_name(manifest.name)
-            updated = False
-            if existing:
-                await heart.retire_procedure(existing.id)
-                updated = True
+            updated = existing is not None
 
-            # 4. Convert to ProcedureInput and store
+            # 4. Convert to ProcedureInput and store / update in place
             proc_input = _skill_parser.to_procedure_input(manifest)
-            if not skill_active:
-                proc_input.active = False
-            result = await heart.store_procedure(proc_input)
+            proc_input.active = skill_active
+            if existing:
+                # Update in place to preserve activation/success/failure counts and
+                # task affinity (retire+store reset them — audit bug 9).
+                result = await heart.update_procedure_body(existing.id, proc_input)
+            else:
+                result = await heart.store_procedure(proc_input)
 
             action = "updated" if updated else "registered"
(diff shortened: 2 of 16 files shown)


tfatykhov:
@codex review — restructured: all code + create/import fixes now land here (learn_skill routed through update_procedure_body), only migration 058 is split to #489. Folded procedure-level counters into the canonical and NULL stale embedding on re-embed failure per your P1/P2.

tfatykhov:
Thanks — re-reviewed against `38c7a5f`:

**Prior P1s are addressed in this commit** (the two re-anchored P1 comments are stale re-posts):
- `learn_skill` now routes through `heart.update_procedure_body` on a name collision (`nous/api/tools.py:968-972`) — `retire_procedure`/`store_procedure` are no longer called on the existing-name path.
- The consolidation folds procedure-level `activation/success/failure/neutral` into the canonical (`scripts/diag/proc_consolidate.py` step 2b), gated on `superseded_by IS NULL` so it's idempotent. Verified empirically: `send_email` 112+655=767, re-run stays 767.

**New P2s — leaving as-is, by design:**
- *Clear embeddings when no provider configured* — in prod an embeddings provider is always present, so the `if self.embeddings` branch always recomputes (NULL on failure, as you asked). The no-provider branch is degraded/test-only; nulling a valid persisted embedding just because *this* process lacks a provider would make the row invisible with no way to recompute — keeping the (searchable) vector is the lesser evil there.
- *Remove obsolete graph links before relinking* — `update_procedure_body` deliberately mirrors `store_procedure` (emit `procedure_stored` → F040 linker). Neither prunes first; that's pre-existing behavior, and graph blast-radius is ~0 in prod (F065 `graph_expansion_used=0`, F053 prune is on). Worth a follow-up for the densifier, not this PR.

PR #490, opened 2026-06-07
feat(procedures): single-deploy dedup — consolidation migration (058) + name-uniqueness index (059)

tfatykhov:
Follow-up to #488 so the dedup completes in **one deploy** (you asked: merge both, deploy once, run consolidation). The original blocker was that the unique index auto-runs on deploy *before* the operator could run the consolidation script. Fix: ship the one-time consolidation **as a numbered data migration**.

On deploy, `run_migrations` applies in order — all **before** bootstrap runs (`main.py:52` vs `:467`), so #488's loop-close keeps archived skill rows from being re-imported:
```
057 (columns, #488) → 058 (consolidate + backfill) → 059 (unique index)
```

## Contents
- **058 — consolidation data migration**, generated by `proc_consolidate.py --emit-sql` from the same validated `CLUSTERS` source (single source of truth). Per cluster: UPSERT-SUM affinity into the canonical → fold procedure-level counters → delete redundant sibling affinity → enrich → delete incident `graph_edges` → archive siblings (`active=false` + `superseded_by`). Plus the 4 block-scalar description backfills. Inline literals, splitter-safe (no `DO $$`, no `BEGIN/COMMIT`), scoped to `nous-default` + known ids → **no-op on fresh installs / CI**. Idempotent via `superseded_by` guards.
- **059 — the `(agent_id, lower(name)) WHERE active` unique index** (was standalone #489/#490), now guaranteed collision-free because 058 ran first in the same deploy.
- `proc_consolidate.py` gains `--emit-sql`; the `--dry-run/--commit` script is kept for ad-hoc use.

## Validation
Through the real auto-migrator splitter on a throwaway DB seeded from prod cluster data: `057→058→059` applies in order; **62 → 55 active**, `send_email` counters 112+655=767, affinity summed, the 4 descriptions fixed, and the index then rejects a second active `send_email`. Re-verified the whole chain is a clean **no-op on a fresh empty DB** (CI-safe).

## Deploy
Merge this → deploy `main` once. Migrations do the consolidation + index automatically. Afterwards the F079 catalog prod-flag (`NOUS_PROC_CATALOG_ENABLED`) is unblockable (gate the flip on a separate selection-accuracy run).


tfatykhov:
@codex review — reworked for single-deploy: the one-time consolidation is now migration 058 (generated from the validated proc_consolidate CLUSTERS, splitter-safe, no-op on fresh DBs) running before the 059 unique index in the same deploy; loop-close (#488) runs after migrations so no resurrection.

tfatykhov:
Addressed:
- **P1 (cross-agent collisions)** — 058 now ends with a generic safety net: archive all-but-newest active row per `(agent_id, lower(name))` across **all** agents (reversible). Validated with an injected other-agent `Foo Skill`/`foo skill` pair — 058 archives the older, 059 creates cleanly, nous-default end-state unchanged.
- **P2 (preflight / outbound arrays)** — leaving as-is: this is archive-not-delete, so a sibling's `censor_ids`/`related_procedures` are never lost — they remain on the (now inactive) row, queryable and restorable. All targeted siblings are verified NULL on these columns, so there is nothing to union onto the canonical. The script's preflight *refused* because it could; the migration can't refuse mid-deploy, but it also can't lose the data.

tfatykhov:
Verified against live prod (read-only):
- **Only one agent exists** — `nous-default` (92 rows total).
- **The only active `(agent_id, lower(name))` collision is `send email via gmail smtp` ×3** — exactly the email cluster 058 consolidates.

So on the real target: after 058's cluster step there are **zero** remaining active collisions, the safety-net step archives **nothing** (pure no-op), and 059 creates cleanly. P1[058:54] (merge state before archiving safety-net rows) cannot fire — there are no cross-agent or out-of-cluster dups. And once 059's unique index exists, no new active collision can form, so the safety net never has work to do on subsequent boots either. The two re-posted items (059:18 / 058:21) are addressed above. Merging.

PR #494, opened 2026-06-09
feat(skills): F081 — persist + render full skill bodies (fix lossy import)

tfatykhov:
### Root cause
`SkillParser.to_procedure_input` stored only the **first H2 section** (`manifest.first_section`) in `implementation_notes`; the full body (`manifest.raw_content`) was parsed but dropped. Multi-section skills imported as stubs — **29/55** prod procedures were compressed (e.g. `investigate`'s multi-section gstack source → ~250 chars). 91.7% of routing/instruction signal lives in the body (decision 199ae559).

### Fixes
- **parser.py** — persist the full body (`raw_content`), `first_section` only as fallback. (Single conversion path — covers `learn_skill` + `bootstrap`.)
- **context.py `_format_procedure_bodies`** — render the real body (`implementation_notes` minus `source:`/`version:` metadata) with newlines preserved; **drop `core_patterns`** (trigger keywords) and `core_tools` (keyword dump) — matcher metadata, not instructions; cap oversized bodies with a `get_procedure('<name>')` pointer to the untruncated full skill.
- **config** — `proc_recommended_body_max_chars` 1200 → 2500 (fit a real skill body; total still bounded by the per-frame procedure budget).
- **scripts/diag/reimport_skills.py** — operator dry-run/commit: re-import skills from their source (URL/local) to restore full bodies via `update_procedure_body` (preserves stats + re-embeds), and archive (reversible) non-recoverable stubs.

### Validated on the fresh prod snapshot
- **16 re-imported** to full fidelity — `moltbook` 1.6K→**33K**, `tool-design`→19K, `multi-agent-patterns`→18K, `xlsx`→10K, etc.
- **11 source-gone-but-functional** (≥300 ch) kept.
- **21 source-gone stubs** (<300 ch) archived (reversible).
- Rendered body sanity: `multi-agent-patterns` → 2,576 chars of real skill, no `Patterns:` noise, `get_procedure` pointer present.

### Notes
- The prod data fix (re-import + archive) runs via the script on prod (operator step); the parser fix prevents future truncation; both flag-independent.
- The 21 archived are reversible (`active=false`) and re-learnable from public repos at full fidelity now that the parser is fixed.

72 tests pass (parser full-body-persist, body-render-drops-keywords, truncation-pointer).

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #495, opened 2026-06-09
fix(memory): storage/retrieval integrity fixes + embedding cache (2026-06-09 audit)

tfatykhov:
## Summary

Fixes the P1/P2 defect families from the [2026-06-09 memory storage/retrieval deep analysis](docs/reviews/memory-storage-retrieval-deep-analysis-2026-06-09.md) (committed in this PR), plus the embedding-cache optimization.

### Storage write path
- **S1/S2 (P1):** Leg-1 fact dedup now probes via **raw cosine** (`Heart.find_similar_facts`) instead of RRF rank scores. RRF encodes rank, not closeness — the nearest fact scored ~0.98 at limit=1 regardless of similarity, so the 0.92 pre-check fired for *every* candidate: the Haiku tiebreaker was the only real gate (tiebreaker outage = silent loss of all candidate facts), and `knowledge_extractor` stored ~nothing. `NOUS_FACT_DEDUP_THRESHOLD` is now a cosine threshold (CLAUDE.md updated).
- **S3 (P1):** dupes in the contradiction band (0.85–0.95) are classified (F027) **before** confirming. At prod's `NOUS_FACT_NATIVE_COSINE_THRESHOLD=0.80`, a conflicting fact previously *incremented confirmation_count on the stale fact* and was discarded. Verdicts route: UNRELATED/REFINEMENT/UPDATE-new/CONTRADICTION → insert + post-flush action; UPDATE-old/failure → confirm (fail-open preserved). `event_date` now merges onto undated duplicates (was silently de-dated).
- **S10:** `_find_duplicate` rewritten HNSW-servable (plain `ORDER BY distance LIMIT 20`, threshold + F075 date preference in Python) — was a per-learn sequential scan.
- **S11:** `exclude_ids` threaded into `_supersede_by_subject` so DISTINCT-ruled facts can't be silently superseded.
- **S9:** dedup probes no longer inflate `recall_count`/`last_recalled_at`.
- **E1 (P1):** F067 transcript chunks allocate from MAX+1 under the same advisory lock as F069 `ingest_document` — a document ingested mid-session no longer silently destroys transcript chunks via `ON CONFLICT DO NOTHING`. Same fix applied to `scripts/backfill_episode_chunks.py` (third writer).
- **E2 (P1):** F055 `record_surfaced` **merges** into working memory (was wholesale-replacing curated items with `"residual fact"` stubs that rendered into the next prompt); entries now carry real content snippets; read-side decay window restored.
- **D1 (P1):** decision delete removes its graph edges (the CASCADE comment was stale since migration 016 dropped the FKs); **migration 060** cleans up historically dangling edges + adds the `lower(subject)` expression index (D4).
- **D9:** censor vector queries rewritten distance-ordered + `SET LOCAL hnsw.ef_search=120` so the HNSW index serves the guardrail path without ANN recall loss.

### Cache
- **D2/S7:** bounded LRU on `EmbeddingProvider` (`NOUS_EMBEDDING_CACHE_SIZE`, default 1024, keyed model+dims+sha256) — eliminates the 4–7× repeat query embeds per recall and repeat content/template embeds per learn + sleep cycle. `embed_batch` requests misses only, with a length guard so a short API response can never cache misaligned vectors.
- `search.py` Settings resolvers cached per env-fingerprint (were re-constructing pydantic Settings 2–3× per hybrid_search).

### Retrieval
- **R1:** §14 cosine procedure fallback applies `procedure_score_floor` before preloading full bodies (was floorless — up to 5×2500-char bodies injected per turn for arbitrarily-poor matches). Lands **before** #494 deploys, preventing the regression going live.
- **R2:** `recall_deep` docstring + schema now state the F080 knowledge-only default pool.

## Validation
- **12/12 live probes** on the 5433 eval clone (`scripts/diag/validate_memory_integrity.py`), incl. a live-Haiku S3 conflict probe at prod threshold 0.80 (conflicting fact stored, stale fact NOT confirmation-bumped) and an EXPLAIN check that the dedup probe is index-served. Migration 060 applied through the app migrator.
- **recall_deep A/B:** byte-identical output main vs branch on the prod-clone corpus (5 queries, expansion disabled).
- **Full suite:** failure set identical to main's pre-existing local-env noise (181 baseline failures both sides, from the repo `.env`; CI has no `.env`). 9 initial regressions (test mocks of the old probe) fixed.
- **3-agent review** (code-reviewer / silent-failure-hunter / test-analyzer): all P1/P2 findings addressed — embed_batch length guard + dup-miss counter, censor ef_search, dedup window 5→20 + docstring caveat, `float(confidence)` fail-open guards (both band sites), WM activation coercion + exc_info, §14 `break`→`continue`, backfill-script E1 parity, S11/_find_duplicate-selection/wiring/orchestration test gaps closed.

## Deferred (documented in the audit, separate follow-ups)
Transaction restructuring around LLM calls (S5/S6), graph taxonomy/weight-space coherence (G1–G8), episode lifecycle E3–E8, whitepaper carry-overs (B5 stats threading, recency-resolver unification).

## Operator notes
- Migration 060 runs at deploy (dangling-edge cleanup + new index).
- `NOUS_FACT_DEDUP_THRESHOLD` semantics changed RRF→cosine (default 0.92 kept; prod doesn't override it).
- New env var: `NOUS_EMBEDDING_CACHE_SIZE` (default 1024; 0 disables).


tfatykhov:
@codex review

tfatykhov:
@codex review — pushed fixes for all three findings: enforcement-path censor matching reverted to exact scan (P1), ef_search=100 + documented iterative_scan caveat on both dedup probes (P2), carried residuals now ranked by turn-decayed activation via the activator's own decay model (P2). Also fixed the CI test (cosine range assertion).

tfatykhov:
@codex review — round-2 findings addressed: confidence floor (0.5, UPDATE keeps 0.8) on every band verdict; working-memory merge bounded at the row's max_items with curated priority; backfill script attempts each episode once per run.

tfatykhov:
@codex review — round-3 addressed: confidence gate extended to _find_contradiction (falls through to the plain warning below 0.5); backfill exclusion moved into SQL before the LIMIT; pre-flight count now uses the dialogue-only predicate.

tfatykhov:
@codex review — round-4 addressed: §14 leg now probes actual cosine (new find_similar_procedures) so the floor is calibrated; structural chunk adjacency gated on same source_kind; curated/residual ref_id dedup in WM; --repair-dialogue mode for collision-damaged episodes.

tfatykhov:
@codex review — round-4.5 addressed: repair mode purges the replaced chunks' graph edges in-transaction; structural adjacency gated on source_ref as well; ef_search widening applied to the procedure probe.

tfatykhov:
@codex review — round-5 addressed: cache vectors packed (array('f'), ~6 MB at default size, float4-equivalent); backfill exclusion set tracks failures only (repair mode keeps all processed ids).

tfatykhov:
@codex review — round-6 P1 addressed: set_local_ef_search now enables hnsw.iterative_scan = strict_order when available (probed via current_setting missing_ok; pgvector 0.8.1 in this stack), making the filtered dedup scan exhaustive in distance order up to the LIMIT. Full-exact scan was rejected deliberately: the per-learn sequential scan is the audit S10 defect this PR removes; strict iterative scan is the pgvector-documented remedy for filtered-ANN completeness.

tfatykhov:
@codex review — round-7 addressed: iterative_scan is now SET directly (no availability probe) under a savepoint; pre-0.8 errors roll back to the savepoint and the transaction degrades to the ef_search margin.

PR #496, opened 2026-06-09
fix(audit): complete Nous code audit + 9 verified P1 fixes (2026-06-09)

tfatykhov:
## Summary

Complete code-driven audit of every major Nous subsystem (11 parallel subsystem-scoped reviews + master synthesis), followed by a batch of 9 verified P1 fixes with regression tests.

Audit docs live in `docs/reviews/*-audit-2026-06-09.md`; the master synthesis is `docs/reviews/nous-complete-audit-2026-06-09.md`. Code-only, file:line evidence, reachability judged against the prod `.env` overlay. Every load-bearing P1 was personally re-verified against source (one agent claim — AT-2 — was refuted on verification and the doc corrected).

## Three dominant defect patterns

1. **RRF rank score compared to a similarity threshold** — the hybrid-search top hit scores ~0.95 for *any* query. PR #495 fixed this at fact-dedup sites; it survived elsewhere.
2. **F036 cache-split ignores the flat `system_prompt`** — F078 fixed censors by writing to both the flat string and the tiered sections; two producers still wrote only the ignored flat string.
3. **Severed self-improvement / telemetry wires** — built, flag-ON in prod, no consumer.

## Fixes in this PR (9)

| ID | Fix |
|----|-----|
| **HD-1** | `procedure_learner._is_duplicate` used the RRF rank score (~0.95) vs a 0.85 threshold → **all auto-procedure creation suppressed** (prod "K-line = 0"). Switch to raw-cosine `find_similar_procedures`. |
| **HD-2** | sleep-handler UPDATES-supersession had the same RRF-vs-threshold bug → could supersede an unrelated fact. Switch to `find_similar_facts`. |
| **CL-1** | Subtask results were appended only to the flat `system_prompt` the F036 split path ignores, while rows were marked *delivered* → parent agent permanently never saw subtask outcomes. Route into `sections_by_tier["dynamic"]`. |
| **CL-2** | F065 hub-shift notice had the same flat-string-ignored bug → route into the dynamic tier too. |
| **BR-1** | Graph fact-neighbor query lacked an `active` filter (only procedures got the F080 fix) → superseded facts resurfaced in recall. Add `active=true` to the fact query (episodes/chunks intentionally excluded — episode `active=false` is the normal *closed* state). |
| **HT-1** | Episode hybrid search filtered `active=true AND outcome!='abandoned'`, excluding both closed and ongoing episodes → episode semantic recall was structurally dead. Drop the active filter; use `IS DISTINCT FROM 'abandoned'`. |
[...]

PR #497, opened 2026-06-10
fix(audit): Tier 1 live-correctness fixes (HD-4/5, CL-4/5, HB-8/9)

tfatykhov:
## Summary

Tier 1 of the post-audit fix program (follow-up to #496): six **live-correctness** fixes, each contained and test-backed.

| ID | Fix |
|----|-----|
| **HD-4** | `decision_reviewer` fed Brain calibration false labels three ways. Dropped ErrorSignal's error-keyword branch (matched the *plan text*, not the outcome — a decision *about* a bug was auto-failed), and unwired EpisodeSignal (mapped a hardcoded episode `"success"` onto every decision in the session) and FileExistsSignal (a pre-existing referenced file auto-succeeded the decision). Only outcome-grounded signals remain: a low-confidence prior + GitHub PR state. Signal classes kept (unit-tested). |
| **HD-5** | Schedule timezone was parsed then **discarded** — "daily at 9am EST" fired at 09:00 UTC. Convert the wall-clock hour in the named tz to UTC at cron generation via `zoneinfo` (abbrev + IANA), fail-safe to UTC on unknown tokens. |
| **CL-4** | `claim_verifier` matched tool **name only**, so a blocked/errored tool call satisfied an action claim ("I deployed" passed if any `bash` ran, even a censored one). Now only `status=='success'` ledger actions ground a claim. |
| **CL-5** | `_check_confidence_drift` crashed on an explicit `confidence=None` (`None < 0.4` TypeError). Coerce None → 0.5 while preserving a legitimate 0.0. |
| **HB-8** | Heartbeat circuit breaker **never auto-recovered** (a 9-min IMAP blip permanently disabled the email check until a manual REST reset). Added a cooldown-based half-open trial; success closes it, a failed trial re-arms the cooldown. |
| **HB-9** | Every escalation forced `urgency="high"`, collapsing the documented low→normal→high ladder so aged low findings paged as urgent. Bump one ladder step instead (timing is already gated per-urgency by `_should_escalate`). |

## Tests
+13 regression tests across the six fixes, all green. Pre-existing env-leak failures (`github_token` / dag-stall-timeout reading the local `.env`) are unaffected and pass in CI.

## Notes / documented follow-ups
- HD-4: the surviving low-confidence prior is weakly circular (derives an outcome from the confidence F058 is calibrating); flagged in-code for a deeper follow-up. The episode-outcome hardcode (`layer.py:1773`) and the rubric loop are Tier 3.
- HD-5: a fixed-hour cron baked from the current offset drifts one hour across a DST transition until re-created; true DST-stable firing needs a tz-aware scheduler (deferred).
- CL-4: arg-level matching (`git push` vs `ls`) is a larger follow-up.

### DIFF
--- a/nous/cognitive/claim_verifier.py
+++ b/nous/cognitive/claim_verifier.py
@@ -86,8 +86,15 @@ class ClaimVerifier:
             return VerificationResult(verified=True)
 
         # Build a set of tool names seen in the last 10 ledger entries.
+        # Audit CL-4 (2026-06-09): only count SUCCESSFUL actions. A blocked /
+        # errored / timed-out tool call must not satisfy an action claim — e.g.
+        # a censored or failed `bash` should not let "I pushed the code" verify.
+        # (Arg-level matching — distinguishing `git push` from `ls` — is a
+        # deeper follow-up; this closes the status hole the audit flagged.)
         recent_ledger_tools: set[str] = {
-            action.tool_name for action in ledger.actions[-10:]
+            action.tool_name
+            for action in ledger.actions[-10:]
+            if action.status == "success"
         }
         turn_tool_set = set(tool_calls_this_turn)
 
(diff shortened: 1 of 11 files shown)


tfatykhov:
@codex review

PR #498, opened 2026-06-10
fix(audit): Tier 2 DAG reliability + multi-agent review follow-ups

tfatykhov:
## Summary

Tier 2 of the post-audit fix program (follows #496, #497): DAG reliability fixes **plus multi-agent review follow-ups** (a 3-agent review stood in for Codex, which is rate-limited).

### Core DAG fixes
| ID | Fix |
|----|-----|
| **DG-6** | Work-queue DAGs were `create()`d but never `start_dag()`d — idled `pending` until the recovery sweep (~300s+). Now started after `mark_dispatched` in both the claim and reconciler paths. |
| **DG-4** | A momentarily-full subtask queue raised a bare `ValueError` the orchestrator treated as a permanent node failure (cascading to dependents). Now raises `SubtaskQueueFull` and **defers** the node (`ready`→`pending`). |
| **DG-2** | `cancel()` only affected `pending` rows → cancelling a DAG left RUNNING subtasks executing. Now covers running rows; `complete()`/`fail()` get a terminal-state guard so a stale worker can't resurrect a cancelled row. |

### Multi-agent review follow-ups (same PR)
- **CL-4 turn-side** (HIGH, incomplete-fix): the claim verifier's *turn* leg still grounded claims on FAILED tools — now filtered to `tr.error is None`.
- **DG-4 check-node path** (HIGH, incomplete-fix): `_launch_check_node` had the same permanent-fail-on-transient bug — added `DynamicCheckLimitReached` + defer.
- **DG-4 silent-hang backstop** (P2): a deferred node could bounce `ready↔pending` forever (stall detection only covers `running`). Added a per-node deferral cap that fails with a clear `saturated` error.
- **DG-6 mask-failure** (P2): `_safe_start_dag` returns a bool; the caller now emits a `normal`-urgency "created but failed to start" Finding instead of a green "Dispatched" on start failure; the benign already-started race logs at DEBUG.
- **HB-8 re-sync regression** (P2, found in already-merged Tier 1): a `DynamicCheckLoader` re-sync dropped `_breaker_opened_at`, leaving an open breaker permanently not-due. `is_due` now derives the open time from `last_run`.
- **HD-5 weekly path** (P3, incomplete-fix): `every monday at 10am EST` silently fired in UTC — added a tz group to `_WEEKLY_RE` + routed through `_local_hour_to_utc`.

## Tests
+20 regression tests total (DG core + follow-ups), all green. Mock-based DAG/heartbeat/time-parser tests pass locally; DB-backed subtask/work-queue tests run in CI.

## Documented residual follow-ups
- DG-2: the reclaim-then-late-finish race (a `reclaim_stale`→`pending` row overwritten by a stale worker) needs worker-identity matching — deeper HT-4.
- HD-4 root cause: `cognitive/layer.py:1773` still hardcodes `outcome="success"` on every episode — Tier 3.
- time_parser unknown-tz still degrades to UTC with a WARNING; expand the abbreviation table or raise at the input boundary later.

### DIFF
--- a/nous/api/runner.py
+++ b/nous/api/runner.py
@@ -2282,7 +2282,11 @@ Rules:
         can't unsend yielded text. Both paths use inject-correction-for-next-turn.
         Accepted limitation from plan review.
         """
-        turn_tool_names = [tr.tool_name for tr in tool_results]
+        # Audit CL-4 (2026-06-09, review follow-up): only SUCCESSFUL tool calls
+        # this turn can ground an action claim. A blocked/errored bash or
+        # web_fetch must not let "I pushed the code" verify — mirrors the
+        # ledger-side status filter in ClaimVerifier.verify.
+        turn_tool_names = [tr.tool_name for tr in tool_results if tr.error is None]
 
         # Claim verification
         if self._claim_verifier:
--- a/nous/dag/orchestrator.py
+++ b/nous/dag/orchestrator.py
@@ -14,6 +14,7 @@ from uuid import UUID
 from nous.config import Settings
 from nous.dag._workspace import assert_inside_root, compute_workspace_path
 from nous.dag.store import DAGStore
+from nous.heart.subtasks import SubtaskQueueFull
 from nous.storage.models import DAGNode, ExecutionDAG
 
 if TYPE_CHECKING:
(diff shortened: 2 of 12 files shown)


tfatykhov:
@codex review

PR #500, opened 2026-06-10
fix: procedure progressive disclosure — complete catalog index + stub-over-cap bodies

tfatykhov:
## Summary

Implements two progressive-disclosure fixes to procedure context rendering, aligning Nous with how Anthropic Agent Skills, the MCP meta-tool pattern, and langgraph-bigtool all handle progressive disclosure.

**Anti-pattern removed:** A truncated procedure body (`…(truncated — call get_procedure(name))`) is advisory, not a forcing function. Empirically the model proceeds on the visible partial steps and never calls `get_procedure`. Similarly, dropping procedure _names_ from the catalog means the model can't call `get_procedure` for them at all — they become invisible.

### Change 1 — Catalog two-pass degradation (`nous/cognitive/context.py`, `nous/config.py`)

**Before:** The size backstop dropped whole rows when the catalog overflowed `proc_catalog_max_chars`, causing procedure names to silently disappear from the catalog and become unfetchable.

**After:** Two-pass build:
1. Pass 1: build full `name (domain) — description` rows for all procedures
2. Pass 2 (if over budget): strip descriptions from lowest-priority rows first — names survive; rows degrade to name-only
3. Pass 3 (last resort): drop whole rows only when even name-only rows overflow

Net guarantee: every procedure name is present whenever it physically fits as a name-only row. Descriptions degrade gracefully first.

Also bumps `proc_catalog_max_chars` default from 4000 → 8000 (`nous/config.py`) so description-strip degradation rarely triggers in practice.

### Change 2 — Stub-over-cap for Recommended Procedures (`nous/cognitive/context.py`)

**Before:** When a procedure body exceeded `proc_recommended_body_max_chars`, it was tail-sliced to the cap and an advisory note appended.

**After:** When a body exceeds the cap, a **non-actionable stub** is emitted instead — heading + one-line description + mandatory-fetch pointer. Zero body/step text. The stub makes `get_procedure` structurally required before acting (the fetch becomes an unavoidable step, not an optional hint).

```
### deploy-to-prod (ops)

Deploy production services safely.

[Full skill body exceeds preview budget — call get_procedure('deploy-to-prod') to load the steps before acting.]
```

Bodies that fit within the cap render exactly as before (regression-guarded).

## Test plan

- [x] `test_body_over_cap_emits_stub_not_slice` — stub present, pointer present, no body/step text, stub is short
- [x] `test_body_under_cap_renders_fully` — under-cap body renders fully unchanged (regression guard)
- [x] `test_catalog_two_pass_names_survive_description_strip` — 10 procedures with large descriptions in tight budget: all 10 names present, descriptions stripped
- [x] `test_catalog_drops_rows_only_when_names_do_not_fit` — 20 procedures in extremely tight budget: rows dropped only after descriptions stripped, note present, shown rows are name-only

All 49 tests in `test_context_dual_track.py` + `test_f079_pull_delivery.py` pass.

## References

Architecture decisions 7f7a3921 and dc65b892.

### DIFF
--- a/nous/cognitive/context.py
+++ b/nous/cognitive/context.py
@@ -326,8 +326,12 @@ class ContextEngine:
                 # Reserve room for the trailing "…N+ more" note so the size backstop never
                 # has to drop IT (the note must survive truncation).
                 _NOTE_RESERVE = 140
-                row_lines: list[str] = []
-                used = len(header) + 1
+
+                # Two-pass progressive degradation — names are never sacrificed before
+                # descriptions (progressive disclosure: complete index is always-on).
+                # Pass 1: build full "name (domain): desc" rows and name-only fallbacks.
+                full_rows: list[str] = []
+                name_rows: list[str] = []
                 for p in deduped:
                     # name: keep exact (it's the get_procedure key) but kill line breaks;
                     # domain/desc: full whitespace-collapse (not lookup keys).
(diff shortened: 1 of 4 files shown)


tfatykhov:
## PR #500 Review

**Status:** ✅ APPROVED

### Summary
Two complementary progressive-disclosure fixes: (1) catalog never drops names before stripping descriptions, (2) over-cap procedure bodies emit a mandatory-fetch stub instead of a tail-sliced partial. Both changes directly address observed failure modes — silent name disappearance and the model rationalising partial steps as sufficient.

### P1 Issues (Must Fix)
None.

### P2 Issues (Should Fix)
- **`context.py:313`** — `getattr(self._settings, "proc_catalog_max_chars", 4000)` hard-codes the old default as the fallback. Config now defaults to `8000`, so this never fires in practice (Pydantic always sets the attribute), but it is cosmetically stale. Not a bug — leaving as a minor nit.

### P3 Observations
- `_catalog_size` is defined as a nested function inside an already-nested block — functional, just slightly deep. Fine at 2 lines.
- `test_catalog_drops_rows_only_when_names_do_not_fit` uses `proc_catalog_max_chars=500`; a one-liner comment explaining that 500 fits ~3-4 `proc-longname-XX` name-only rows would help future readers, but the docstring covers it.

### Highlights
- Three-pass degradation algorithm is clean and correct: descriptions degrade → names survive → whole rows drop only as last resort.
- `_catalog_size(rows)` helper keeps the budget check DRY across all three passes.
- Stub-over-cap rationale is well-documented inline — the comment explaining *why* tail-slicing is the anti-pattern is exactly the context future readers need.
- `test_body_over_cap_emits_stub_not_slice` validates all three stubs properties (pointer present, no body text, stub length < 400) — tight test.
- Regression guards present: `test_body_under_cap_renders_fully` confirms non-stub path is unchanged.
- CI: Tests ✅ Lint ✅

### Verdict
Logic is sound, tests are comprehensive, CI is green. Merging.

PR #501, opened 2026-06-10
fix: bound oversized description in over-cap procedure stub (Codex P1 #500)

tfatykhov:
## What
Addresses the **Codex P1** left on #500.

The `stub-over-cap` branch in `_format_procedure_bodies` copied the procedure `description` in full. A description longer than `per_item_cap` therefore **defeated the cap the stub exists to enforce** — and because the budget loop always admits the first block (`shown_blocks` starts empty), one oversized description could swallow the whole procedure/context budget.

## Fix
- Bound the description within `per_item_cap`, reserving room for the heading and the mandatory `get_procedure(...)` fetch pointer.
- The fetch pointer is **never** sacrificed (if the pointer alone fills the cap, the description is dropped entirely).
- Reserves one char for the ellipsis so the stub is provably `<= per_item_cap`.

## Tests
- Adds `test_stub_bounds_oversized_description` (cap=200, 5000-char description → stub `<= cap`, pointer present, no body leak).
- `tests/test_context_dual_track.py` + `tests/test_f079_pull_delivery.py`: **50 passed**.
- (DB-integration suites skipped in CI sandbox — missing `aiosqlite`, unrelated.)

Refs: Codex review on #500.

### DIFF
--- a/nous/cognitive/context.py
+++ b/nous/cognitive/context.py
@@ -1538,32 +1538,45 @@ class ContextEngine:
                 # A non-actionable stub forces the model to call get_procedure before acting.
                 # (A tail-sliced partial body is the anti-pattern — the model rationalizes it
                 # as sufficient and never fetches the rest.)
+                #
+                # The fetch POINTER is the irreducible floor of the stub: it carries the
+                # procedure name and the load instruction, so the stub is useless without it
+                # and the name cannot be truncated without breaking get_procedure fetchability.
+                # Therefore the cap guarantee is: len(block) <= max(per_item_cap, len(pointer)).
+                # (Codex P2 on #501: with a tiny cap and an up-to-500-char name, heading+pointer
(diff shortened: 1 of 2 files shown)


tfatykhov:
Codex P2 addressed in 0243a32. The fetch pointer is now treated as the stub's irreducible floor (it carries the procedure name and the get_procedure instruction, which cannot be truncated without breaking fetchability). The stub is built up from the pointer: heading and description are dropped as needed, and when `per_item_cap` < pointer length the stub falls back to the pointer alone. Guarantee is now `len(block) <= max(per_item_cap, len(pointer))`, covering tiny caps and up-to-500-char names. Added two regression tests (`test_stub_bounded_when_cap_below_heading_and_pointer`, `test_stub_pointer_floor_with_long_name`). Full suite green (29 passed).

tfatykhov:
@codex

PR #502, opened 2026-06-10
fix(F078.1.4): decode HTML entities before secret scan in html_body path

### LINKED ISSUE #484: fix(F078.1.4): decode HTML entities before secret scan in html_body path
## Context

PR #483 (F078.1.3) extends `_scan_secrets()` to cover `html_body` by appending it to the scanned string. However, the scan runs against **raw HTML source** — HTML-entity-encoded secrets pass through undetected.

## The Gap

```python
# Current (line ~253 in email_tools.py after #483)
if _scan_secrets(f"{subject}\n{body} \n{html_body or ''} "):
```

A secret like `sk-abcdEFGH1234567890zz` split across tags or encoded as `sk-&#97;bcdEFGH...` won't match `_SECRET_PATTERNS`, but mail clients render it correctly.

## Fix

Before passing `html_body` to `_scan_secrets`, normalize:
1. **HTML entity decoding:** `html.unescape(html_body)`
2. **Strip tags:** extract text content (e.g. via `re.sub(r'<[^>]+>', '', ...)`)

```python
import html, re

def _html_to_text(s: str) -> str:
    """Strip HTML tags and decode entities for secret scanning."""    return html.unescape(re.sub(r'<[^>]+>', ' ', s))

# In _scan_secrets call:
html_text = _html_to_text(html_body) if html_body else ''
if _scan_secrets(f"{subject}\n{body}\n{html_text}"):
```

## Priority

**P2 / non-blocking.** The allowlist (trusted recipients only) is the primary guard. This is defense-in-depth. Must be addressed before `premarket-daily` migrates onto the html_body path in production — scheduled jobs could embed auto-generated content including tokens.

## Acceptance Criteria
- [ ] `_html_to_text()` helper added
- [ ] `_scan_secrets` call in `send_email` uses decoded text for html_body
- [ ] New test: `test_html_body_entity_encoded_secret_rejected` — entity-encoded API key in html_body triggers rejection
- [ ] All 23 existing tests still pass

Identified by Codex in review of #483.

tfatykhov:
Closes #484 (Codex finding from #483 review).

## Problem
`_scan_secrets` ran against raw HTML source only. An entity-encoded secret (`sk-&#97;bcdEFGH...`) passes the regexes but renders as a real key in any mail client.

## Fix
- New `_html_to_text()` helper: strip tags (`<[^>]+>` → space), **then** `html.unescape` — decode-after-strip so a decoded `&lt;` can't fabricate a tag that hides content.
- The scan now sees **both views**: raw HTML source **and** decoded text. The issue's snippet replaced raw with decoded, which would have regressed secrets inside tag attributes (e.g. `href="...?key=sk-..."`) that tag-stripping removes — scanning both is a strict superset.

## Tests
- `test_html_body_entity_encoded_secret_rejected` — **failed before the fix** (email sent), passes after ✅ (issue acceptance)
- `test_html_body_attribute_secret_still_rejected` — regression guard locking the raw-scan view
- 25/25 `tests/test_email_tools.py` pass

## Known residual (out of scope)
Tag-split secrets (`sk-ab<b>cd</b>...`) still evade — tags become spaces per the issue spec. Joining with `''` instead would catch it but risks false positives from adjacent text runs; can revisit if it ever shows up.

### DIFF
--- a/nous/api/email_tools.py
+++ b/nous/api/email_tools.py
@@ -265,12 +265,17 @@ def create_send_email_tool(settings: Settings):
                     "(operator action)."
                 )
 
-        # 3. Secret scan (secondary guard). html_body is scanned twice: raw
-        # source (catches secrets inside tag attributes, which tag-stripping
-        # would remove) AND entity-decoded text (#484 — catches secrets a
-        # mail client would render from entity-encoded source).
-        html_text = _html_to_text(html_body) if html_body else ""
-        if _scan_secrets(f"{subject}\n{body}\n{html_body or ''}\n{html_text}"):
+        # 3. Secret scan (secondary guard). html_body is scanned in three
+        # views (#484): raw source (secrets in tag attributes), entity-decoded
+        # source (codex P1 — encoded secrets *inside* attributes, which
+        # tag-stripping would remove before decoding), and tag-stripped +
+        # decoded text (secrets a mail client renders from encoded content).
+        html_views = (
+            f"\n{html.unescape(html_body)}\n{_html_to_text(html_body)}"
+            if html_body
+            else ""
+        )
+        if _scan_secrets(f"{subject}\n{body}\n{html_body or ''}{html_views}"):
             return _error(
                 "email appears to contain a secret (API key, password, or token); "
                 "refusing to send."
--- a/nous/api/email_tools.py
+++ b/nous/api/email_tools.py
@@ -16,6 +16,7 @@ return only a generic ``email send failed: <ExceptionType>`` to the model.
 from __future__ import annotations
 
 import asyncio
+import html
 import logging
 import os
 import re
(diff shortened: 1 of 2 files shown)


tfatykhov:
@codex review

tfatykhov:
@codex review — pushed fix for the P1: added the entity-decoded (pre-strip) view to the scan; html_body is now scanned raw + unescaped + stripped-and-unescaped. New regression test test_html_body_entity_encoded_attribute_secret_rejected covers the encoded-attribute case. 26/26 tests pass.

tfatykhov:
@codex review — addressed the P2: replaced the tag-strip regex with a linear single-pass char walk; added test_html_body_malformed_html_fast_and_scanned (80K '<' flood + secret after unclosed tag → still rejected, <2s). 27/27 tests pass.

PR #503, opened 2026-06-10
fix(#369): heartbeat SelfInitiatedCheck — question-form filter + staleness age cap

### LINKED ISSUE #369: Heartbeat SelfInitiatedCheck emits stale & question-form false positives
## Problem

The proactivity heartbeat check (`SelfInitiatedCheck` in `nous/heartbeat/checks.py`) generates false-positive findings that recur across sweeps even when the underlying work is complete. Recent example — a single sweep flagged 5 items, all of which were stale or non-actionable:

- **HL-01 PR #38** — already merged 2026-04-25
- **F032 Subtask Completion Validation** — already resolved via censors `0b0ce2c4` + `2c4da9f1`
- **"back to nous, get latest..."** — conversational opener, not a commitment
- **"Hey how was your conversation with Emerson?"** — question fragment
- **"ok what outstanding items we do have?"** — question fragment

## Root Cause

Two heuristic paths emit these findings, both rule-based (no LLM):

### 1. `_embedding_search` (line ~210-262)
Surfaces facts via embedding similarity to actionable prototypes. Falls back to substring heuristics (`_looks_like_pending` / `_is_observation`) when `fact.actionable` is `None`. Resolved facts that were never explicitly marked `actionable=False` keep matching forever.

### 2. `_promise_scan` (line ~264-310)
Hardcoded query list:
```python
promise_queries = [
    "I'll look into",
    "let me research",
    "I'll draft",
    "unfinished commitment",
    "ongoing task not completed",
]
```
Searches episodes via embedding similarity, then flags any result where `outcome == "ongoing"` **OR** `age_hours > 48`. The age-based fallback means **any episode older than 2 days that semantically resembles a promise gets flagged indefinitely** — including questions Tim asked the assistant.

## Proposed Fix

Three layers, in increasing scope:

### A. Filter question-form episode summaries (small, high-value)
Add a regex guard in `_promise_scan` before flagging:
```python
QUESTION_PREFIX = re.compile(r"^\s*(what|how|where|when|why|who|is|are|do|does|did|can|could|should|would|hey|ok|okay)\b", re.I)
if QUESTION_PREFIX.match(summary_text):
    continue
```
Eliminates conversational openers and questions being treated as commitments.

### B. Cap age-based staleness window
Currently any episode >48h old + ongoing-ish embedding match is flagged forever. Add an upper bound:
```python
if is_stale and age_hours > MAX_STALE_AGE_HOURS:  # e.g. 14 days
    continue  # too old to be actionable, drop
```

### C. Auto-set `actionable=False` on resolution
When a fact is superseded or its referenced PR/issue is closed, write `actionable=False` so embedding search stops re-flagging it. Could be a sleep-cycle pass or hooked into existing supersession logic.

## Acceptance

- The 5 findings listed above stop appearing in subsequent heartbeat sweeps without manual intervention
- Add unit tests covering: question-form filter, age cap, resolved-fact suppression
- No regression on genuine ongoing commitments (e.g. an episode with `outcome=ongoing` that's <14 days old and not question-form)

## Files

- `nous/heartbeat/checks.py` (SelfInitiatedCheck, ~line 200-310)
- Tests: add to `tests/heartbeat/test_checks.py` (or create)

## Related

- Censors `0b0ce2c4` and `2c4da9f1` already suppress F032 false-positives at runtime; this issue addresses the upstream extractor instead.

tfatykhov:
Closes #369 (the two filters remaining after the F047 `actionable` mitigation shipped).

## Changes
- **Question-form guard** (`_QUESTION_PREFIX`, issue's verbatim prefix list): episode summaries opening as a question/conversational filler are skipped before flagging — they're not commitments.
- **Staleness upper bound**: the age-based heuristic is now `48 < age_hours <= max_stale_age_hours` instead of `> 48`-forever. Shipped as a `TunableParam` (`max_stale_age_hours`, default 336 h = 14 d, range 72–720, step 24) so the F034.3 self-tuning system owns it rather than a config global.
- **Deliberate scope choice:** `outcome=ongoing` is an explicit recorded state, not a heuristic — the age cap does **not** suppress it (test locks this in). The issue's snippet gated only the `is_stale` path; this matches it.

## Tests (4 new, TDD — the two filter tests failed pre-fix)
- `test_question_form_summary_not_flagged`
- `test_stale_episode_beyond_age_cap_not_flagged`
- `test_stale_episode_within_age_cap_flagged` (regression guard)
- `test_old_ongoing_episode_still_flagged` (regression guard)

124/124 heartbeat tests pass.

### DIFF
--- a/nous/heartbeat/checks.py
+++ b/nous/heartbeat/checks.py
@@ -198,7 +198,13 @@ class SelfInitiatedCheck(BaseCheck):
             "max_pending_items": TunableParam("max_pending_items", 5, 2, 15, 1),
             # #369: upper bound on the age-based promise heuristic (hours).
             # Episodes older than this are too old to be actionable.
-            "max_stale_age_hours": TunableParam("max_stale_age_hours", 336, 72, 720, 24),
+            # pinned=True (codex P2): the generic tuner's relax=+step would
+            # WIDEN this window on noisy feedback — inverted for an upper
+            # bound. Operator-adjustable via the params API; excluded from
+            # auto-tuning until TunableParam grows direction metadata.
+            "max_stale_age_hours": TunableParam(
+                "max_stale_age_hours", 336, 72, 720, 24, pinned=True
+            ),
         }
 
     async def _ensure_prototypes(self) -> list[list[float]]:
--- a/nous/heartbeat/checks.py
+++ b/nous/heartbeat/checks.py
@@ -11,6 +11,7 @@ from __future__ import annotations
 import asyncio
 import imaplib
 import logging
+import re
 from collections.abc import Callable
 from datetime import UTC, datetime, timedelta
 from typing import TYPE_CHECKING
@@ -145,6 +146,13 @@ PENDING_PROTOTYPES = [
     "Action required but not yet taken",
 ]
 
+# #369: episode summaries that open as a question or conversational filler are
+# not commitments — _promise_scan skips them before flagging.
+_QUESTION_PREFIX = re.compile(
+    r"^\s*(what|how|where|when|why|who|is|are|do|does|did|can|could|should|would|hey|ok|okay)\b",
+    re.IGNORECASE,
+)
+
 
 class SelfInitiatedCheck(BaseCheck):
     """Check for pending actions and due schedules (permanent).
@@ -173,6 +181,9 @@ class SelfInitiatedCheck(BaseCheck):
             "similarity_threshold": TunableParam("similarity_threshold", 0.75, 0.6, 0.9, 0.02),
             "lookback_days": TunableParam("lookback_days", 14, 3, 30, 1),
             "max_pending_items": TunableParam("max_pending_items", 5, 2, 15, 1),
+            # #369: upper bound on the age-based promise heuristic (hours).
+            # Episodes older than this are too old to be actionable.
+            "max_stale_age_hours": TunableParam("max_stale_age_hours", 336, 72, 720, 24),
         }
 
     async def _ensure_prototypes(self) -> list[list[float]]:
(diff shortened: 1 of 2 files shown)


tfatykhov:
@codex review

tfatykhov:
@codex review — addressed the P2: replaced the bare prefix match with _is_question_form() (strip hey/ok/okay filler → suppress only on interrogative wh-start or trailing '?'). Modal starts like 'Should update the deployment docs' no longer suppress; your three counter-examples are now a regression test. 125/125 heartbeat tests pass.

tfatykhov:
@codex review — pinned max_stale_age_hours (pinned=True) so the generic tuner can't widen the upper bound on noisy feedback; comment documents the inversion. Note the same relax/tighten inversion already exists for lookback_days and max_pending_items (relax=+step increases findings for both) — that's pre-existing F034.3 behavior, out of scope here; happy to file it as a follow-up issue. Your other two round-2 comments were on the superseded commit (272d1ae) — both already addressed in f708cb1 (_is_question_form strips filler then requires interrogative structure).

tfatykhov:
Re the two round-3 P2s — declining both, with rationale:

1. **Declarative wh-clauses** ('What I need to do next is update the docs'): real but rare phrasing for an episode summary, and the failure cost is one missed *heuristic* reminder (the embedding leg can still surface the underlying fact). Loosening the wh-rule re-admits the observed production FP class ('what outstanding items we do have' — these don't always carry a '?'). This filter is an FP-reducer targeting the 5 observed classes in #369, not an English question parser.

2. **Auxiliary-led questions without '?'** ('Can you look into the migration'): that's an *unresolved user request* — flagging it as a pending commitment is the check working as intended. The #369 false positives were social/meta questions ('Hey how was your conversation with Emerson?'), which the filler+wh+'?' rules catch.

Both trade-offs are now documented in the _is_question_form docstring intent. If either class shows up as a real production FP/FN, the next step is a Haiku tiebreak (like F377's), not more regex.

PR #1, opened 2026-02-22
feat: implement 001-postgres-scaffold foundation

tfatykhov:
## Summary
- **Docker**: pgvector/pgvector:pg17 container with healthcheck, configurable port/password
- **SQL Schema**: 18 tables across 3 schemas (brain, heart, nous_system) with HNSW vector indexes, CHECK constraints, generated tsvector FTS columns, updated_at triggers, and seed data
- **Python Package**: async SQLAlchemy 2.0 ORM models for all 18 tables, pydantic-settings config, async Database class with connection pool and context manager
- **Tests**: 18 tests with SAVEPOINT isolation covering infrastructure, seed data, ORM CRUD, cross-schema FKs, constraint enforcement, and trigger behavior

## Design Decisions (from 3-agent review team)
- `nous_system` schema instead of `system` (tooling ergonomics)
- HNSW indexes everywhere (not ivfflat — works on empty tables)
- Generated tsvector columns for FTS (cleaner ORM queries)
- SAVEPOINT test fixtures (tests can commit freely)
- CHECK constraints on all enum-like VARCHAR columns
- `updated_at` triggers on 6 tables
- Guardrails: UNIQUE(agent_id, name) for multi-agent support

Full implementation plan: `docs/implementation/001-postgres-scaffold-plan.md`

## Test plan
- [ ] `docker compose up -d postgres` starts successfully
- [ ] `pip install -e ".[dev]"` installs without errors
- [ ] `pytest tests/ -v` — all 18 tests pass
- [ ] Verify 18 tables across 3 schemas in pgAdmin/psql
- [ ] Verify seed data: 1 agent, 6 frames, 4 guardrails

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #505, opened 2026-06-10
fix(#179): bulk-result decay profile + anti-replay stubs (replay-loop P1)

### LINKED ISSUE #179: Conversation replay loop after high-volume tool operations (F025 sweep)

tfatykhov:
Closes #179 (P1 — Nous replays completed sweeps until `/new`).

## Root cause
Tool pruning operates per tool_result *message*. A 350-call sweep inside one `run_python` is ONE block on the 'standard' (3, 8, 12) profile — un-prunable for ~12 tool calls — and its stub text never says the work finished, so the model re-derives 'I should run the sweep' from its own context.

## Fix (three layers)
1. **`bulk` decay profile (1, 2, 4)** — assigned dynamically by size (`NOUS_TOOL_BULK_RESULT_CHARS`, default 50 000; 0 disables), not by tool.
2. **Anti-replay stubs** — bulk degrade/clear text states the operation **ran and COMPLETED** and says *do NOT re-run it*. Bulkness survives soft-trim (original size parsed from the trim marker — stateless, nothing added to API-bound blocks) and survives degrade (hint stamped into the stub).
3. **Compaction prompt rule** — bulk/repetitive sequences must compress to a single COMPLETED line, never read as a pending plan (issue fix #5).

## Deliberately not built
- Repeated-signature dispatch guard (issue fix #1): ActionGate (#285) already enforces ALL-match turn-window dedup — building a second arbiter would split that logic.
- Assistant-message pruning (issue fix #4): the issue itself flags it 'needs careful design'; deferred.

## Tests
5 new in `TestBulkResultPruning` (bulk clear at age 4 with anti-replay stub, sub-threshold keeps standard ages, detection survives soft-trim, degrade carries hint into clear tier, 0-disables). 34/34 `test_compaction.py` pass. The 13 failures in `test_model_aware_compaction.py`/`test_compaction_phase2.py` are byte-identical on clean main (pre-existing Python 3.14 event-loop issue, verified via stash).

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -367,6 +367,7 @@ DB connection vars are **unprefixed** (shared with docker-compose). All others u
 | `NOUS_TOOL_SOFT_TRIM_TAIL` | `1500` | Chars to keep from end when soft-trimming |
 | `NOUS_TOOL_METADATA_DEGRADE_AFTER` | `8` | Tool result age (in results) before metadata degradation |
 | `NOUS_TOOL_HARD_CLEAR_AFTER` | `12` | Tool result age before hard-clear replacement |
+| `NOUS_TOOL_BULK_RESULT_CHARS` | `50000` | #179: results at/above this original size (survives soft-trim + SmartCompress via markers) from operation-shaped tools (`BULK_ESCALATION_TOOLS` = bash, run_python) escalate to the `bulk` (1, 2, 4) decay profile with anti-replay stubs. `0` disables. |
 | `NOUS_KEEP_LAST_TOOL_RESULTS` | `2` | Number of most recent tool results always protected |
 | `NOUS_COMPACTION_ENABLED` | `true` | Enable LLM-powered history compaction |
 | `NOUS_COMPACTION_THRESHOLD` | auto | Token count triggering compaction (auto-scales per model context window) |
--- a/nous/api/builtin_tools.py
+++ b/nous/api/builtin_tools.py
@@ -107,8 +107,11 @@ async def bash_tool(
             parts.append(stdout_text)
         if stderr_text:
             parts.append(f"STDERR:\n{stderr_text}")
-        if proc.returncode != 0:
-            parts.append(f"Exit code: {proc.returncode}")
+        # Always appended (#179 codex round 13): the trailing line is the
+        # AUTHORITATIVE wrapper status, so downstream consumers (compaction
+        # bulk-failure detection) can disambiguate a quoted "Exit code: N"
+        # inside the command's own output from the real return code.
+        parts.append(f"Exit code: {proc.returncode}")
 
         output = "\n".join(parts) if parts else "(no output)"
         return _mcp_response(output)
--- a/nous/api/compaction.py
+++ b/nous/api/compaction.py
@@ -576,9 +576,13 @@ class ConversationCompactor:
     # not semantic success — a clean exit can hide internal partial failures.
     # "(tool reported success)" (codex round 7) keeps the reported outcome
     # visible at the degrade stage so checkpoint summaries can mirror it.
+    # "to redo completed work" (codex final P1): bash/run_python also serve
+    # retrieval (cat/grep of large files) — the anti-replay instruction must
+    # forbid REDOING the operation, not legitimate re-fetching of output
+    # that is needed again.
     _BULK_HINT = (
         "bulk operation already ran earlier (tool reported success) — "
-        "do NOT re-run it"
+        "do NOT re-run it to redo completed work"
     )
     # codex P1: a failed sweep must not be stamped as completed — that would
     # suppress a legitimate retry and corrupt later summaries.
@@ -702,8 +706,9 @@ class ConversationCompactor:
         return (
             f"[Bulk tool output cleared — this {name} operation already ran "
             f"in an earlier turn and its output was processed then. "
-            f"{self._BULK_HINT} — work from its saved results or the earlier "
-            f"conversation instead.]"
+            f"{self._BULK_HINT}; re-run it only if you genuinely need its "
+            f"output again — otherwise work from its saved results or the "
+            f"earlier conversation.]"
         )
 
     def _extract_facts_before_clear(self, tool_name: str, content: str) -> list[str]:
--- a/nous/api/compaction.py
+++ b/nous/api/compaction.py
@@ -610,7 +610,9 @@ class ConversationCompactor:
     # the raw tail would miss it.
     _EXIT_CODE_RE = re.compile(r"^Exit code: (-?\d+)\s*$", re.MULTILINE)
 
-    def _item_reports_failure(self, item: dict[str, Any], text: Any) -> bool:
+    def _item_reports_failure(
+        self, item: dict[str, Any], text: Any, tool_name: str | None
+    ) -> bool:
         """True if this result reports failure — via is_error, a trailing
         bash exit-code marker, a timeout message, or a previously stamped
         error hint. Must be called BEFORE the item is mutated."""
@@ -620,6 +622,12 @@ class ConversationCompactor:
             return False
         if self._BULK_ERROR_HINT in text:
             return True
+        # The exit-code and timeout markers are emitted by bash_tool only
+        # (builtin_tools.py) — applying them tool-agnostically would misread
+        # e.g. fetched page content mentioning "Exit code: 1" (codex P2,
+        # round 10).
+        if tool_name != "bash":
+            return False
         if "Command timed out after" in text[:200]:
             return True
         # Widened tail window: the exit-code line may be followed by the
@@ -758,7 +766,7 @@ class ConversationCompactor:
 
                 # Failure status must be read BEFORE any mutation below
                 # destroys the text it lives in (codex round 8).
-                failed = is_bulk and self._item_reports_failure(item, text)
+                failed = is_bulk and self._item_reports_failure(item, text, tool_name)
 
                 # Tier 4: Hard-clear (age >= clear_age)
                 if age >= clear_age:
--- a/nous/api/compaction.py
+++ b/nous/api/compaction.py
@@ -593,10 +593,15 @@ class ConversationCompactor:
                 f"an earlier turn and its error output was cleared. "
                 f"{self._BULK_ERROR_HINT}.]"
             )
+        # codex round 4: is_error=False only means the tool exited cleanly —
+        # a sweep can swallow internal failures — so state facts (ran, tool
+        # reported success, output was processed in earlier turns), not a
+        # semantic COMPLETED claim.
         return (
-            f"[Bulk tool output cleared — this {name} operation ran and "
-            f"COMPLETED in an earlier turn; its results were already "
-            f"processed. {self._BULK_HINT}.]"
+            f"[Bulk tool output cleared — this {name} operation already ran "
+            f"in an earlier turn (tool reported success) and its output was "
+            f"processed then. {self._BULK_HINT} — work from its saved "
+            f"results or the earlier conversation instead.]"
         )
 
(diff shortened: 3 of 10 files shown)


tfatykhov:
@codex review

tfatykhov:
@codex review — both findings addressed: (P1) base_profile now drives conservative fact extraction independently of the bulk decay ages — new test test_bulk_conservative_tool_still_extracts_facts proves a bulk web_fetch still yields its URLs at clear; (P2) the repetitive-operations rule is now in BOTH prompts with a test asserting it. 36/36 pass.

tfatykhov:
@codex review — all four round-2 P1s addressed: (1) per-item profile/bulk resolution (test: bulk sibling keeps own profile, conservative web_fetch untouched at age 5); (2) error-aware stubs — is_error bulk results get FAILED + 'fix the cause before any retry', never COMPLETED (test); (3+4) conservative facts extracted at the first content-destroying tier with mutation+stub guards so the incremental degrade→clear path captures them exactly once (test walks the two-pass lifecycle). 39/39 pass.

tfatykhov:
@codex review — round-3 finding addressed: the repetitive-operations rule in both prompts is now outcome-aware ('COMPLETED or FAILED, as the tool results show' + 'never describe a failed one as completed'), consistent with the error-aware stubs. Test updated. 39/39 pass.

tfatykhov:
@codex review — round-4 finding addressed: the success stub no longer asserts semantic completion; it states only what is knowable at prune time ('already ran in an earlier turn (tool reported success) and its output was processed then') and points the model at saved results. The FAILED branch is unchanged. 39/39 pass.

tfatykhov:
@codex review — round-5 finding addressed: _BULK_HINT is now outcome-neutral ('bulk operation already ran earlier — do NOT re-run it'); no remaining text asserts semantic completion anywhere in the bulk path (the success stub states only 'tool reported success'). 39/39 pass.

tfatykhov:
@codex review — round-6 finding addressed: both summary prompts now require stating the outcome exactly as the surviving tool results word it, and explicitly forbid upgrading 'tool reported success' into a stronger completion claim. The no-semantic-success principle is now uniform across all four layers it can appear in (clear stub, degrade hint, checkpoint prompt, update prompt). 39/39 pass.

tfatykhov:
@codex review — round-7 findings addressed: (P1) preserve-profile items (read_file) are exempt from bulk escalation — new test test_preserve_profile_exempt_from_bulk locks it; (P2) _BULK_HINT now carries '(tool reported success)' so the reported outcome survives the degrade stage for checkpoint summaries (clear stub delegates outcome text to the hint, no duplication). 40/40 pass.

tfatykhov:
@codex review — round-8 finding addressed: failure detection now also reads the reported outcome from the result content (trailing 'Exit code: N' per builtin_tools.py:110-114, the timeout message, or a previously stamped error hint), evaluated before any mutation. Detection chain survives soft-trim (marker lives in the kept tail) and degrade (error hint carries it to clear). New test: bulk bash with nonzero exit gets the FAILED stub, never 'tool reported success'. 41/41 pass.

tfatykhov:
@codex review — round-9 SmartCompress finding addressed: (1) a [SmartCompressed:] marker now triggers bulk handling directly — F020 only fires on sweep-shaped repetitive content, exactly the #179 class, and the original lives in tool_cache; (2) the exit-code regex is line-anchored multiline over a wider tail window, so 'Exit code: N' followed by the SmartCompress marker is still detected. Both production-order cases have tests. Your other round-9 comment (bash is_error) was reviewed against the superseded commit 581f2fb — ea82241 already derives failure from the reported outcome in content. 43/43 pass.

tfatykhov:
@codex review — round-10 P2 addressed: the exit-code/timeout content heuristics now apply only to bash results (tool_name threaded into _item_reports_failure); is_error and the stamped error hint stay tool-agnostic. New regression test: run_python output quoting 'Exit code: 1' keeps the success stub. 44/44 pass.

tfatykhov:
@codex review — round-11 finding addressed (and generalized): bulk escalation is now restricted to 'standard'-profile tools — the operation-shaped class #179 targets. All pure-retrieval tools (read_file, web_fetch, web_search, list_files, recall_deep) keep their own profiles and never receive do-NOT-re-run stubs, since re-running retrieval is cheap and benign. Conservative fact extraction still fires at the first destructive tier (tests reworked to walk the conservative 5/10/15 lifecycle). 44/44 pass.

tfatykhov:
@codex review — round-12 finding addressed structurally: bulk escalation is now a positive allowlist (BULK_ESCALATION_TOOLS = {bash, run_python} in cognitive/schemas.py) instead of profile-default. No registered or future retrieval tool can escalate by default. New test covers an unlisted tool with an oversized result. 45/45 pass.

tfatykhov:
@codex review — round-13 addressed: (P1) smart_compress() now records 'N chars original' in its marker and the pruner compares THAT against tool_bulk_result_chars — marker presence alone no longer implies bulk (test: 300-char original not escalated); (P2) bash_tool always appends the authoritative exit-code line and the failure check takes the last match, so quoted 'Exit code: 1' in successful output is overridden by the trailing 'Exit code: 0' (tests both sides). The bash-scoping P2 was already shipped in round 10 (997a8f2). 96/96 across compaction/builtin_tools/smart_compress.

tfatykhov:
**Internal 3-agent review completed** (code-reviewer / silent-failure-hunter / pr-test-analyzer), on top of the codex rounds. Findings fixed in c9e4c86:

**P2 (silent-failure-hunter):** SmartCompress could break the round-13 invariants two ways — (a) the always-appended exit line broke DICT_ARRAY json parsing, so bash JSON output got line-scrambled into broken reordered JSON; (b) to_text()'s first-occurrence line-dedup could displace the authoritative trailing 'Exit code: N' out of the pruner's tail window. **One mechanism fixes both:** smart_compress detaches the status line before classification/compression and re-attaches it as the final line.

**P3s:** max-over-markers original-size (trim marker no longer shadows the SmartCompress size), stub-shape guard on hint matching (grep over this repo quoting the hint strings can't classify as bulk/failed), removed the dead-or-misfire-only timeout branch, marker append targets exactly the final marker line (quoted markers can't be grafted), CLAUDE.md env row added.

**Considered and deliberately NOT changed:** run_python content-level failure sniffing — its genuine errors are short text (never bulk-sized), and tool-agnostic Traceback regexes were already rejected in round 10 as FP-prone; a PIN test now documents this limitation.

**Test gaps closed (pr-test-analyzer):** format-lock e2e test running REAL smart_compress output through the pruner (if either side's marker format drifts, this fails — previously both suites used hand-crafted markers and could drift together), FAILED-idempotency across repeated prunes + clear fixed-point, bash-JSON DICT_ARRAY preservation, dedup-displacement guard.

103/103 across compaction/smart_compress/builtin_tools.

@codex review

tfatykhov:
@codex review — good catch, the timeout branch was NOT dead: the message echoes the command, so a timed-out inline script can be bulk-sized. Restored in precise form — bash-scoped + startswith-anchored (the wrapper emits it as the first characters of the result), so quoted mentions inside command output remain inert, which was the original removal concern. Test: bulk-sized timeout result gets the FAILED stub. 104/104 pass.

tfatykhov:
@codex review — both addressed: (P1) run_python results starting with 'Error: ' (the handler's entire-result error shape, is_error=False) now get the FAILED stub — test with a bulk-sized exception message; the content-sniffing PIN remains for the separate self-printed-failure-lines case. (P2) both size-marker regexes are line-anchored MULTILINE — real markers occupy their own line, grep-quoted ones carry a file:line prefix and no longer classify (test). 106/106 pass.

tfatykhov:
@codex review — final P1 addressed via instruction semantics rather than command classification: the hint now forbids re-running 'to redo completed work' and the clear stub explicitly permits re-running when the output is genuinely needed again — covering the cat/grep retrieval case without an FP-prone read-only command parser (the heuristics would misclassify compound commands, and fast decay itself is correct for re-fetchable bash). The 'Preserve authoritative bash status across degradation' comment (15:50) is stale: line-anchored exit-code matching (added later) makes the quoted line inside the degrade stub unmatchable — the scenario no longer reproduces. 106/106 pass.

tfatykhov:
@codex review — final P2 resolved structurally rather than by tuning the heuristic: your last two findings (detect run_python errors / don't misread printed 'Error:' output) were in tension because run_python multiplexed failures and arbitrary output through one channel. run_python's exception/timeout returns now set the MCP is_error field and ToolDispatcher.dispatch honors it (absent = success, so no other handler changes), and the pruner's content-prefix check is gone — failure rides exclusively on is_error for run_python. Tests cover both sides (bulk exception message → FAILED; bulk output starting with 'Error:' → success). 107/107.

PR #507, opened 2026-06-10
fix(heartbeat): tuner direction metadata for volume-type params

tfatykhov:
Surfaced during #369 (codex flagged the inversion would hit the new `max_stale_age_hours` param; investigation showed the same inversion has affected `lookback_days` and `max_pending_items` since F034.3 shipped).

## Problem
`HeartbeatTuner._apply_adjustment` applies `relax = +step` / `tighten = −step` to every param. 'Relax' means *produce fewer findings* — correct for sensitivity thresholds, **inverted** for volume params: raising a lookback window or an item cap produces MORE findings. On >60% negative feedback the tuner widened exactly the params it meant to narrow.

## Fix
- `TunableParam.increases_findings: bool = False` — direction metadata
- tuner flips the step sign when set
- `BaseCheck.set_param` threads the flag through its dataclass reconstruction (would otherwise silently reset after the first adjustment)
- params marked per **verified usage** (each checked against the consuming query/cap, citations in comments); ambiguous ones deliberately left default
- `max_stale_age_hours` un-pinned — the #369 pin existed only because this metadata didn't exist (note: `pinned` also blocked operator `set_param`, so un-pinning restores runtime adjustability too)

## Exposure
Low in prod today (`NOUS_HEARTBEAT_TUNING_ENABLED` defaults false), but anyone flipping tuning on would get a self-defeating loop on the SelfInitiated and Health checks.

## Tests
3 new (volume relax decreases / tighten increases / set_param preserves flag). 138/138 heartbeat tests pass.

### DIFF
--- a/nous/heartbeat/checks.py
+++ b/nous/heartbeat/checks.py
@@ -50,10 +50,21 @@ class HealthCheck(BaseCheck):
         self._brain = brain
         self.interval = settings.heartbeat_health_interval
         self._params = {
-            "stale_decision_days": TunableParam("stale_decision_days", 7, 3, 30, 1),
+            # increases_findings: raising stale_decision_days WIDENS the
+            # created_at >= cutoff window in Brain.get_unreviewed → more
+            # findings. stale_fact_days is the opposite (older_than filter).
+            "stale_decision_days": TunableParam(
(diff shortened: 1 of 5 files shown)


tfatykhov:
@codex review

Issue #379, opened 2026-04-28
EpisodeSummarizer quality has 15x per-question-type variance — single-session-assistant 0.05 vs single-session-preference 0.78


tfatykhov:
## Finding

F056 PR #4 summary handler smoke produced first real production data on \`EpisodeSummarizer.summarize_episode\` (\`nous/handlers/episode_summarizer.py:104\`). With Haiku LLM-judge on 80 hand-curated transcripts spanning 6 LongMemEval question types:

| question_type | n | mean_kpc | mean_sf | summary_quality |
|---|---|---|---|---|
| single-session-preference | 13 | 0.92 | 0.85 | **0.781** |
| multi-session | 13 | 0.49 | 0.85 | 0.412 |
| single-session-user | 13 | 0.58 | 0.65 | 0.377 |
| knowledge-update | 14 | 0.39 | 0.68 | 0.263 |
| temporal-reasoning | 14 | 0.24 | 0.61 | 0.145 |
| **single-session-assistant** | 13 | 0.26 | **0.19** | **0.051** |
| **overall** | 80 | **0.475** | **0.637** | **0.303** |

Where:
- \`mean_kpc\` = mean key-point coverage (of N gold key-points, fraction the produced summary surfaced)
- \`mean_sf\` = mean summary faithfulness (1.0 = no unsupported claims, 0.0 = unsupported claims found, 0.5 = borderline)
- \`summary_quality\` = mean_kpc * mean_sf (F056 §D primary gated metric)

\`single-session-assistant\` is **~15× worse** than \`single-session-preference\`. Most strikingly, faithfulness drops to **0.19** — Haiku judges the produced summaries as containing unsupported claims on most assistant-action rows.

## Reproducer

\`\`\`bash
NOUS_EVAL_DB_NAME=nous_eval_scratch uv run python -m nous_eval.handlers.summary
\`\`\`

Report: \`reports/handlers/summary-20260428-171155.md\`

## Sample failure mode (single-session-assistant)

Fixture row \`s041\` transcript:
> User: Can you remind me to call mom on Sunday? Assistant: Sure, I've added a reminder for Sunday at 10 AM.

Gold key-points the summary should surface:
- "Assistant added a reminder to call mom"
- "Reminder is for Sunday at 10 AM"

Result: \`key_point_coverage=0.5\`, \`summary_faithfulness=0.0\` (Haiku flagged the produced summary as containing unsupported claims).

[...]

PR #510, opened 2026-06-10
fix: recall_recent reads structured summaries (parity with recall_deep) + unsummarized marker

tfatykhov:
User-reported gap: `recall_deep`'s parent-episode path was codex-patched to `COALESCE(structured_summary->>'summary', summary)` (tools.py:205-209), but `recall_recent` (tools.py:915+) never got the same fix — it reads legacy `e.title`/`e.summary` directly. For summarized episodes the columns converge, so it's invisible; for un-summarized episodes the raw creation echo surfaces with zero placeholder signal (which is exactly what masqueraded as a #379 NO-PADDING failure).

**Validation found it one layer deeper:** `episodes._list_recent` never populated `structured_summary` on `EpisodeSummary` at all — the schema field existed but stayed `None`, so the tool couldn't prefer it even in principle.

## Fix
- `_list_recent` populates `structured_summary`
- `recall_recent` prefers structured `title`/`summary`; un-summarized episodes get ` (unsummarized)` on the **title line** (short echoes become the title, so a body-line suffix wouldn't always be visible)

## Tests
`test_recall_recent_structured.py`: structured preferred over echo / echo shown but marked. 11/11 across the episode-rendering files.

### DIFF
--- a/nous/api/tools.py
+++ b/nous/api/tools.py
@@ -914,11 +914,21 @@ def create_nous_tools(brain: Brain, heart: Heart, settings: Settings | None = No
 
             lines = [f"Recent episodes (last {hours}h):"]
             for e in episodes:
-                title = e.title or (e.summary[:60] if e.summary else "Untitled")
+                # Prefer the summarizer's output over the legacy columns —
+                # mirrors recall_deep's COALESCE(structured_summary->>'summary',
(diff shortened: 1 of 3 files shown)


tfatykhov:
@codex review

PR #511, opened 2026-06-11
fix: audit Batch 0 — mechanical hardening (AS-3/6/7/10, CR-7, R-11)

tfatykhov:
First of the audit-remediation PRs. Batch 0 = the trivial, no-regression mechanical fixes from the reviewed plan (`docs/plans/2026-06-11-audit-fixes-plan.md`). Each item traces to a verified-real issue in `docs/reviews/verified-issues-2026-06-11.md` (15 verified-real issues distilled from ~87 audit findings via adversarial adjudication + personal code verification).

## Fixes
| ID | Issue | Fix | File |
|----|-------|-----|------|
| AS-10 | MCP `nous_recall` limit uncapped → expensive query / pool pressure | `min(limit, 100)` | `api/mcp.py` |
| AS-7 | `context_budget_overrides` accepted negatives → silent context underflow | after-validator rejects `< 0` | `config.py` |
| CR-7 | tool-cache key 64-bit → same-session collision returned wrong content | widen to 128-bit | `api/tool_cache.py` |
| R-11 | adjacency-boost `except` was silent | log on failure (still fail-open) | `api/retrieval_pipeline.py` |
| AS-3 | `send_file` accepted any outbound `chat_id` (exfiltration) | allowlist to configured chat | `api/telegram_tools.py` |
| AS-6 | `send_file` blocked the event loop on `open()` | read via `asyncio.to_thread` | `api/telegram_tools.py` |

## Tests
- New/updated unit tests for AS-7, CR-7, AS-3 (matching-chat allowed + mismatched-chat rejected), AS-6 (bytes uploaded). `test_send_custom_chat_id` was the AS-3 vuln case (sent to a non-configured chat) → rewritten to assert rejection.
- AS-10 is a self-evident `min()` cap covered by existing recall tests; a dedicated test needs real-PG + 100 seeded facts (disproportionate).
- R-3 (PipelineStats CE/MMR threading) deferred to the ranking PR where `heart.recall` is already touched.

## Verification
- Batch 0 test files: all green (`test_tool_cache.py`, `test_config.py`, `test_telegram_tools.py`).
- One unrelated pre-existing failure remains on this branch and on `main`: `test_default_equals_max_is_ok` (`.env` contamination — the test reads the repo `.env` so an ambient `NOUS_DAG_NODE_MAX_STALL_TIMEOUT` collides with its `max_timeout=900`). Not touched by this PR.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #512, opened 2026-06-11
fix: audit Batch 1 — write-path hardening (W-8 dedup race, W-7 bypass audit)

tfatykhov:
Second audit-remediation PR. Write-path fixes from the reviewed plan. Builds on #511 (Batch 0).

## W-8 — concurrent dedup race (P2, **was LIVE in prod**)
The fact_extractor tiebreaker runs its LLM check *outside* any transaction, so two concurrent learns of identical content could both pass dedup and double-insert (no UNIQUE on content). Fix: a per-`(agent, content)` **transaction-scoped advisory lock** at the top of `Heart._learn`, serializing the racers so the second sees the first's commit and dedups.

- **Advisory lock, not a UNIQUE constraint** — the plan review rejected the constraint because it would false-reject the legitimate F075 case (same content, different `event_date` = distinct events), mis-return ids on conflict, and fail migration on existing dup-active rows. The lock preserves all existing event_date-aware dedup semantics.
- Postgres-only via a dialect guard (`pg_advisory_xact_lock`); no-op on the serial SQLite test backend.
- **2 new concurrency integration tests** (real Postgres, own sessions): identical concurrent learns → exactly 1 active row; same content on distinct `event_date`s → 2 active rows (F075 preserved under the lock).

## W-7 — admission bypass audit (P2)
Adds an audit log distinguishing **authoritative** (`user_stated`/`identity`/`censor`) from machine-**derived** (`supersede`/`contradict`/consolidation) admission bypasses, so the derived-bypass false-positive rate is measurable.

**Bypass behavior is intentionally unchanged.** On closer reading: the `<30-char` content floor already runs upstream in `_learn`, and the consolidation bypass was deliberately added after admission *wrongly rejected valid merges* (see `bypass_sources` note, admission.py). The plan's structural reject-gate would have re-introduced that regression and broke 3 deliberate bypass tests, so it was dropped in favor of the observability the reviewer actually asked for.

## W-1 — deferred
The admission LLM call holding the open write transaction is a **perf-only P2** requiring a restructure of the hot `_learn` read/write interleaving + the concurrency benchmark the plan calls for. Too risky to bundle; it gets its own focused PR.

## Verification
- `test_admission.py` (56) + `test_admission_integration.py` (8, incl. the 2 new W-8 tests) green on Postgres.
- W-7 unit bypass tests (`test_supersede_bypasses`, `test_contradict_bypasses`, `test_cluster_consolidation_bypasses`) still green — behavior preserved.
- Pre-existing/unrelated: 3 `test_admission_dashboard.py` failures fail identically on `main` (verified via stash) — a separate data-isolation issue.
- Requires `NOUS_TEST_DB=postgres` for the integration lane (pgvector + advisory lock).

### DIFF
--- a/nous/heart/admission.py
+++ b/nous/heart/admission.py
@@ -48,6 +48,11 @@ DEFAULT_WEIGHTS = {
 
 DEFAULT_THRESHOLD = 0.55
 
+# W-7: bypass sources split into authoritative (fully trusted) vs derived
+# (machine-generated supersede/contradict/consolidation — bypass scoring but
+# must still carry an embedding). Any bypass source NOT listed here is derived.
+_AUTHORITATIVE_SOURCES = frozenset({"user_stated", "identity", "censor"})
+
 # Exponential decay: λ = 0.01/hour → half-life ≈ 69 hours (~3 days)
 RECENCY_DECAY_LAMBDA = 0.01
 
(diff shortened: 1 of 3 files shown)


tfatykhov:
@codex review

PR #513, opened 2026-06-11
fix: audit Batch 2 — resource & safety hygiene (CR-4 session leak, CR-10 compaction)

tfatykhov:
Third audit-remediation PR. Resource/safety fixes from the reviewed plan.

## CR-4 — in-process session-state leak (P2)
`end_session` pops 7 per-session dicts, but a crashed or abandoned session never calls it → unbounded growth (`_session_metadata` holds the full transcript). Fix:
- `_session_last_activity` stamp refreshed at the **start** of `pre_turn` (before the sweep, so an active session is never evicted mid-conversation — review guard).
- `_purge_session_state(session_id)` shared helper, now used by **both** `end_session` and the sweep so they can't drift on which dicts hold state.
- `_sweep_stale_sessions()` rate-limited to once per session-timeout window, evicts sessions idle past 2× the timeout.

Tests: `test_cr4_sweep_evicts_stale_sessions` (stale evicted, live retained), `test_cr4_sweep_is_rate_limited`.

## CR-10 — compaction value-substitution guard (P2)
The hallucination guard's multi-word partial-match rule excused value substitutions: `port 6379` (summary) vs `port 6380` (input) passed because the word `port` matched while the *number* changed. Fix: skip that rule when any word carries a digit (port/IP/version values), requiring the value to match. Names like `marcus webb` (unpacked from an email) are unaffected.

Test: `test_cr10_port_value_substitution_flagged`. All 5 existing guard tests still green (names have no digit-words; single-token versions/ports unaffected).

## Verification
- `test_compaction_hallucination_guard.py` (incl. new) + 2 new CR-4 tests: green.
- Full `test_cognitive_layer.py` on Postgres: my changes add 2 passing tests (39→41) with **zero new failures** — the 15 failures present are pre-existing on `main` (verified via stash; SAVEPOINT/env issues unrelated to this PR).

### DIFF
--- a/nous/api/compaction.py
+++ b/nous/api/compaction.py
@@ -203,8 +203,18 @@ def detect_hallucinated_entities(input_text: str, summary: str) -> list[str]:
         if ent in input_lc:
             continue
         words = ent.split()
-        if len(words) > 1 and any(
-            len(w) >= 4 and w in input_lc for w in words
+        # Rule 2 rescues a multi-word entity when one of its words appears in
+        # the input — meant for names unpacked from an email (`marcus webb`
+        # from `marcus.webb@acme.com`). CR-10: it must NOT rescue a VALUE
+        # substitution. `port 6379` (summary) vs `port 6380` (input) was
+        # passing because the word `port` matched while the number changed.
+        # So skip rule 2 when any word carries a digit (port/IP/version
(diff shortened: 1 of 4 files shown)


tfatykhov:
@codex review

PR #514, opened 2026-06-11
fix: audit Batch 4a — remove dead quality_block_threshold config (BR-1)

tfatykhov:
Fourth audit-remediation PR. A small, standalone P3 cleanup.

## BR-1 — misleading dead config
`quality_block_threshold` (config.py) had **zero code readers** — decision `quality_score` is advisory (surfaced for human review / dashboard); nothing gates on it. The `quality.py` docstring claimed `Block threshold: < 0.55 (enforced by guardrail, not here)`, pointing at the **unwired** CEL guardrail engine. Removed the field and corrected the docstring.

Forward-compatible: a stale `NOUS_QUALITY_BLOCK_THRESHOLD` in an operator `.env` is ignored (`extra="ignore"`).

## Deferred (needs a decision)
The rest of Batch 4 — **deleting the dead CEL `GuardrailEngine`** (BR-4/6: `Brain.check()` has no runtime caller, it's test-only) — removes a nominal capability, so per the plan it's **operator-gated**: kill it, or wire it into the deliberation path? Not included here.

## Verification
- `config` + `brain.quality` import clean; `quality_block_threshold` attribute gone.
- `test_config.py`: 7 passed (the 1 failure is the pre-existing `.env`-contamination `test_default_equals_max_is_ok`, unrelated).

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #515, opened 2026-06-11
fix: audit Batch 4b — wire the CEL guardrail engine (advisory) (BR-4/6)

tfatykhov:
Wires the dead CEL guardrail engine (you chose **wire** over kill). **Stacked on #514** (Batch 4a / BR-1) — base is that branch, so this diff shows only the wire.

## What
`GuardrailEngine` had no runtime caller (`Brain.check()` was test-only). Now `DeliberationEngine.finalize()` evaluates every finalized decision against active guardrails — `guardrail_blocked`/`guardrail_warned` events fire, activation counts increment, and matches are logged.

## Advisory, not enforcing — deliberately
The seed guardrails are **block-severity and unvalidated**. One (`no-critical-without-review`) is just `decision.stakes == 'critical'` — it would block **every** critical decision unconditionally. Hard-enforcing these would break the agent's ability to record critical/high-stakes/reasonless decisions. So the wire is advisory (log + events + counts + surface); the check is wrapped so an engine error never breaks `finalize`. Enforcement is a follow-up that first needs the seed CEL expressions fixed. Gated by `guardrail_check_enabled` (default True).

## Tests (`test_guardrail_wire.py`, mocked Brain — no DB)
- a blocking guardrail is evaluated but the decision is still finalized (advisory, not rejecting)
- the check is skipped when `guardrail_check_enabled=False`
- an engine exception does not break `finalize`

## Verification
- 3 new wire tests + 2 `test_brain.py` guardrail tests green.
- 2 `test_cognitive_layer.py` deliberation tests fail on this branch **and on `main`** (verified via stash) — a stale local test-DB schema (`procedures.superseded_by` missing) aborts the transaction during *context build*, before the guardrail check runs. Unrelated.

### DIFF
--- a/nous/cognitive/deliberation.py
+++ b/nous/cognitive/deliberation.py
@@ -9,6 +9,7 @@ from __future__ import annotations
 import logging
 import re
 from datetime import UTC, datetime, timedelta
+from typing import TYPE_CHECKING
 from uuid import UUID
 
 from sqlalchemy import select as sa_select
@@ -18,6 +19,9 @@ from nous.brain.brain import Brain
 from nous.brain.schemas import ReasonInput, RecordInput
 from nous.cognitive.schemas import FrameSelection
 from nous.storage.models import Decision as DecisionModel
+
+if TYPE_CHECKING:
+    from nous.config import Settings
 
 logger = logging.getLogger(__name__)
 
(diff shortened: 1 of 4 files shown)


tfatykhov:
@codex review

PR #517, opened 2026-06-13
fix: audit W-1 — move admission LLM utility call off the open write txn

tfatykhov:
## What

`FactManager.learn()` opened the write session, then inside `_learn()` called `AdmissionController.score()`, which made the Haiku **utility LLM call while the dedup/insert transaction (and the W-8 advisory lock) held a pooled DB connection**. A slow or timing-out Haiku call pinned that connection for its full duration, under lock — a write-path connection-pool hazard flagged in the 2026-06-11 audit (W-1).

## Fix

Precompute the utility score in `learn()` **before** opening the session and thread it through:

```
learn()  ──precompute_utility()──►  Haiku call (no txn, no lock)
   │
   └─ _learn(session, utility_override=…)  ──►  score(utility_override=…)
```

- `AdmissionController.precompute_utility()` runs `_score_utility` outside any session; returns `None` for bypass sources (so `score()` short-circuits them with no wasted call).
- `score()` gains a keyword-only `utility_override`; uses it when present, otherwise falls back to the original in-band LLM call.
- Gated on the same `>=30`-char floor `_learn` already enforces.

Behavior is identical — same `_score_utility` path, heuristic fallback preserved. Only the call site moves off the transaction. The single wasted call is for a fact that later turns out to be a Leg-2 dupe (rare; Leg-1 filters most upstream); documented in code.

## Tests

4 new tests in `TestUtilityPrecompute`:
- precompute returns a score for non-bypass sources / `None` for bypass
- `score()` honors the override **without** making an LLM call
- `score()` still computes utility when no override is supplied

Full `tests/test_admission.py` (60 tests) green. (`tests/test_facts.py` failures are pre-existing — pgvector `<=>` requires a real Postgres backend, unrelated to this change.)

### DIFF
--- a/nous/heart/admission.py
+++ b/nous/heart/admission.py
@@ -378,6 +378,16 @@ Respond with ONLY a number between 0.0 and 1.0."""
     # Public API
     # ------------------------------------------------------------------
 
+    async def precompute_utility(self, fact_input: FactInput) -> float | None:
+        """W-1: compute the (LLM) utility score OUTSIDE any DB session so
+        score() doesn't hold a pooled connection through the Haiku call while
+        the dedup/insert transaction (+ the W-8 advisory lock) is open. Returns
+        None for bypass sources (score() short-circuits them — no wasted call).
+        """
+        if fact_input.source in self.config.bypass_sources:
+            return None
+        return await self._score_utility(fact_input)
+
     async def score(
         self,
         fact_input: FactInput,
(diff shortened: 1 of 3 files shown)


tfatykhov:
@codex review

tfatykhov:
@codex review — addressed P2: precompute now only runs when learn() owns the session; injected sessions keep inline scoring.

PR #528, opened 2026-06-14
fix(cache): stable tool superset across frames to stop prompt-cache busting

tfatykhov:
## Summary

Prod prompt-cache audit (read-only, `nous_system.context_log`, 2026-05-14→06-14). Caching is **healthy overall (88.2% hit rate, 11.8% creation, ~0% uncached input)** and the tier grouping is correct — no volatile content leaks into the cached prefix. One real, quantified inefficiency: **frame-scoped tool schemas bust the entire cached prefix on every frame change.**

`FRAME_TOOLS` gives each cognitive frame a different tool array (conversation=27, question=19, decision=14, creative=8, task=all, debug=24). Tools sit at the **front of the Anthropic cacheable prefix**, so when the frame changes between turns — constant in interactive chat — the tools array changes and busts everything downstream (tools + all system blocks). The turn's first call becomes a full re-creation.

**Measured cost:** of 32 warm full-prefix re-creations with a prior in-session call, 13 were >5 min apart (legit 5-min TTL) but **19 were ≤5 min real busts, 17 of them on a frame change** — **~424K cache-creation tokens (10.1% of all cache-creation) per month**, re-created at 1.25× that could have been 0.1× cache reads.

## Fix

When `NOUS_STABLE_TOOL_SET_ENABLED` (default `true`), `ToolDispatcher.available_tools` collapses all non-`initiation` frames to the `task` (`*`) tool superset, using the **effective frame as both the schema-cache key and the FRAME_TOOLS lookup** so every conversational frame returns a byte-identical, deterministically-ordered list. The tools array no longer varies turn-to-turn → the cached prefix holds across frame changes.

- `FRAME_TOOLS` still drives the **textual** frame instructions (`_get_frame_instructions`); only the wire-level tool array is stabilized.
- `initiation` keeps its distinct minimal set (isolated one-time protocol, never interleaved with chat).
- Per-turn refuse (F078) and subtask tool filters run **after** `available_tools` and are unchanged.
- Behavioral note: read-only frames (question/creative) gain action tools — frame instructions already steer tool use, and the `task` frame already sent all tools. Flag-gated (default on) for instant rollback.

## Architecture review

`feature-dev:code-reviewer` returned **APPROVE-WITH-NITS**; both issues incorporated:
1. **Redirect the cache KEY** to `effective_frame` for both lookup and store — otherwise the per-frame cache is bypassed on every call.
2. **Exclude `_INITIATION_ONLY_TOOLS` = {store_identity, complete_initiation}** from the `*` superset — also fixes a pre-existing `task`-frame leak of protocol tools into normal turns.

Also clarified in docs: F036's "tool schema cache" is only **process-side Python memoization** — it does nothing for the Anthropic-side prefix.

## Tests

- New `tests/test_stable_tool_set.py`: identical lists across all conversational frames, `initiation` distinct, protocol-tool exclusion, single-key cache, deep-copy isolation, flag-off legacy gating (7 tests).
- Existing per-frame tests (`test_f036_tool_cache.py`, `test_tools.py`) pinned to `stable_tool_set_enabled=False` to isolate the legacy gating path they assert.
- `test_runner_background`, `test_f061_runner_subtask_hooks`, `test_tool_loop`, `test_streaming` all green (63 passed). The 4 `test_tools.py` `TestLearnFact`/`TestRecallDeep` failures are pre-existing pgvector-on-sqlite issues (need the Postgres lane), confirmed identical on the base branch.

## Operator

Default-on. To revert: `NOUS_STABLE_TOOL_SET_ENABLED=false` + restart (no redeploy).

Audit: `docs/reviews/2026-06-14-llm-caching-audit.md`. Decision: `ea4f4adb`.

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -477,7 +477,8 @@ DB connection vars are **unprefixed** (shared with docker-compose). All others u
 | `NOUS_CACHE_BREAK_DETECTION_ENABLED` | `true` | Enable cache break detection logging (F036) |
 | `NOUS_CACHE_SPLIT_SYSTEM_PROMPT` | `true` | Enable 3-tier system prompt splitting (F036) |
 | `NOUS_CACHE_SINGLE_BREAKPOINT` | `true` | Use single cache breakpoint strategy (F036) |
-| `NOUS_TOOL_SCHEMA_CACHE_ENABLED` | `true` | Cache tool schemas per frame (F036) |
+| `NOUS_TOOL_SCHEMA_CACHE_ENABLED` | `true` | Cache tool schemas per frame (F036) — process-side Python memoization only; does NOT affect the Anthropic prompt cache |
+| `NOUS_STABLE_TOOL_SET_ENABLED` | `true` | Send a STABLE tool superset across all non-`initiation` frames instead of frame-scoped tool arrays. Tools sit at the front of the Anthropic cacheable prefix, so frame-scoped tools busted the whole prefix on every frame change (~10% of prod cache-creation tokens, measured 2026-06-14). Collapses conversational frames to the `task` (`*`) superset; `initiation` keeps its distinct minimal set; `store_identity`/`complete_initiation` are excluded from the superset. `FRAME_TOOLS` still drives the textual frame instructions. Set `false` to restore per-frame tool gating. |
 | `NOUS_DAG_ENABLED` | `true` | Enable DAG orchestration (F038) |
 | `NOUS_CORRECTION_EXTRACTION_ENABLED` | `true` | Enable correction learning pipeline (F039) |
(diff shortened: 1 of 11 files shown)


tfatykhov:
@codex please review. Focus areas: (1) the effective_frame redirect in ToolDispatcher.available_tools — is the cache key correctly routed for both lookup and store across all frames? (2) the _INITIATION_ONLY_TOOLS exclusion from the '*' superset — any frame that legitimately needs those tools and would now lose them? (3) interaction with the downstream F078 refuse-denylist and subtask _SUBTASK_EXCLUDED_TOOLS filters (run after available_tools) — any double-filtering or ordering issue? (4) default=True on the live chat path — any behavioral risk from read-only frames gaining action tools beyond what frame instructions already steer?

PR #531, opened 2026-06-14
feat(F044): tinyHippo-Lite v1 — STC consolidation (telemetry-only slice, co-designed with Nous)

tfatykhov:
## Summary

First slice of **F044 tinyHippo-Lite** (algorithmic sleep consolidation), **co-designed turn-by-turn with the deployed Nous agent** over its `/chat` API. Telemetry-only: the STC (Synaptic Tagging & Capture) state machine + PRP promotion gate. **NO** homeostatic α-downscale, weight-floor mortality, or replay telemetry (deferred). Flag-gated `NOUS_TINYHIPPO_LITE_ENABLED`, **default off → inert**.

Notably, F044 was **spec-only** before this (PR #422 merged the design doc with zero implementation); this is the first code.

## What's in v1

- **migration 061**: `brain.graph_edges` += `consolidation_state` (`tagged`|`consolidated`, default `tagged`), `ltp_count` (default 0), `last_ltp_at`; CHECK constraint + partial index on the tagged frontier. Column-level defaults keep feature-off **bit-identical** to pre-F044 main.
- **`nous/brain/tinyhippo_lite.py`**: idempotent promotion gate (`tagged` + `ltp_count >= PRP` → `consolidated`) + telemetry aggregate (the ltp histogram that answers "is the reinforcement signal starved?") + `increment_ltp_on_rederivation`.
- **`_phase_stc_consolidation`** wired after F040 densification, before F053 dead-edge prune. No weight change, no deletion.
- **Reinforcement hooks at the two LIVE similarity-linker sites** (`brain._auto_link`, `graph_linker.create_edge`). They increment on the existing `ON CONFLICT` branch via a separate targeted UPDATE, so the linkers' rowcount-based new-edge accounting is unchanged. `graph_linker` excludes `structural` provenance (F070 chunk anchors) so deterministic rebuilds don't inflate the counter. Deterministic densifier builders are deliberately **not** hooked.
- config: `NOUS_TINYHIPPO_LITE_ENABLED` (False), `NOUS_TINYHIPPO_PRP_THRESHOLD` (3).

## Design decisions (co-designed with Nous)

- **Why only live linkers, not all upsert sites:** Nous flagged that F040 densification *targets orphans*, so hooking it would starve the re-derivation signal (false null); and the deterministic builders (`happened_before`, co-mention, F070 structural) rebuild the same edges every cycle, which would inflate `ltp_count` uniformly. The live similarity linkers are where a conflict means *genuine independent rediscovery* — there, **conflict rate == re-derivation rate, measured for free**.
- **Drift-guard test** (Nous's requested CI tripwire) instead of a risky shared-helper refactor: asserts exactly 2 `F044-STC-HOOK` sites, so a 3rd bypassing producer fails CI.

## Tests (`tests/test_f044_tinyhippo_lite.py`)

- drift-guard (pure source): exactly 2 hook sites in `brain.py` + `graph_linker.py`.
- Postgres lane: promotion gate promotes at PRP & is idempotent; one conflict == one increment (no debounce); 3 re-derivations → `consolidated`.
- All green; 118 edge-linking/sleep tests unaffected (hooks inert at default-off).

## Validated end-to-end on a dense prod-shape graph

Applied migration 061 to the `:5433 nous_eval_prod` clone (**12,149 edges**). Cycle-0 baseline reads clean (all tagged, ltp 0); re-deriving 3 real edges ×3 drove them through the PRP gate to `consolidated`. Instrument confirmed on real data.

## Next (deferred, per co-design)

- **Write re-derivation** is only observable where live activity re-fires the linkers (prod), not on a frozen clone — measuring it means flipping the flag **shadow-on in prod** (safe: additive cols, idempotent gate, no weight change/deletion). Operator's call.
- **v1.1**: buffered recall-touch counter (reinforce on retrieval), measured on the clone via real recall-log replay — the theoretically-faithful STC trigger and where the bet likely lives.


tfatykhov:
@codex please re-review at HEAD 0a150bb. This pushes the full F044 chain (the prior review at 6ef09f0 predated 5 unpushed commits) plus a fix for the two P1s you flagged:

- **P1b** (link_fact_to_decisions / link_fact_to_facts bypassed create_edge): added the F044-STC-HOOK at both ON CONFLICT upsert sites; drift guard now pins 4 live hook sites.
- **P1a** (handlers commit only when edges_created>0, dropping reinforcement-only LTP increments): all four similarity-linker handlers now OR the F044 flag into their commit gate. Default-prod (flag off) behavior is byte-identical.

Source-pinned regression test added for both.

tfatykhov:
@codex re-review at HEAD 2694e53c44. Addressed all three round-2 findings:

- **P1** (consolidated boost on master flag): added dedicated opt-in flag `tinyhippo_consolidated_boost_enabled` (default off); retrieval gate now ANDs it, so master-alone shadow mode is byte-identical again.
- **P2** (alpha unbounded): field validator enforces `(0.0, 1.0]` decay bound; 0.42 experiments stay valid.
- **P2** (flush race): `flush_recall_touches` subtracts the snapshot synchronously before awaiting, restores on failure — concurrent mid-flush touches survive.

3 new tests added.

tfatykhov:
@codex re-review at HEAD c2f222e12d. Round-3 findings addressed:

- **P2** (downscale decayed structural edges): `_DOWNSCALE_SQL` now excludes the deterministic tier (`extraction_method IS DISTINCT FROM 'deterministic'`), mirroring the LTP hook's structural skip — F070 part_of/adjacent-chunk anchors are exempt.
- **P2** (PRP threshold): `tinyhippo_prp_threshold` now `ge=1` so a 0/negative value fails config init instead of promoting the whole graph on first sleep.

2 new tests.

tfatykhov:
@codex re-review at HEAD a9341b2573. Round-4 findings (both on the dev-only control script) addressed: random-exempt arm now samples only downscale-eligible (tagged, non-deterministic) rows; sanity arm now does a true uniform scale of all tagged edges so the ~baseline expectation holds post-exemption. All production-code findings from rounds 1-3 are resolved.

tfatykhov:
@codex re-review at HEAD 0eb5ed186c. Round-5 finding addressed: `_record_recall_reactivation` now excludes the deterministic tier (`extraction_method IS DISTINCT FROM 'deterministic'`), so all three F044 reinforcement touchpoints (write hook, downscale, recall-touch) operate on the same associative edge set — structural F070 anchors are never reinforced/promoted/boosted.

tfatykhov:
@codex re-review at HEAD dade882540. Round-6: applied the same control-arm fix to the sibling `f044_prod_revalidate.py` (random arm samples downscale-eligible rows; sanity arm is a true uniform scale). All production code has been codex-clean since round-5; remaining rounds have been dev-only diag-script methodology.

tfatykhov:
@codex re-review at HEAD 77ae529cc8. Round-7 addressed: `_RECALL_TOUCH_BUFFER` is now keyed by `(agent_id, source, target, relation)`; `record_recall_touches`/`flush_recall_touches` thread `agent_id`, flush drains only the sleeping agent's keys, and the LTP UPDATE filters on `agent_id`. Multi-agent isolation test added.

tfatykhov:
@codex re-review at HEAD a0a22c5792. Round-8 addressed:

- **Production**: both LTP UPDATEs now carry `extraction_method IS DISTINCT FROM 'deterministic'`, so a non-structural producer conflicting onto a deterministic edge can't promote it. Structural anchors are now exempt at all four reinforcement touchpoints (producer check, recall-touch read, downscale, LTP write).
- **Diag**: all six warm-up scripts pass agent_id to `flush_recall_touches` (the round-7 signature change had left a TypeError); beam scripts flush per-agent; prod A/B scripts force `tinyhippo_lite_enabled` in warm-up Settings.

tfatykhov:
@codex re-review at HEAD 8479f532ce. Round-9 addressed: `run_stc_consolidation` now commits the recall-touch flush in its own transaction ahead of promotion, so a promotion/commit abort can't roll back already-drained touches (promotion is idempotent and re-runs next sleep).

tfatykhov:
@codex re-review at HEAD 2126a12d8e. Round-10 — both findings closed by one change: `flush_recall_touches` now commits INSIDE its restore guard (`commit=True` default), so a commit failure also restores the drained snapshot (no lost touches). `run_stc_consolidation` drops the redundant explicit commit; tests pass `commit=False` for fixture isolation. The diag warm-ups' `flush_recall_touches(sess, AGENT)` now self-commits before promotion, making the flush durable in its own transaction.

tfatykhov:
@codex re-review at HEAD fd41946a31. Round-11 addressed: the A/B ON arms now also enable `tinyhippo_consolidated_boost_enabled` (prod graph A/B) / export `NOUS_TINYHIPPO_CONSOLIDATED_BOOST_ENABLED=$flag` (BEAM 3-arm), so consolidation actually affects ranking in the ON arms instead of measuring the baseline scorer. This is the last cascade item from the round-2 flag-decoupling; all production code has been codex-clean since round 10.

PR #533, opened 2026-06-16
fix(heart): tolerate legacy list-valued structured_summary (sleep-cycle crash)

tfatykhov:
## Problem
A pre-#525 episode row whose `structured_summary` was persisted as a bare JSON **list** crashes `EpisodeSummary`/`EpisodeDetail` pydantic validation. Because `list_recent` builds the whole batch eagerly, **one** bad row takes down `list_episodes` entirely — killing sleep-cycle procedure learning:

```
ValidationError: 1 validation error for EpisodeSummary
structured_summary  Input should be a valid dictionary [input_value=["A truncated body..."], input_type=list]
  ... heart/episodes.py:389 EpisodeSummary(
  ... handlers/procedure_learner.py:373 _learn_from_episodes -> list_episodes(limit=50)
```

## Root cause
- **Write path is already fixed** by PR #525 (`_summarize_single` skips non-dict LLM parses). The crashing row is **legacy** — written before #525.
- **Read-side gap:** `episodes.py:470` already guards with `isinstance(e.structured_summary, dict) else None`, but `:400` (`list_recent`) and `:688` (`_to_detail`) passed the raw DB value → crash.

## Fix
Mirror the existing `:470` guard at both unguarded read sites. A legacy bad row is now treated as "no structured summary" (`None`) instead of crashing the whole batch — the episode itself is still listed.

## Test
`test_list_recent_tolerates_legacy_list_structured_summary` — inserts a list-valued `structured_summary` row and asserts `list_episodes`/`get_episode` return it with `structured_summary=None` instead of raising. Full `test_episodes.py` green (12/12).

## Operator note (optional)
The read guard makes legacy rows harmless. If desired, prod's bad rows can be nulled directly: `UPDATE heart.episodes SET structured_summary = NULL WHERE jsonb_typeof(structured_summary) <> 'object';`

### DIFF
--- a/nous/heart/episodes.py
+++ b/nous/heart/episodes.py
@@ -397,7 +397,10 @@ class EpisodeManager:
                 # legacy creation-time echo even for summarized episodes —
                 # the recall_deep parent-episode path already COALESCEs to
                 # structured_summary->>'summary' (tools.py, codex fix).
-                structured_summary=e.structured_summary,
+                # Guard against legacy pre-#525 rows where a bare-list summary
+                # was persisted — a non-dict crashes EpisodeSummary validation
+                # and takes down the whole batch (sleep-cycle procedure learner).
+                structured_summary=e.structured_summary if isinstance(e.structured_summary, dict) else None,
             )
             for e in episodes
         ]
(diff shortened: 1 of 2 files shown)


tfatykhov:
@codex review

PR #534, opened 2026-06-17
feat(F024): Inbound multimodal attachments (Telegram + REST -> Claude)

tfatykhov:
## Summary

Implements **F024 — Inbound Multimodal Attachments** (revives the long-stale `docs/implementation/011.2-multimodal-file-support.md`, which was Draft and never built — `Message.content` was bare `str`). Users can now send **images** (JPEG/PNG/GIF/WebP), **PDFs**, and **text/code files** to the agent via **Telegram** and the **REST API**; Claude analyzes them in-turn, originals are saved to the workspace with a retrievable memory reference, and base64 never touches the database.

**Gated by `NOUS_ATTACHMENTS_ENABLED` (default `false`) — lands dark.**

### What it does
- **Telegram**: bot downloads photos/documents/static-stickers via `getFile` (raw httpx — the spec's `python-telegram-bot` Phase 2 was rewritten), base64-encodes, and forwards through the existing `/chat/stream` proxy. Helpful errors for audio/voice/unsupported types.
- **REST** `/chat` + `/chat/stream`: accept an `attachments[]` array (base64 JSON); size is **recomputed server-side from base64** (client-declared size is never trusted); a request-body cap returns 413.
- **Runner**: builds Claude `image`/`document`/text content blocks; `_build_api_payload` already passed list content through. After the turn, the live user message is compacted (blocks → on-disk reference, base64 cleared) on **every** exit path (success / exception / censor-block).
- **Persistence**: originals saved under `workspace_dir/attachments/<session>/` (sandboxed by `builtin_tools._validate_path`, so the agent's own `read_file`/`bash` can retrieve them); a Heart fact records a memory-reference (grounded on Claude's analysis); text/code file bodies are chunk-ingested into `episode_chunks` (reuses an extracted `ingest_document_text` helper). **Base64 is stripped at the `_save_conversation` serialization boundary — it never reaches the DB.**
- **Token estimation**: `TokenEstimator` is now block-aware so a base64 image no longer falsely triggers history compaction.

### Scope / boundaries
- Native Claude types only: images + PDF. Text/code files injected as text. `.docx`/`.xlsx`/audio/video are rejected with a helpful message (no new deps).
- No `pypdf` — PDF bodies aren't chunk-ingested; PDFs are analyzed in-turn + saved + fact-referenced (the analysis is stored as the summary).
- No media-group/album batching, no per-user rate limiting (deferred).
- Requires a vision-capable `NOUS_MODEL` when enabled (documented).

### Notable bugs caught in review (and fixed)
- A DB-backed test under **prod admission config** (`shadow_mode=false`, threshold 0.60) revealed attachment facts were being **silently rejected**: `source_text` was the file path, so admission's ROUGE-L confidence was 0.00 and the composite score 0.5625 fell below 0.60. Fixed by grounding `source_text` on the analysis text.
- The bot token could leak into logs via `raise_for_status()`; download failures now raise a token-free error.

### Process
Plan: `docs/superpowers/plans/2026-06-17-inbound-multimodal-attachments.md`. A 3-agent pre-implementation review found 4 P0s; subagent-driven implementation followed with per-task and whole-branch review (final verdict: merge-ready).

## Test Plan
- [x] `tests/test_attachments.py` — pure functions (classify / validate / build / compact / sanitize)
- [x] `tests/test_attachment_store.py` + `tests/test_attachment_store_integration.py` — persistence, fact landing (DB-backed), text-file ingest
- [x] `tests/test_attachments_integration.py` — runner builds blocks + compacts; base64 never persisted
- [x] `tests/test_token_estimator_multimodal.py` — no false compaction on images
- [x] `tests/test_rest_attachments.py` — endpoint accepts attachments, server-side size, 413 cap
- [x] Regression: `tests/test_runner.py`, `tests/test_compaction.py`, `tests/test_rest.py` — **159 passed**, text-only behavior unchanged
- [ ] Manual: send photo / PDF / .py / unsupported via Telegram with `NOUS_ATTACHMENTS_ENABLED=true`


tfatykhov:
@codex review

tfatykhov:
@codex review

tfatykhov:
Both codex P2 findings are addressed in `06893c6` (codex hit its usage limit and could not re-review):

- **P2 — text-file ingest after base64 cleared (runner.py):** `maybe_ingest_text_file` was running on the success branch, *after* the `finally`'s `compact_message_for_history` cleared `data_base64` → empty body → ingest skipped. Moved it into the persist block (bytes still present, before the tool loop) in both `run_turn` and `stream_chat`; `record_attachment_fact` stays on the success path (it needs the analysis, not the bytes). Now called exactly once per attachment.
- **P2 — empty base64 for a supported type (attachments.py):** `validate_attachment` now rejects `image`/`document`/`text_file` attachments with empty/omitted `data_base64` ("appears to be empty"), so a malformed request no longer becomes an opaque Anthropic 500.

CI green on `06893c6` (Lint + Tests). New unit test `test_validate_attachment_rejects_empty_data`; full suite 77 passed under Postgres.

tfatykhov:
## Deep-review remediation round (4 reviewers: architecture advisor + silent-failure + adversarial-security + test-coverage)

A second deep review found issues the per-task passes and codex missed. The **core leak guarantee held** (no new base64→DB path), but real defects were found and are now fixed across `90e6a4b..fcf0ba3` (CI green, full suite **213 passed**):

**Security / data-loss**
- **Filename → prompt injection:** `sanitize_filename` preserved whitespace incl. newlines, which were interpolated into the `--- File: … ---` text-block header (forgeable via Telegram document / REST filename). Now collapses all whitespace.
- **Caption data-loss:** a user caption starting with `--- File: ` was misclassified as a file body and dropped on persist. Synthesized file blocks now carry a zero-width-space sentinel (invisible to the model, uncollidable with user text); detection keys off that, not the rendered prefix.

**Robustness / observability**
- `_parse_attachments` no longer 500s on a malformed (non-dict) `attachments[]` entry — bad entries are skipped.
- The two "log-and-continue" memory writes now actually observe failure: `maybe_ingest_text_file` inspects `ingest_document_text`'s `{error,code}` return (WARNING on `embed_failed`/`no_episode`/`vector_mismatch`), and `record_attachment_fact` logs a `FactRejected` return (both signal failure by *return value*, so the prior `except` was dead for those cases).
- Telegram downloads are now per-item: one failed download no longer discards the others, and a warning is always surfaced on ≥1 failure.
- Aggregate text-file size cap (2 MB across a message) prevents a context-overflow → opaque Anthropic 400; `find_cut_point` is now block-aware.

**Core value**
- Attachment facts + text-file ingest are now linked to the active episode (`cognitive.get_active_episode_id`) instead of `None` — restoring graph-augmented recall, the point of the memory reference.

**Tests added**
- `stream_chat` attachment path (build + compaction); base64 stripped on the **exception** path with **no** fact recorded; **censor-block** path stays text-only; a new `tests/test_telegram_attachments.py` (photo/document/sticker/voice extraction, **token redaction**, partial-failure-keeps-others, animated-sticker skip); `/chat/stream` attachment + non-dict-entry REST tests.

**Documented as v1 limitations** (not fixed this round): no attachment-file cleanup (unbounded disk on a persistent `workspace_dir` volume; moot while the flag is off), and client-retry orphan files + fact↔chunk `source_ref` skew on a failed-then-retried turn.

Ruled out as non-issues: path/traversal escape, media-index mapping, unknown-content_type drop.

tfatykhov:
@codex review

PR #535, opened 2026-06-17
feat(F024.1): PDF text ingestion (pypdf + Claude transcription fallback)

tfatykhov:
## Summary

**F024.1 — PDF text ingestion** (follow-up to F024 #534). Inbound PDFs are now **chunk-ingested into `heart.episode_chunks`** so their *contents* are recall-searchable — closing the image/PDF-vs-text-file asymmetry where only text/code files had their bodies ingested.

**Hybrid extraction, on by default:**
- **`pypdf`** extracts text deterministically (fast, complete, handles long PDFs). This path ingests **inline** (same as text files).
- If `pypdf` yields too little text (< 100 chars → scanned/image-only PDF), fall back to a **Claude transcription** call (cheap model, `claude-haiku-4-5-20251001`), run **fire-and-forget** so it never blocks the turn.
- Both paths reuse the existing `ingest_document_text` helper.

No PDF reaches the API a second time on the common path; the model is only invoked for scanned PDFs `pypdf` can't read.

### Changes
- `pyproject.toml`: `pypdf>=5.0,<6.0` added to the `agent` optional group (prod ships `.[runtime,agent,rerank]`).
- `nous/config.py`: `attachments_ingest_pdfs` (default `true`), `attachments_pdf_transcription_model`, `attachments_pdf_max_transcription_tokens` (bare annotations, matching the F024 block).
- `nous/api/attachment_store.py`: `_extract_pdf_text` (guarded `pypdf` import — degrades to `""` if absent), `maybe_ingest_pdf` (guard → decode → off-thread extract → inline ingest or fire-and-forget fallback; whole body wrapped so it never breaks a turn), `_transcribe_and_ingest_pdf` (best-effort transcription + ingest), `_parse_transcription_response` (matches the real `ApiResponse` shape).
- `nous/api/runner.py`: wired into the attachment persist block in **both** `run_turn` and `stream_chat`, passing `self._api`.

### Caveats (documented)
- Long **scanned** PDFs truncate at the transcription output cap (logs a warning; page-range paging deferred). Text PDFs via `pypdf` are complete.
- Fire-and-forget fallback is best-effort (lost if the process exits before it finishes) — matches the codebase's existing fire-and-forget convention (F026/F055).

## Test Plan
- [x] `tests/test_pdf_ingest.py` (6) — pypdf inline ingest, scanned fallback transcription, `create_task` scheduling, disabled/non-PDF skip, missing-pypdf + garbage-bytes degrade to `""` without raising.
- [x] `tests/test_attachments_integration.py` — runner invokes `maybe_ingest_pdf` for a PDF attachment (both paths share the code).
- [x] Regression: `tests/test_runner.py` green; full PDF+integration+runner run **68 passed**.
- [x] Independent review verdict: SHIP (fire-and-forget safety, base64 lifecycle, response parsing, guards, dep degradation all verified).
- [ ] Manual: send a text PDF (→ pypdf, searchable in recall) and a scanned PDF (→ transcription fallback) via Telegram.


tfatykhov:
@codex review

tfatykhov:
Fixed in 3903b7a — added a non-empty `system` to the transcription payload (and a regression test asserting it), so the default sdk backend's `_payload_to_kwargs` no longer KeyErrors and silently kills the scanned-PDF fallback. Good catch.

tfatykhov:
@codex review

tfatykhov:
Fixed in 86590e3 — `session_id` is now threaded into the scanned-PDF fallback so `ingest_document_text` can resolve the active episode (parity with the pypdf path), with a regression assertion. Both P2s addressed.

tfatykhov:
@codex review

PR #539, opened 2026-06-20
feat(dashboard): Svelte 5 dashboard v2 — reactive rewrite (refresh-state + mobile)

tfatykhov:
@
## Why

Two user-reported pains, one root cause: every dashboard view rendered via `container.innerHTML = html`, so each auto-refresh poll **destroyed UI state** (expand/collapse, scroll, filters — only `ledger.js` hacked around it), and the partial responsive CSS left **mobile broken**. Both trace to "blow-away-and-rebuild rendering + no client state model."

This rewrites the dashboard as a reactive **Svelte 5 + Vite + TypeScript** app where keyed diffing preserves component-local state for free, and real responsive components replace the media-query retrofit.

## What

A new self-contained app in `dashboard-app/` that builds to `static/dashboard-v2/dist` and is served by the existing Starlette app at `/dashboard/v2`. Strangler migration; `/dashboard` now **307-redirects** to it and the legacy vanilla-JS dashboard is retired.

- **Foundation**: typed API client, a shared poll store (recursive `setTimeout`, in-flight guard, abort-on-stop, error-retains-last-good), hash router, viz wrappers around the existing **Chart.js / Cytoscape / D3** (reused, not rewritten), shadcn-svelte UI primitives, responsive app shell (drawer a11y, `100dvh`, safe-area).
- **All 15 views migrated** with field-accuracy checked against the real backend (`dashboard_queries.py` + `rest.py`): overview, graph, browser, decisions, activity, heartbeat, observability, health, admission, rubric, execution(ledger), cache, density, dag, subtasks.
- **The headline win**: refresh preserves expand/scroll/filter state by construction across every view; the `ledger.js` manual state-scrape is **deleted**.
- **Cutover**: `/dashboard` → `/dashboard/v2/` redirect, legacy mount removed, ~8,700 lines of legacy assets deleted, `node:22` multi-stage Dockerfile builds `dist` into the image.

## Notable fixes found during the port
- Heartbeat legacy field-name bugs (`circuit_breaker_open` vs `_tripped`, `interval` vs `interval_seconds`).
- Browser decision detail was silently empty in legacy → now lazy-loads `/decisions/{id}`; per-tab search/pagination now survives tab switches.
- Cross-platform serving: Windows Python maps `.js` → `text/plain`, which blocks ES-module execution → `_NoCacheStaticFiles` now forces JS/CSS MIME types.
- `.dockerignore` excludes node_modules/dist/.venv/.git.

## Process
Design → spec → plan, 3-agent plan review (caught two whole-view design errors + a startup NameError before any code), then strangler implementation in 4 phases with per-task spec/quality review. Specs/plan under `docs/superpowers/`.

## Verification
- `npm run check` (tsconfig.app.json) **0 errors**, 14 Vitest tests, clean build.
- `docker build --target dashboard` succeeds (832 modules).
- Live boot verified: `/dashboard`→307→`/dashboard/v2/`, `/dashboard/v2/`→200 SPA, `/dashboard/*` JSON routes unaffected, legacy `index.html`→404. Desktop browser-tested across all Phase-2 views + the redirect landing.

## Caveats before merge/deploy
- **Mobile is desktop-tested only** — the responsive code (drawer/`dvh`/card-tables/bottom-sheets/44px) is in place and type-checked but not yet validated on a real device / DevTools device-mode.
- **Dark-only** (no light/dark toggle) — matches the legacy dashboard; intentional.
- **Deploy requires the Docker rebuild** — `dist` is gitignored and built by the new Dockerfile stage; the cutover leaves no legacy fallback at `/dashboard`.

@


tfatykhov:
@codex review

tfatykhov:
@codex review — fixed the P1: load-once views now use fetch-once mode (interval<=0, no reschedule) instead of the MAX_SAFE_INTEGER sentinel; added a regression test.

tfatykhov:
@codex review — addressed both P2s: /dashboard redirect is now gated on the v2 build existing (no redirect-to-404 on source installs); the poll store's in-flight branch no longer double-schedules timers (regression test added).

tfatykhov:
@codex review — addressed all 5 P2s: moved censor-edit init and decision-detail loading out of render (now onrowclick), clear pending poll timers before rescheduling, router ignores non-route fragments (skip link), and facts load via onMount instead of a search-subscribed effect. 2 regression tests added.

tfatykhov:
Round-4 review re-flagged two items that are **already fixed in 9380453** (verified in code + covered by passing regression tests):
- *Clear poll timeout before rescheduling* — `schedule()` now calls `clearTimeout(timer)` before arming a new one (test: "refresh() while a poll timer is armed does NOT cause double-polling").
- *Skip link routing* — fixed router-side (not at the App.svelte anchor): `parse()` returns null for non-route fragments and both init + hashchange skip on null, so `#main-content` no longer resets the route (test: "unknown fragment leaves current route unchanged").

All genuine findings across the review (1 P1 + 7 P2s) are addressed. CI green, mergeable. Merging.

PR #541, opened 2026-06-23
feat(brain): agent-facing decision resolution to close the calibration loop

tfatykhov:
## Summary

Consulted the live Nous agent about a decisions tool it needs for self-evaluation, verified its claims against code, and built the missing primitive.

**The gap:** Nous can `record_decision` and recall, but had **no session-time way to persist a decision's outcome**. Its triage sweeps were read-only, so the pending queue never drained and noise accumulated. (Nous's own framing — "calibration denominator is empty" — was empirically false: 662/771 decisions are already auto-reviewed by a handler. The real gap is the agent's *own* sweep classifications + the absence of `noise`/`superseded` states.)

`Brain.review()` + REST `POST /decisions/{id}/review` already perform the mutation, so this is mostly **tool wiring** plus two new outcome states.

Cites Nous Brain decision `06d62894`.

## Changes

- **New agent tools** `resolve_decision`, `resolve_decisions` (batch), `list_decisions` (surfaces existing `Brain.list_decisions`), wrapping the pre-existing `Brain.review()`. Gated by `NOUS_DECISION_RESOLUTION_ENABLED` (default on, kill-switch).
- **Migration 062** widens `ck_decisions_outcome` to add `noise` + `superseded` and adds a `superseded_by` self-FK. Drops both the inline (`decisions_outcome_check`) and ORM (`ck_decisions_outcome`) constraint names for idempotency; the new CHECK is a strict superset so existing rows pass. `init.sql` + `models.py` updated to match.
- **Calibration** denominator switched from `outcome != 'pending'` to the scorable set `IN ('success','partial','failure')`, so `noise`/`superseded` are **excluded** (otherwise they'd score as `0` = failure).
- `Brain.review`/`review_many` extended for `superseded_by`; `ReviewInput` Literal widened (not `pending` — review resolves, it doesn't un-resolve).
- **Hard-block:** background turns (heartbeat/subtask, `is_background=True`) cannot self-resolve — threaded via `ToolDispatcher.dispatch`. Code-level guard, not a censor (agent-created censors cap at `refuse`-tier, per live Nous).

## Testing

- 9 new tests (`test_brain.py`, `test_tools.py`) covering noise/superseded review, `superseded_by` semantics, batch resilience, calibration exclusion, and the background hard-block.
- Green on **SQLite and real Postgres** (`NOUS_TEST_DB=postgres`).
- Migration 062 applied + verified on a live Postgres (constraint widened, column present, existing rows pass).
- `ruff` clean on changed lines; migrator splitter parses the COMMENT semicolon correctly.

## Deploy notes

- Migration auto-applies on startup. Tools go live with the default-on flag; set `NOUS_DECISION_RESOLUTION_ENABLED=false` to disable.
- Backfill of the existing pending pile is intentionally left to Nous's first live sweep with the new tools.

### DIFF
--- a/nous/api/runner.py
+++ b/nous/api/runner.py
@@ -98,9 +98,9 @@ def _build_exclude_ids(
 
 # Frame-gated tool access (D5)
 FRAME_TOOLS: dict[str, list[str]] = {
-    "conversation": ["record_decision", "learn_fact", "learn_skill", "recall_deep", "recall_recent", "recall_hubs", "get_procedure", "create_censor", "bash", "read_file", "write_file", "web_search", "web_fetch", "cache_retrieve", "spawn_task", "spawn_sync", "schedule_task", "list_tasks", "cancel_task", "run_python", "send_file", "send_email", "heartbeat_check_create", "heartbeat_check_manage", "dag_create", "dag_manage", "ingest_document"],
-    "question": ["recall_deep", "recall_recent", "recall_hubs", "get_procedure", "bash", "read_file", "write_file", "record_decision", "learn_fact", "learn_skill", "create_censor", "web_search", "web_fetch", "cache_retrieve", "list_tasks", "cancel_task", "run_python", "dag_manage", "ingest_document"],
-    "decision": ["record_decision", "recall_deep", "recall_recent", "recall_hubs", "get_procedure", "create_censor", "bash", "read_file", "web_search", "web_fetch", "cache_retrieve", "list_tasks", "cancel_task", "dag_manage"],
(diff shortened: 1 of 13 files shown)


tfatykhov:
@codex review — focus areas: (1) migration 062 idempotency + that the new CHECK is a true superset; (2) calibration denominator change (noise/superseded excluded, not scored as failure); (3) the is_background hard-block wiring through ToolDispatcher.dispatch (heartbeat/subtask path only, foreground stream path unaffected); (4) review_many per-item failure handling within one transaction.

tfatykhov:
@codex review — re-review after 6b65dc6 (test-only: runner mock dispatchers now accept the new is_background kwarg). No production changes since your clean review of 29e5489.

PR #546, opened 2026-06-27
F035.6 Consolidation Audit Diff — per-sleep-cycle changelog + dashboard

tfatykhov:
## Summary
Implements **F035.6 Consolidation Audit Diff** — a reviewable, per-sleep-cycle changelog of every memory mutation, persisted to the database for historical/drift analysis, with a Svelte v2 dashboard view. Behind a default-off kill-switch (`NOUS_CONSOLIDATION_AUDIT_ENABLED`), so merged-but-unflipped is byte-for-byte unchanged sleep.

Spec hardened + corrected in #545 (this PR is the code; #545 is the spec).

## What lands (Full Phase 1a + 1b)
- **migration 063** — `nous_system.consolidation_cycles` (envelope) + `consolidation_actions` (per-mutation). Nullable `cycle_id` (`ON DELETE SET NULL`) so orphan rows survive an envelope-write failure; `trace_id varchar(12)` causal link.
- **`ConsolidationAuditor`** — cycle_id + in-process trace_id, per-phase buffered records, batched `executemany ... ON CONFLICT DO NOTHING` flush as tracked tasks, **drain-before-close** (A9), **orphan degradation** on open failure (A8), retention sweep. Every DB failure suppressed — a broken audit can never break a sleep cycle.
- **sleep_handler wiring** — envelope opened at `_run_sleep` start (gated), closed + retention swept in `finally`. Fact-mutation phases (reflect, stale_scan, F031, F027) record **per-action** with before/after; graph/episode phases record **one per-phase summary** (they only return aggregate counts — per-edge rows are a scoped follow-up). Auditor stored on `self` (matches the `_graph_densifier` idiom) instead of threading a `cycle_id` param through 11 signatures.
- **dashboard** — `GET /dashboard/consolidation` + `/{cycle_id}`, `dashboard_queries` functions, new **Consolidation** Svelte view + nav (cycles table → phase-grouped diff).

## Invariant correction (from the #545 review)
`totals` is derived from *attempted* mutation counts, so the integrity invariant is **`actions_persisted <= totals`** (equality only on a fully-successful cycle), not `totals == count(actions)`. `_phase_reflect` is correctly in audit scope (it `learn`s + supersedes).

## Verification
- 12 new unit tests (envelope, totals invariant, orphan degradation, drain ordering, retention disable, backpressure, kill-switch on/off) — green.
- Existing sleep/F031/F027 suites green (158 tests).
- Migration applied on a fresh schema; `ConsolidationAuditor` + both dashboard queries smoke-tested end-to-end against real Postgres (jsonb cast, array params, FK, before/after round-trip).
- `dashboard-app` `npm run check`: 0 errors.

## Rollout
Default OFF. Flip `NOUS_CONSOLIDATION_AUDIT_ENABLED=true` to start persisting; `NOUS_CONSOLIDATION_AUDIT_RETENTION_DAYS` (default 30) bounds the action table (per-night totals kept indefinitely as the F035.3 drift series).


tfatykhov:
@codex review

tfatykhov:
@codex review — addressed all three P2s in 6f1ae25:
- **stale_scan**: audit rows now buffered and emitted only after `session.commit()` succeeds (no phantom rows on rollback).
- **F031**: contradiction action recorded only when `contradictions_resolved` actually advanced (post-apply), not before the apply try-block.
- **ordering**: added a monotonic per-cycle `seq` column; dashboard detail now orders by `seq ASC NULLS LAST, created_at`. Validated end-to-end on real Postgres.

Also fixed the CI `test_all_tables_exist` failure (added the two new tables to the expected set; verified with NOUS_TEST_DB=postgres).

tfatykhov:
@codex review — fixed in the latest commit: the auditor now uses the configured `heart.agent_id` (== `settings.agent_id`, the dashboard's filter) instead of `event.agent_id`, so scheduler-emitted sleeps (`agent_id="system"`) are no longer hidden from `/dashboard/consolidation`. `parent_trace_id` still links to the triggering event.

tfatykhov:
@codex review — fixed: `_apply_supersede()` now returns whether it actually committed; the SUPERSEDE_A/B branches gate `contradictions_resolved` + the audit row on that, so a raced no-op (loser already superseded) no longer reports a supersession. Also corrects a pre-existing F031 over-count. 114 tests green.

tfatykhov:
@codex review — round 4 addressed:
- **P1 migration**: removed semicolons from inline `--` comments in 063 (the migrator splits on `;` after stripping only full-line comments). Verified `_split_sql_statements()` now yields 7 clean statements.
- **P2 partial commit**: `_run_audited_phase` records the summary regardless of the phase's final success flag (committed mutations always appear).
- **P2 diagnostic counters**: replaced the generic int-delta with an explicit per-phase mutation-counter allowlist; `*_error`/`*_skipped_*` no longer create phantom rows; bool flags record only on False→True. 117 tests green.

tfatykhov:
@codex review

PR #547, opened 2026-06-28
F035.6: surface F044 STC promotion + reinforcement in audit summary

tfatykhov:
## Summary
Follow-up to #546. The `stc_consolidate` (F044 tinyHippo-Lite) audit summary only fired on `f044_downscaled`, so a sleep cycle that **promoted** edges (`tagged→consolidated`) or flushed **recall-touch** `ltp_count` writes — but did no downscale — produced no STC action row in the consolidation diff. Observed live in the first prod cycle (`5773fa93`): `f044_promoted` and reinforcement activity rode in `totals` but weren't a changelog action.

## Change
Widen the `stc_consolidate` mutation-counter allowlist to the three real per-cycle UPDATE rowcounts:
- `f044_promoted` — `consolidation_state` tagged→consolidated
- `f044_recall_touches_flushed` — `ltp_count` writes flushed from the recall buffer
- `f044_downscaled` — weight downscale (unchanged)

## What stays excluded (and why)
`f044_n_edges` / `f044_n_tagged` / `f044_n_consolidated` / `f044_ltp_ge1..3` / `f044_reinforced_24h` are **state/24h-window snapshots**, not this-cycle mutations (verified against `tinyhippo_lite.py`: the telemetry `SELECT count(*) FILTER ...` vs the `PROMOTE`/`LTP_INCREMENT`/`DOWNSCALE` `UPDATE`s). Since `_run_audited_phase` records the delta-from-0 each cycle, including a 15k-edge snapshot would record a bogus 15k "mutation" — exactly the phantom-row trap codex flagged in #546 round 4.

## Tests
New `test_stc_phase_records_only_f044_mutations_not_telemetry` feeds a realistic mixed stats dict and asserts only `{f044_promoted, f044_recall_touches_flushed}` are recorded. 62 audit+sleep tests green.

### DIFF
--- a/nous/brain/tinyhippo_lite.py
+++ b/nous/brain/tinyhippo_lite.py
@@ -18,10 +18,26 @@ downstream consumer reads yet) and reports counts.
 
 from __future__ import annotations
 
+import logging
 from collections import Counter
 from typing import Any
 
 from sqlalchemy import text
+
+logger = logging.getLogger(__name__)
+
+# Keys produced by ``stc_promote_and_measure``; used to zero-fill a partial
+# result when promotion/telemetry fails after the (durable) recall flush.
+_STC_PROMOTE_KEYS = (
+    "f044_promoted",
+    "f044_n_edges",
+    "f044_n_tagged",
+    "f044_n_consolidated",
+    "f044_ltp_ge1",
+    "f044_ltp_ge2",
+    "f044_ltp_ge3",
+    "f044_reinforced_24h",
+)
 
 # F044 v1.1 — buffered recall-touch reinforcement. Edges among co-retrieved
 # results are "reactivated" by a recall (retrieval == STC reactivation). We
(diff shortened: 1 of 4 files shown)


tfatykhov:
@codex review

tfatykhov:
@codex review — fixed: `run_stc_consolidation` now captures the durable `f044_recall_touches_flushed` count before the promotion/telemetry step, and on a post-flush failure returns a full-shape stats dict (flush count + zero-filled rolled-back promotion keys) with a logged warning instead of propagating — so the committed reinforcement writes always reach the audit summary. Promotion stays idempotent. 2 new unit tests; 16 tinyhippo tests green on real PG.

tfatykhov:
@codex review — round 2 addressed: `run_stc_consolidation` now sets an `f044_stc_error` sentinel on the post-flush failure path, and `_phase_stc_consolidation` returns `False` when set — so `phases_completed` no longer claims `stc_consolidation` succeeded. The durable flush count still reaches `sleep_stats` and the audit (`_run_audited_phase` records regardless of phase return). Sentinel is a diagnostic `*_error` key, excluded from the audit allowlist. 78 tests green.

tfatykhov:
@codex review — round 3 addressed: promotion stats are now held in a local and merged into the returned dict ONLY after `session.commit()` succeeds, so a commit-time rollback (serialization/deadlock) zero-fills `f044_promoted` instead of auditing an uncommitted promotion. Added a promote-ok/commit-raises test. 33 tests green.

tfatykhov:
@codex review — round 4 addressed: `flush_recall_touches()` now returns the summed `UPDATE rowcount` (edges actually reinforced), not `len(items)` — so a buffered touch whose edge was deleted between recall and sleep is excluded from the audited `f044_recall_touches_flushed` count. New postgres test covers the deleted-edge case. 34 tests green.

tfatykhov:
@codex review — round 5 addressed: `f044_stc_error` is now a **bool** (`True`) rather than `1`, so it isn't summed by `_sum_totals` into `_attempted` (which counts non-bool ints). The phase's truthiness check is unchanged; tests assert `is True`. 80 tests green.

tfatykhov:
@codex review — round 6 addressed: the post-flush failure path now zero-fills only the committed-mutation counter (`f044_promoted=0`) and OMITS the telemetry snapshots (`f044_n_edges`/`n_tagged`/`ltp_ge*`/`reinforced_24h`) rather than fabricating zeros — so a failed cycle no longer reports the graph as empty in the drift totals. The phase returns False before its telemetry `logger.info`, so omission can't KeyError. Tests assert telemetry absent on both failure paths. 34 tests green.

PR #548, opened 2026-06-29
fix(dashboard): force detoast before LEFT() so compressed chunk labels don't 500

tfatykhov:
## Problem

`GET /dashboard/graph` returns **500** (`asyncpg CharacterNotInRepertoireError`, SQLSTATE 22021: `invalid byte sequence for encoding "UTF8": 0xe2 0x80`) whenever a graph node's text label comes from a **pglz-compressed** column.

```
nous-agent | ERROR Dashboard graph error: ... invalid byte sequence for encoding "UTF8": 0xe2 0x80
nous-agent | [SQL: SELECT id::text, LEFT(content, 120) AS label, chunk_index::text AS category, created_at
nous-agent |       FROM heart.episode_chunks WHERE agent_id = $1 AND id = ANY($2)]
```

## Root cause

The data is **valid UTF-8** — `length()`, full-value reads, and a strict Python decode all succeed. The crash only fires when `content` is stored **pglz-compressed**: `heart.episode_chunks` rows carry a 1536-dim embedding (~6KB) that pushes the tuple over the TOAST threshold, so PG compresses the raw-transcript `content`. A length-limited `LEFT()`/`substring()` then **slice-detoasts** the compressed value and mis-walks a multibyte UTF-8 boundary, emitting a 2-byte fragment of a 3-byte char (an em-dash, `e2 80 94`). The whole batch query aborts → 500.

Confirmed on prod (read-only): **36 of 2079** chunks are compressed; one (a PDF/document chunk) tripped it. The same bytes work fine in a local *uncompressed* table and as a bare expression — only the compressed column read fails (`pg_column_compression(content)='pglz'`).

Full-value reads (the node body, `recall_deep` chunk hydration) detoast fully and are **unaffected** — only the truncated-label path is at risk. There is **no corrupt data to repair**: re-storing just re-compresses and re-triggers.

## Fix

New `_safe_left(expr, n)` → `LEFT((expr) || '', n)`. The `|| ''` forces a full in-memory detoast first, so `LEFT` truncates the decompressed value. Applied to every label truncation in `get_graph_data` (decision, fact, episode, procedure, chunk) and the `get_node_detail` neighbor-label query — all of those rows carry embeddings and share the latent bug.

## Tests

- New `TestCompressedChunkLabel` reproduces the **exact** prod content (compressed via a real embedding); it fails on the old code with the identical error and passes on the fix. Includes a precondition test asserting the row actually compresses (`pglz`) so the others can't pass vacuously.
- `34` dashboard + `16` rest/density tests green.
- Re-verified read-only on real prod: the fixed query processes all 2079 chunks and yields a clean label for the formerly-failing row.

## Deploy note

The live container keeps 500-ing until this is deployed — because the bug is in the query path, **no data write can clear it live**.

### DIFF
--- a/nous/api/dashboard_queries.py
+++ b/nous/api/dashboard_queries.py
@@ -36,6 +36,20 @@ def _active_clause(table: str, alias: str = "t") -> str:
     return f" AND {alias}.active = true" if table in _ACTIVE_FILTER_TABLES else ""
 
 
+def _safe_left(expr: str, n: int = 120) -> str:
+    """`LEFT(expr, n)` that survives pglz-COMPRESSED multibyte text.
+
+    A length-limited LEFT()/substring() slice-detoasts a *compressed* text value
+    and can mis-walk a multibyte UTF-8 boundary, raising SQLSTATE 22021
+    ("invalid byte sequence ..."). This bit GET /dashboard/graph on a chunk whose
+    raw-transcript `content` compressed (its 6KB embedding pushed the tuple over
+    the TOAST threshold). Concatenating `''` forces a full in-memory detoast
+    first, so LEFT then truncates the decompressed value. Full-value reads (the
+    node body, recall hydration) are unaffected and need no wrapping.
+    """
+    return f"LEFT(({expr}) || '', {n})"
+
+
 # ── Task 5: Dashboard stats (used by GET /status?dashboard=true) ─────────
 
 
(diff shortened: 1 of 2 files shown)


tfatykhov:
@codex please review this fix. Focus areas: (1) the `_safe_left` / `LEFT((expr) || '', n)` detoast workaround for the pglz-compressed slice-detoast bug — is the SQL correct/safe for NULL and for the COALESCE(title, summary) expr; (2) whether all label-truncation sites are covered in get_graph_data + get_node_detail; (3) the regression test's reliance on pg_column_compression='pglz' as a precondition.

PR #556, opened 2026-07-11
feat(retrieval)!: heart fact seeds for spreading activation (default behavior)

tfatykhov:
## What

`NOUS_SPREADING_HEART_SEEDS_ENABLED` (default **false**, land-dark): when on, the spreading-activation CTE is seeded with the **top-3 heart FACT results** (with their RRF scores) alongside the existing decision seeds, and the Stage-4 gate becomes `graph_recall_enabled AND (decision_results OR heart_fact_seeds)` — so spreading can fire on decision-less corpora.

Also (flag-gated, same reason): spreading results are deduped against ids already in the heart/chunk candidate set, since fact-seeded traversal constantly re-reaches them. Flag off = byte-identical wiring (pinned by test).

## Why

- Owner directive (2026-07-11): decision-only seeding loses most of the graph — the fact/chunk side is where prod's edges are (11.5k inferred + 5.1k deterministic `related_to`, 2.5k `summarized_by`), while decision seeds only exist when a recall happens to hit `brain.decisions`.
- MemoryAgentBench corpora carry **~0 decisions** (census of 25 MAB run agents), so stock spreading is structurally invisible to our end-to-end benchmark. Heart seeds make the mechanism measurable there.
- Seed scores stay coherent: fact RRF [0,1] and decision RRF-or-0.5 share the same normalizer family; cap of 3 mirrors Path A's heart-seed window.

## Evaluation plan (this PR is the eval vehicle — flag stays off until gates pass)

Three-arm F051 retrieval A/B on the prod-shape eval corpus (`spread_force_off` / `spread_force_on` / **`spread_heart_on`** — the new config ships in this PR) on graph-targeted + probes + hand-label qrels, plus a MemoryAgentBench slice with `configs/spread_heart_on.env`. Separates the mechanism's lift from the seeding change. Known caveat measured as-is: spreading's SUM/no-cycle activation scoring (simplification-plan item 1.2) is untouched here.

## Tests

4 mock-based tests (TDD, watched fail first): fact seeds fire without decisions (seed tuple wiring asserted), flag-off decision-only wiring, combined seeding, candidate-dedup. Full pipeline + resolver suites green (37).


tfatykhov:
@codex review

tfatykhov:
@codex review — addressed the P2: the flag-gated dedup set now includes prior graph-stage outputs (acc.heart_graph_decisions + acc.heart_graph_memory_neighbors) alongside direct heart/chunk ids, so a Stage-2/Path-A node cannot be re-appended by spreading in the decision-less path. Regression test (test_flag_on_spread_hit_already_in_graph_stage_not_duplicated, watched fail first).

tfatykhov:
@codex review — REWORK per owner directive (no flag, no dark landing): NOUS_SPREADING_HEART_SEEDS_ENABLED removed entirely; top-3 heart fact seeding, the decision-less Stage-4 gate, and the candidate/graph-stage dedup are now unconditional default spreading behavior. Evidence: MAB paired A/B (accurate_retrieval, n=40/arm, spread_off vs heart-seeded ON at this PR's HEAD): 0 memory regressions, +2.5pp accuracy, identical +25pp memory lift. Eval config spread_heart_on deleted (spread_force_on now subsumes it); CLAUDE.md row folded into the spreading row. 48 tests green.

tfatykhov:
@codex review — addressed round-2 P2: known duplicates (seeds, decision ids, heart/chunk candidates, prior graph-stage outputs) are now excluded INSIDE the CTE's final SELECT via exclude_ids (id != ALL(uuid[])), before the LIMIT, so they never consume the result window; traversal still passes through excluded nodes. Python-side guards retained as a belt. Wiring regression test (test_candidate_exclusions_pushed_into_cte, watched fail first).

tfatykhov:
@codex review — addressed round-3 P2: _graph_expanded_to_pipeline now tags stage_origin by node_type — decisions keep 'brain_graph' (=== Brain Decisions ===), while fact/episode/chunk/procedure spreading hits get 'heart_graph_memory' so the formatter interleaves them into the typed === Heart Memory === section with the [via spreading_activation] provenance tag, exactly like Path A neighbors. Regression test (test_non_decision_spread_hits_route_to_heart_memory, watched fail first: fact hit was tagged brain_graph).

tfatykhov:
@codex review — addressed round-4 P2: heart_fact_seeds construction and the whole Stage-4 graph-expansion block moved OUT of the decision-search gate, so Heart-only scopes (memory_types=['fact'] etc.) seed spreading; decision query still only runs under decision/all scope, and the 1-hop fallback no-ops on an empty decision list. Regression test (test_fact_scoped_recall_seeds_spreading: brain.query not awaited, fact seeds passed, spread hit surfaced; watched fail first).

PR #559, opened 2026-07-13
feat(context): reliable pre-turn fact injection — render depth, fact pin, supersession lineage, recall backstop (all land-dark)

tfatykhov:
## What

Five flag-gated fixes to the pre-turn ("Block-2") fact-injection path, closing the counterfactual-fact failure chain from the 6k LongMemEval investigation: a stored fact that `recall_deep` finds in one hop was missing from the pre-turn system prompt, leaving correctness to a stochastic tool-call election. **Every flag defaults to byte-identical current behavior** — a build()-level golden captured on unmodified HEAD before any implementation enforces this in CI.

| Flag | Default | Change |
|------|---------|--------|
| `NOUS_FACT_FORMAT_MAX_CHARS` | `200` | Per-fact render cap in pre-turn context (was hardcoded 200 — facts rendered as near-meaningless stubs) |
| `NOUS_FACT_FORMAT_FULL_TOP_N` | `0` | Render top-N Relevant Facts untruncated |
| `NOUS_FACT_PIN_TOP_K` | `0` | Holdout-and-reinsert pin: top-K post-recency-resolve search hits are guaranteed into the injected list past diversity/dedup/relevance demotion. Captured AFTER `_resolve_recency`; superseded-tagged facts never re-inserted (can't resurrect the stale value). Pinned IDs land in `recalled_fact_ids` (F071 contract). |
| `NOUS_SUPERSESSION_LINEAGE_MODE` | `off` | `tag` / `named` lineage annotation on injected facts that supersede an earlier fact. Reads the authoritative `superseded_by` column (agent-scoped, includes inactive rows). Threaded as a dict into `_format_facts` — never attribute mutation (`_ScoredWrapper.__slots__` forbids writes). `named` quotes the stale value (anchoring risk — A/B before any prod use). |
| `NOUS_RECALL_BACKSTOP_ENABLED` | `false` | When the FINAL fact list is empty (search failed / returned nothing / pipeline emptied it), inject a "call recall_deep before answering" instruction. Deterministic empty-set trigger, no score thresholds. |

Plus `scripts/diag/probe_preturn_fact_gate.py`: a read-only diagnostic that replays the REAL pre-turn pipeline — including the intent-plan query rewrite (keyword-bag) and plan limit that `recall_deep` does not apply — and attributes exactly which gate drops a gold fact. Smoke-validated live against `nous_eval_prod` (gold rank 1, survives all gates; requires `NOUS_EMBEDDING_MODEL` matching the corpus — see README).

## Why this shape

Code verification showed pre-turn and `recall_deep` share the same `hybrid_search` fact leg — the miss comes from pre-turn's *gates* (intent-plan query rewrite + limit, diversity, dedup, relevance gap-cut, budget/render truncation; staleness is a phantom for facts since `FactSummary` has no `created_at`). The pin bypasses all demotion gates at a single touchpoint; the render fix makes surviving facts actually readable; lineage names the supersession so the model can override its parametric prior; the backstop converts the residual "no facts injected" case into a deterministic recall instruction.

## Process

- Plan (`docs/plans/2026-07-13-preturn-fact-injection.md`) reviewed by a 3-agent team (architecture / phantom-API / test-eval) — 5 P1-class flaws caught and fixed **before** implementation (incl. lineage-by-mutation that would have shipped silently dead on `_ScoredWrapper`, and a pin design that could resurrect recency-demoted stale facts); re-review verdict CLEAN.
- Subagent-driven implementation, per-task spec+quality review on every task, final whole-branch review (all findings triaged; verdict READY TO MERGE).
- 27 new tests. Full-suite failure set diffed against main-equivalent baseline: identical except the new real-PG test which matches its `test_facts.py` siblings' convention (passes under `NOUS_TEST_DB=postgres`).

## Validation & flag flips (post-merge gates — nothing flips in this PR)

- **Probe diagnostic run** on the 6k counterfactual question set: pending the failing-questions file from that investigation. Two conditional remedies (`fact_fetch_limit_override`, `fact_query_use_raw_input`) were deliberately NOT built pending that attribution.
- **A/B gate before any flag flip**: `qa_context_ab.py` harness (local `eval/h2-injection-precision` branch — real `ContextEngine.build` path, Opus generator, LLM-judged accuracy), arms baseline / L1(pin=3+full_top_n=3) / L1+tag / L1+named / L1+backstop, one variable per step; counterfactual subset (~12) directional + n≥100 prod-shape non-regression. Losing arms' flags get deleted, not left dark.

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -571,6 +571,11 @@ DB connection vars are **unprefixed** (shared with docker-compose). All others u
 | `NOUS_RECALL_BEFORE_CLARIFY_PROMPT` | `true` | F083 C2. When true, inject a static system-prompt instruction telling the agent to call `recall_deep`/`recall_recent` to resolve a referent before asking the user to clarify. |
 | `NOUS_FOLLOWUP_FIRST_TURN_EPISODE` | `false` | F083 A2 (**land-dark**, pending local A/B). When true, on a verified first turn of a new session the temporal tier injects the most-recent episode's FULL `structured_summary.summary` (+ `open_threads`, truncated to `recall_parent_episode_truncate`) instead of titles-only. First-turn signal = `session_id not in _active_episodes` (survives LRU eviction). |
 | `NOUS_EPISODE_OPEN_THREADS` | `false` | F083 B (**land-dark**, pending local A/B). When true, the episode summarizer extracts a top-level `open_threads` array (unfinished items / next steps) into `structured_summary`; raises summary `max_tokens` to 3000 so the extra field doesn't truncate the JSON. Consumed by A2. |
+| `NOUS_FACT_FORMAT_MAX_CHARS` | `200` | Per-fact char cap in pre-turn context rendering (was hardcoded 200). Shared by Relevant Facts AND User Profile sections. |
+| `NOUS_FACT_FORMAT_FULL_TOP_N` | `0` | Render the top-N Relevant Facts untruncated (0 = all capped). |
+| `NOUS_FACT_PIN_TOP_K` | `0` | Pin top-K post-recency-resolve fact hits into pre-turn context past diversity/dedup/relevance demotion (0 = off; superseded-tagged facts never pinned). Counterfactual-injection fix, 2026-07-13 plan; flip gated on prod-generator A/B. |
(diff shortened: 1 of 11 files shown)


tfatykhov:
@codex review

PR #557, opened 2026-07-12
fix(sleep): F053 prune erased the episode graph layer - liveness predicate + restore

tfatykhov:
## Problem — F053's prune erases the entire episode graph layer

`heart.episodes.active` is semantically overloaded: on facts/procedures `active=false` means soft-deleted, but on episodes it is the **normal closed lifecycle state** (008.3 — `Episode._end()` flips it on every session close). The HT-1 audit (2026-06-09) hit this trap in episode search and fixed only search. F053's dead-edge prune (designed for F031/F027-deactivated *facts*) inherited `active=false` as its episode dead-node predicate — so it deletes the edges of every normally-closed episode, nightly. F040's backfill (`t.active = true` orphan filter) can then never rebuild them, and the orphan gate makes the loss permanent (a chunk keeps its chunk↔chunk edges → never re-orphans → its pruned `part_of` anchor is never recreated).

### Prod evidence (read-only audit, 2026-07-12)

| Measurement | Value |
|---|---|
| Episode census | 756 total = 96 ongoing + **657 closed-normal** + 3 trivial discards + 10 abandoned |
| Edges held by the 657 closed episodes | **6** |
| Episode-incident edges on the 96 ongoing episodes (pruned when they close) | 348 / 354 |
| chunk→episode `part_of` anchors | **3 / 3,060 chunks** — the 3 survivors are exactly the 3 chunks whose parent episodes are still active |
| episode→decision `discussed_in` | 1 |
| Genuine soft-deletes in the inactive population | 13/670 (2%) — the rest are normal closes |

Every graph consumer (Path A Stage 2b, adjacency boost, dashboard graph, any future episode-graph experiment) sees closed episodes as disconnected. Retrieval was shielded (HT-1 fixed search on `ended_at`; facts keep the `source_episode_id` FK), which is why this went unnoticed.

## Fix

Episode liveness centralized in `graph_constants.py` (mirrors HT-1's search predicate):

- **DEAD** (prunable): `(active = false AND ended_at IS NULL) OR outcome = 'abandoned'` — trivial discards + F060.2 abandoned marks only. Writer census verified no valuable episode lands here (abandoned is terminal; `_end()` always sets `ended_at`).
- **LIVE**: `(active = true OR ended_at IS NOT NULL) AND outcome IS DISTINCT FROM 'abandoned'`.

Four code changes:

1. **F053 prune** (`sleep_handler.py`): episode branch of the `inactive_nodes` CTE uses the DEAD predicate — closed episodes' edges survive.
[...]

PR #561, opened 2026-07-14
feat: F084 write-path adjudication — enumerative extraction + store-time supersession (land-dark)

tfatykhov:
## Summary

Fixes the two measured MAB write-path failures: (1) enumerable content lossy-compressed by the ~150-word episode summary before fact extraction (1% of answerable facts stored), and (2) update-chains stored as unresolved variants (19 superseded_by rows vs hundreds of chains) forcing the answering model to re-adjudicate conflicts per question.

**R1 — Enumerative extraction (modal, density-adaptive):** a heuristic classifies transcripts; enumerable ones route fact storage through a new chunked atomic-fact extraction leg (raw text, not the summary) producing normalized `subject_key`/`attribute_key` + positional `source_ordinal`. Narrative episodes byte-identical.

**R2 — Store-time supersession:** indexed same-key conflict lookup at write time + capped sleep sweep + backfill, all confirming via the existing F027 classifier and resolving through a shared `apply_supersession` primitive. F075 date-bypass takes precedence over ordinal; CONTRADICTION resolved only by the classifier's `current_fact` (never ordinal); fail-open KEEP-BOTH. Cycle guard in `get_current` (latest-learned wins, break persisted).

**R2.4 — decoupled arm:** `overrides_prior` marker rendering (`[memory override — trust this over general knowledge]`).

**Backfills:** `scripts/backfill_enumerative_facts.py` + `scripts/backfill_supersession.py` (#557 conventions: dry-run, rollback watermark, idempotent, budget overrides for clone remediation).

**Entirely land-dark:** 13 new flags, all default OFF; golden flags-off tests pin byte-identical behavior. Migration `064` (nullable columns + partial index, idempotent).

## Process

Requirements (MAB eval program) → 3-reviewer validation → plan → 3-reviewer plan validation → 14 TDD tasks each with implementer + spec/quality reviewer + fix loops → final whole-branch review (4 findings, all fixed). Plan: `docs/plans/2026-07-13-write-path-adjudication.md`; spec: `docs/features/F084-write-path-adjudication.md`.

## Test plan

- 54 new tests in `tests/test_write_path_adjudication.py` (11-row R2 contract table, golden flags-off, PG-lane pinning, cycle-guard durability, backfill dry-run zero-writes/termination)
- Full suite: 5,039 passed; 6 failures are pre-existing at the merge-base (test_admission_dashboard ×2, test_spreading_activation ×4 — local-env, also fail on untouched main)

## Rollout (per F084 doc)

Acceptance is external (MAB harness on a backfilled clone): R1 coverage probe is a gate, the stage-2 paired CR n=320 QA replay is the decisive metric; conversation-shaped prod-generator A/B + 5k-fact ingest wall-clock measurement required before any prod flag flip. Rollout order: R2.4 marker → R1 → R2.

## Known follow-ups (from final review, non-blocking)

- Sweep tie blind spot: equal-timestamp same-key pairs invisible to the sweep join (`(learned_at, id)` tiebreaker)
- Verdict-gate duplication between `_resolve_key_conflicts` and `resolve_key_conflict_pair` (extract shared `_adjudicate_pair`)
- Per-fact `float(confidence)` conversion can abort an enumerative episode on malformed LLM output (wrap per-fact)
- SQLite test shims: `FactSummary.overrides_prior` + cycle repair divergence (PG-only markers cover today)
- Combo test: override marker + lineage annotation on one fact


tfatykhov:
@codex review

Both round-1 P2s addressed:
- Long-chain rewrite: `_get_current` now uses path-array cycle detection (`NOT f.id = ANY(c.path)`, depth<100 backstop only) — legitimate chains of any length reach their NULL tip; repair fires only on genuine cycles (296159f, +12-link-chain regression test).
- Sweep starvation: `find_key_conflict_pairs` gained an `after` paging cursor; the sleep phase advances it past keep-both pairs and resets on table exhaustion (28bedf5, +3-cycle paging test).

Also reverted the conftest migration replay that broke CI (3342d78). CI green.

tfatykhov:
@codex review

Round-2 P2s addressed:
- Depth-exhaustion vs cycle: the fallback now computes `is_cycle = (superseded_by = ANY(path))` on the deepest visited row in SQL — genuine cycles keep the repair; deep healthy chains (>100) log a WARNING and return the deepest visited row with ZERO mutation (8f10a9b, +120-link no-repair regression test).
- Sweep cursor: paged by the full 4-tuple `(f1.learned_at, f1.id, f2.learned_at, f2.id)` with matching ORDER BY and row-comparison predicate — same-f1 pairs adjudicate in order across cycles (263e690, rewritten 1×3-partner starvation test).

tfatykhov:
@codex review

Round-3 P2s addressed:
- Equal-timestamp pairs: join is now `(f1.learned_at, f1.id) < (f2.learned_at, f2.id)` in BOTH SQL branches (cursor + base) — id tiebreak, deterministic (e8c92ef, +same-timestamp pair test).
- Interrupted sweep: cursor advances only past the last pair actually processed; exhaustion-reset fires only on a completed page (2ab52e9, +mid-page interruption test).
- Zero-stored enumerative leg: early-return only when stored_ids is non-empty; empty result logs a WARNING and falls through to the legacy candidate path — variant-safe since zero enumerative facts means no variant pairs can form (cd56893, +empty-result fallthrough test).

tfatykhov:
@codex review

Round-4 P1 addressed (c9bfafc): `resolve_key_conflict_pair` now refetches both rows BEFORE the classifier and skips the pair (return False, no Haiku spend) unless both are active and unsuperseded — a page's earlier resolution can no longer produce an inactive winner or a cycle. Two new PG tests: stale-pair skip (sentinel proves zero classifier calls) and a full A/B/C sweep-page integration test asserting all chains transitively resolve to the active tip with no cycle-repair warnings.

tfatykhov:
@codex review

Round-5 P2s addressed:
- Rollback edge DELETE now agent-scoped in script header + F084 doc (edbdc6e).
- Live enumerative backfill fails fast without an embedding key — idempotency depends on cosine dedup (edbdc6e).
- `attribute_key` capped at 100 chars via `normalize_key(..., max_len=100)` to match VARCHAR(100) (88401bd).
- Legacy subject supersession restored for keyed facts whenever keyed resolution doesn't own them (R2 off or key missing): gate is now `not (R2_enabled and both keys)`; the remaining keyed-vs-unkeyed exact-match gap with R2 on is documented as F084 deviation #5 (359efff).
- `get_current` walks arbitrarily deep chains via iterative CTE restarts (Python-side visited set for cross-restart cycle detection, 100-restart bound raises) — the 120-link test now asserts the TRUE tip with zero mutation (9e2f29d).

tfatykhov:
@codex review

Round-6 P2 addressed (cd42fe0): the live path now lazily caches ONE `EnumerativeExtractor` on the `FactExtractor` instance, so `NOUS_ENUMERATIVE_EXTRACTION_MAX_PER_HOUR` is genuinely per-hour across episodes (test: cap=1, two enumerable episodes → exactly one extraction call, second episode falls back to the legacy path).

tfatykhov:
@codex review

Round-7 P2s addressed:
- Partial enumerative writes: per-chunk failures are now caught INSIDE `process_transcript` (log + stop, return partial ids) — a non-empty partial result skips the legacy fallback, so mixed keyed/unkeyed variant pairs cannot form; the wiring's except-branch is only reachable pre-store (942ca1f, +2 tests incl. wiring sentinel).
- Backfill throttling: new non-consuming `FactManager.key_budget_exhausted()` peek that rolls the hour bucket forward exactly like `_key_budget_ok` without consuming a slot; both raw `_key_calls >= cap` peeks in backfill_supersession.py replaced — throttled runs resume after the hour boundary (1ada15a, +rollover test).

tfatykhov:
@codex review

Round-8 P2 addressed (b7e784e): execution order in `resolve_key_conflict_pair` is now fetch → staleness guard → budget → classifier — stale pairs return False without consuming an hourly slot (extended the r4 stale-pair test to assert `_key_calls` unchanged).

tfatykhov:
@codex review

Round-9 P2 addressed (e0955ae): `_store_batch` isolates per-fact failures — on a mid-batch `Heart.learn` exception it logs and returns the ids already committed instead of propagating, so partially-persisted chunks are never invisible to the fallback guard (2 new tests: direct unit + wiring integration asserting the legacy path is not invoked).

tfatykhov:
@codex review

Round-10 P2s addressed:
- Overlap duplicates: `process_transcript` keeps an in-run normalized-content seen-set — a statement re-extracted from the adjacent chunk's overlap window is never stored twice, so the FIRST occurrence's ordinal wins (a global counter would still assign later ordinals to re-extractions; dropping the duplicate is the robust fix). Overlap retained for boundary coverage (ab4d24b, +first-ordinal-wins test).
- Malformed confidence: both keyed-resolution sites now use the existing F027 defensive idiom (`try/except (TypeError, ValueError) → 0.0`) — non-numeric verdict confidence fails open to KEEP-BOTH instead of aborting the learn transaction (5e0cb87, +2 tests).

tfatykhov:
@codex review

Round-11 P2s addressed:
- Dedup-confirm metadata: `_confirm_duplicate` now fills adjudication fields if-empty — keys copied as a pair only when both row keys are NULL, source_ordinal fill-if-NULL, overrides_prior=True sticky, existing values never overwritten (3e90ad6, +no-overwrite test).
- Lost-fact downstream scans: `_resolve_key_conflicts` returns bool; `_learn` skips the 0.85–0.95 contradiction scan AND domain-threshold check when the just-flushed fact was superseded by an existing winner — no contradicts edges or confidence penalties for already-resolved conflicts (581ceeb, +sentinel test both directions).

tfatykhov:
@codex review

Round-12 P2 addressed (8caca1f): the enumerative path now extracts `event_date` with the exact F075 producer conventions from `_store_candidate_facts` (flag-gated consumption, validator round-trip, classified_at stamped only when a raw date wasn't dropped — dropped stays backfill-eligible); dated same-key facts now hit the F075 different-dates guard instead of ordinal supersession. The in-run overlap seen-set keys on (content, event_date) so identical dated statements on different dates are preserved (4 new tests incl. flag-off golden).

tfatykhov:
@codex review

Round-13 P2s addressed — shared theme: skip-tracking never outruns actual processing:
- Sweep: budget-exhaustion is checked at loop top via the non-consuming peek and BREAKS without advancing `last_processed` — the cursor resumes at the first unadjudicated pair next cycle (ada7442).
- Backfill: budget check moved BEFORE the seen/pairs_examined update; budget-stopped pairs stay unseen, counted into `budget_stops`, and the run ends with an explicit deferred-N note — re-runs resume naturally (53a9003).
- Extractor: `_store_batch` now returns (input, id) pairs and only successfully-stored contents enter the overlap seen-set — a mid-batch learn failure no longer suppresses the statement's overlap retry copy (543054b).

tfatykhov:
@codex review

Round-14 P2 addressed (46ac722): `_store_batch` returns `(stored_pairs, store_failed)`; on a mid-batch failure `process_transcript` logs a WARNING naming the chunk, sets truncated=true (the loud end-of-run truncation warning fires), and stops the chunk loop rather than continuing against a failing backend. Partial stores remain visible to the fallback guard; no silent coverage loss.

tfatykhov:
@codex review

Round-15 P2s addressed:
- Failed vs empty chunks: `_extract_chunk` returns None only for helper failure, [] for a well-formed zero-fact response; failures log a WARNING, set truncated=true, and CONTINUE to later chunks (an LLM blip ≠ the r14 store-failure stop, contrast documented) (ea5d8d7, +2 tests).
- Long-cycle repair: when the cross-restart visited-set detects a cycle, the winner is selected over the FULL visited set (plain ANY(:ids) query, learned_at DESC with id tiebreak) instead of the 100-depth window — verified by a 150-node cycle test whose latest-learned node sits past the first window (ef98126).

tfatykhov:
@codex review

Round-16 P2 addressed (4f8f61f): the supersession backfill now wires the classifier with `settings.contradiction_model` — matching prod's F027 wiring at nous/main.py:212 and the docstring's Haiku budget math. (The enumerative backfill's `background_model` is intentional — it drives extraction, not the F027 classifier.)

PR #562, opened 2026-07-14
fix: enumerative backfill includes closed episodes — active is lifecycle, not liveness (#557 lesson)

tfatykhov:
## Summary

F084's `backfill_enumerative_facts.py` filtered episodes with `AND active = true` — but `episodes.active` is a **lifecycle** flag: `_end()` sets `active=False` on every properly closed episode (`nous/heart/episodes.py:235`). The backfill therefore selected only still-open episodes and silently skipped the entire closed corpus it exists to remediate (operator dry-runs returned 0 episodes). This is the #557 liveness lesson recurring.

**Fix:** `(active = true OR ended_at IS NOT NULL)` — open and closed episodes included; never-ended deactivated orphans stay excluded.

## Test plan

- New PG regression test `test_select_backfill_episodes_includes_closed_episodes`: closed (ended, active=False) episode INCLUDED, open episode included, never-ended deactivated orphan excluded — fails on the old filter, passes on the fix.
- Full `tests/test_write_path_adjudication.py`: 88 passed.

### DIFF
--- a/scripts/backfill_enumerative_facts.py
+++ b/scripts/backfill_enumerative_facts.py
@@ -84,8 +84,11 @@ async def select_backfill_episodes(session, agent_id: str, since, limit: int):
               -- _end() sets active=False on every properly CLOSED episode,
               -- which are exactly the transcripts this backfill remediates.
               -- Include open (active) and closed (ended_at set) episodes;
-              -- exclude only never-ended deactivated orphans.
+              -- exclude never-ended deactivated orphans and abandoned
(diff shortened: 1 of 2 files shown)


tfatykhov:
@codex review

Addressed: selection now adds `outcome IS DISTINCT FROM 'abandoned'`, mirroring the recall predicate at nous/heart/episodes.py:531 (F060.2 stamps ended_at on abandonment, so the liveness fix alone would have admitted those). Test extended with an abandoned episode asserting exclusion.

PR #575, opened 2026-07-25
fix(run_python): enforce the timeout for real, drop the useless builtins allowlist

tfatykhov:
## What

`run_python` blocked the harmless and permitted the harmful. This fixes both halves.

**P2 — the builtins allowlist is gone.** `SAFE_BUILTINS` (27 names) provided zero
security: the same agent holds an unrestricted `bash` tool in the same tool belt,
running full `python3` as a subprocess with the same env, credentials and
filesystem. There is no actor who has `run_python` but not `bash`. What the
allowlist did do was break harmless code — `import`, `try/except Exception`
(NameError on `Exception`), class definitions — costing the agent wasted turns.
The namespace now gets real `builtins`. The pre-injected convenience modules
(`json`, `re`, `math`, `datetime`, `collections`, `itertools`, `functools`,
`statistics`) stay, and the `_MAX_WRITES = 5` `learn_fact` cap and memory-function
wiring are untouched — that cap is memory hygiene, not sandboxing.

**P1 — the timeout is now real.** `asyncio.wait_for` cancels the *await*, not the
worker thread, and `executor.shutdown(cancel_futures=True)` only drops *queued*
futures. `while True: pass` needs zero builtins, sailed through the allowlist,
survived the timeout, held the GIL, and degraded the whole API — that exec runs on
a thread inside pid 1, alongside the event loop and the Postgres pool.

## P1: chose Option A (in-process kill switch)

Option B (subprocess isolation) would need an IPC bridge back to the parent's
asyncio DB pool for `recall_deep`/`learn_fact`/etc. — an RPC layer to reimplement
the one property that makes this tool worth having over `bash` (direct in-process
memory access). Nothing about the failure mode requires it: the observed hole is a
Python-level spin loop, which a trace hook does interrupt. So Option A, ~40 lines:

- `ScriptDeadlineExceeded(BaseException)` — deliberately **not** an `Exception`
  subclass, so now that `except Exception` works, agent code cannot swallow its
  own deadline.
- A `sys.settrace` hook installed inside the worker thread, checking
  `time.monotonic()` every 64 traced events (cheap; empirically fires within
  milliseconds of the deadline) and raising in the executing frame.
- `NOUS_PROGRAMMATIC_TOOLS_MAX_CONCURRENT` (default 4, `ge=1`). Over the cap,
  the call is rejected with a clear error and a WARNING log instead of stacking
  threads. `run_python_active_runs()` exposes the count; the per-run INFO log
  now includes `N/M slots in use`.
- The slot is released by the worker thread itself, so a thread that outlives its
  deadline keeps holding its slot — that is what bounds the blast radius.

### What the timeout now guarantees — and what it does not

**Guarantees:** any runaway that is executing Python bytecode — `while True: pass`,
runaway recursion, a pathological comprehension — is interrupted at the deadline,
the call returns `is_error` with a timeout message, and the worker thread exits.
Test-proven, including the case where the script wraps the loop in
`except Exception`.

**Does not guarantee:** a thread blocked *below* Python level cannot be
interrupted by any in-process mechanism — `time.sleep(3600)`, a blocking socket
read, a long C extension call. No trace event ever fires, so the hook never runs.
For those, the awaiting coroutine gives up after `timeout + 2s` and returns the
timeout error, but the thread lives until its call returns. That is precisely why
the concurrency cap exists: the leak is bounded at `max_concurrent` threads rather
than unbounded. A bare `except:` (catching `BaseException`) inside an infinite loop
can also re-swallow the deadline exception forever — an explicit, documented
residual, not something a subprocess design would fix any better than closing the
tool to the agent.

## Tests

`tests/test_run_python.py`: 60 passed in ~23s.

New coverage: `import base64` / `import os`; `try/except Exception`; class
definition; `getattr`; `open()` round-trip; **`while True: pass` stopped at the
deadline** (asserts timeout error + `run_python_active_runs() == 0` + thread count
back to baseline + elapsed under deadline+grace); deadline not swallowed by
`except Exception`; concurrency cap rejects when saturated and releases slots; two
config tests.

The `TestRunPythonBlockedBuiltins` class was replaced by `TestRunPythonFullBuiltins`
— those tests asserted the exact behaviour P2 removes (`import os` must error), so
they could not survive unchanged. Every other pre-existing test is untouched and
passes.

Lint: `ruff check nous/api/tools.py` reports 55 findings before and 55 after this
change — no new ones (all pre-existing E501/UP037 noise; CI lint is advisory).
`tests/test_run_python.py` is ruff-clean.

Also ran `tests/test_episode_id_injection.py tests/test_compaction.py
tests/test_execution_ledger.py tests/test_config.py`: 187 passed, 2 failed —
`test_config.py::test_date_leg_settings_defaults` and
`TestDAGTimeoutValidation::test_default_equals_max_is_ok`, both caused by ambient
`NOUS_*` env vars in this job's environment (e.g. `NOUS_DATE_LEG_ENABLED=true`)
leaking into bare `Settings()` construction. Unrelated to this change.

## Out of scope (noticed, not touched)

- `docs/implementation/012.3-programmatic-tool-calling.md` still quotes the
  `SAFE_BUILTINS` snippet. Left as-is — it is a historical build spec, not live docs.
- Those two `test_config.py` tests are environment-sensitive: they call bare
  `Settings()` and so assert defaults that any operator env var overrides. Worth
  hardening separately.

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -363,7 +363,8 @@ DB connection vars are **unprefixed** (shared with docker-compose). All others u
 | `NOUS_INLINE_SUBTASK_TIMEOUT` | `90` | Default timeout for inline (await_result) subtasks |
 | `NOUS_FRAME_DEFAULT_MODELS` | `{}` | JSON map of frame type to default model (falls back to `NOUS_BACKGROUND_MODEL`) |
 | `NOUS_PROGRAMMATIC_TOOLS_ENABLED` | `true` | Enable run_python tool for client-side code execution |
-| `NOUS_PROGRAMMATIC_TOOLS_TIMEOUT` | `10` | Timeout in seconds for run_python code execution |
+| `NOUS_PROGRAMMATIC_TOOLS_TIMEOUT` | `10` | Timeout in seconds for run_python code execution. Enforced in-thread by a `sys.settrace` deadline hook, so Python-level runaway code (`while True: pass`) is actually interrupted; a blocking C-level call still cannot be interrupted and holds its slot until it returns. |
+| `NOUS_PROGRAMMATIC_TOOLS_MAX_CONCURRENT` | `4` | Max concurrent run_python executions. Excess calls are rejected with an error instead of stacking threads inside the API process. |
 | `NOUS_CONTEXT_WINDOW` | auto | Override model context window size in tokens (0 = auto-detect from model name) |
 | `NOUS_ANTI_HALLUCINATION_PROMPT` | `true` | Inject "don't guess, re-fetch" safety prompt into system context |
 | `NOUS_TOOL_PRUNING_ENABLED` | `true` | Enable 4-tier tool result pruning pipeline |
--- a/nous/api/tools.py
+++ b/nous/api/tools.py
@@ -16,6 +16,8 @@ from __future__ import annotations
 import copy
 import json
 import logging
+import sys
+import threading
 import time
 from collections.abc import Callable
 from datetime import UTC, datetime, timedelta
@@ -3238,14 +3240,50 @@ def register_subtask_tools(
 # Programmatic tool calling (012.3)
 # ---------------------------------------------------------------------------
 
-SAFE_BUILTINS = {
-    "len", "list", "dict", "set", "str", "int", "float", "bool",
-    "print", "range", "enumerate", "zip", "sorted", "filter",
-    "map", "max", "min", "sum", "any", "all", "isinstance",
-    "repr", "round", "abs", "type", "tuple",
-}
-
 _MAX_WRITES = 5
+
+# Seconds the awaiting coroutine waits past the script deadline before giving
+# up on the worker thread. The in-thread trace hook normally raises at the
+# deadline, so this grace only matters when the thread is stuck below Python
+# level (a blocking C call), which no in-process mechanism can interrupt.
+_TIMEOUT_GRACE = 2.0
+
+# Trace-hook check interval — number of traced events between deadline checks.
+# Keeps the per-line overhead down without letting an overrun go unnoticed.
+_DEADLINE_CHECK_EVERY = 64
+
+_active_runs = 0
+_active_runs_lock = threading.Lock()
+
+
+class ScriptDeadlineExceeded(BaseException):
+    """Raised inside the run_python worker thread when its deadline passes.
+
+    Derives from BaseException, not Exception, so ordinary agent code using
+    `try/except Exception` cannot swallow its own timeout.
+    """
+
+
+def run_python_active_runs() -> int:
+    """Number of run_python executions currently occupying a worker thread."""
+    with _active_runs_lock:
+        return _active_runs
+
+
+def _acquire_run_slot(max_concurrent: int) -> bool:
+    """Take a concurrency slot if one is free."""
+    global _active_runs
+    with _active_runs_lock:
+        if _active_runs >= max_concurrent:
+            return False
+        _active_runs += 1
+        return True
+
+
+def _release_run_slot() -> None:
+    global _active_runs
+    with _active_runs_lock:
+        _active_runs -= 1
 
 
 def create_programmatic_tools(
@@ -3344,10 +3382,11 @@ def create_programmatic_tools(
         def _print(*args: object) -> None:
             output_buf.write(" ".join(str(a) for a in args) + "\n")
 
-        safe_builtins = {k: getattr(builtins, k) for k in SAFE_BUILTINS if hasattr(builtins, k)}
-
         namespace: dict[str, Any] = {
-            "__builtins__": safe_builtins,
+            # Full builtins: the allowlist that used to live here bought no
+            # security (the same agent holds an unrestricted `bash` tool) and
+            # broke harmless code — `import`, `except Exception`, class bodies.
+            "__builtins__": builtins,
             # Nous memory functions
             "recall_deep": _recall_deep,
             "recall_recent": _recall_recent,
@@ -3355,7 +3394,7 @@ def create_programmatic_tools(
             "learn_fact": _learn_fact,
             "print": _print,
             "result": None,
-            # Safe stdlib modules (pre-injected — import statement is disabled)
+            # Pre-injected for convenience — `import` also works.
             "json": json,
             "re": re,
             "math": math,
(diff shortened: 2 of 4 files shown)


tfatykhov:
## Nous Review — PR #575

**Status: CHANGES REQUESTED**

Main is green (PR #576 landed), CI is green on this branch, and the core design is sound — the `sys.settrace` deadline hook and the concurrency cap are the right architecture for in-process timeout enforcement. The bash-equivalence argument for dropping `SAFE_BUILTINS` is correct.

However Codex surfaced two P1s that block merge as-is.

---

### P1 — `sys.settrace(None)` disables the deadline hook [tools.py, `_run()`]

A script can call `import sys; sys.settrace(None)` before entering a spin loop. `sys.settrace` is per-thread in CPython — once cleared in the worker thread, the trace hook never fires, the deadline is never raised, and the thread holds its concurrency slot until the process restarts. Four such calls permanently exhaust `NOUS_PROGRAMMATIC_TOOLS_MAX_CONCURRENT`. Codex is right that this is a reliability bug independent of the bash/security argument.

**Fix:** Shadow `sys.settrace`/`sys.setprofile` in the exec namespace with no-ops before `exec()` so user code cannot uninstall the hook. The real `sys.settrace` is already called by the tracer setup code outside the exec call.

```python
# Inside _run(), after _tracer is installed:
_exec_ns["sys"] = types.SimpleNamespace(**{
    k: getattr(sys, k) for k in dir(sys)
    if not k.startswith("set")  # block settrace, setprofile, setrecursionlimit
})
```
Or more surgical: inject a shim `sys` that overrides just `settrace` and `setprofile` to no-ops while delegating everything else.

---

### P1 — `SystemExit` / `KeyboardInterrupt` escapes the tool boundary [tools.py]

`ScriptDeadlineExceeded` derives from `BaseException` intentionally (cannot be swallowed by `except Exception`). But that means a script calling `sys.exit()` or raising `SystemExit` / `KeyboardInterrupt` also produces a `BaseException` in the future that only `ScriptDeadlineExceeded` is explicitly caught for. The uncaught exception propagates past `ToolDispatcher.dispatch` and can terminate the event loop / crash the API process.

**Fix:** Wrap the future `.result()` call to also translate `SystemExit` and `KeyboardInterrupt` (and any other non-deadline `BaseException`) into an `is_error` response:

```python
except ScriptDeadlineExceeded:
    ...  # existing path
except BaseException as exc:  # SystemExit, KeyboardInterrupt, etc.
    return {"is_error": True, "content": [{"type": "text", "text": f"Error: script raised {type(exc).__name__}: {exc}"}]}
```

---

### P1 (Codex) — ActionGate bypass [deferred as out-of-scope]

Codex flagged that `urllib.request.urlopen()` / `subprocess.run()` calls from run_python bypass Tier-3 external-action gating. Agree in principle but **deferring**: the bash tool in the same tool belt already bypasses ActionGate for all external calls with zero friction. Reclassifying `run_python` as external-action adds per-call overhead with no net security improvement given that actor constraint. Filing as a separate issue instead.

---

### P3 — `time.sleep()` slot hold

The PR body correctly documents this residual: a blocking C call (including `time.sleep(3600)`) holds the slot until it returns. The concurrency cap bounds the blast radius. Worth a brief comment in the test that exercises `time.sleep(1)` to acknowledge this is the *expected* behaviour (slot held until sleep returns), not a latent bug.

---

### Verdict

Two fixes needed before merge (~30 min of work):
1. Shadow `sys.settrace`/`sys.setprofile` in the exec namespace
2. Translate non-deadline `BaseException` from the worker future into `is_error`

Everything else — the `sys.settrace` hook design, the concurrency cap, the `ScriptDeadlineExceeded(BaseException)` hierarchy, dropping `SAFE_BUILTINS`, the test coverage — is solid.

tfatykhov:
## PR #575 Review — fix(run_python): enforce the timeout for real, drop the useless builtins allowlist

**Status: CHANGES REQUESTED** — one P1 security issue remains; everything else is solid.

---

### ✅ What's well-done

- **P1 timeout design (Option A) is correct.** `_tracer` fires every 64 events, raises `ScriptDeadlineExceeded(BaseException)` when the deadline passes. `while True: pass` with zero builtins is actually killed — the core ask.
- **`_run()` finally block is correct.** `sys.settrace(None)` + `_release_run_slot()` are both in the `finally`, so no slot leak regardless of how the thread exits (normal, deadline exception, or anything else). The original "settrace bypass → slot leak" concern is **resolved**.
- **`ScriptDeadlineExceeded(BaseException)` vs `except Exception` test is correct.** The test at line 572 proves agent code can't swallow the deadline with `except Exception`.
- **SAFE_BUILTINS removal is correct.** The allowlist blocked harmless code (imports, class bodies) while `while True: pass` with zero builtins sailed through it. Removing it fixes both halves.
- **Concurrency cap + slot accounting** looks clean. `_acquire_run_slot` / `_release_run_slot` with a `threading.Lock` is the right pattern here.
- **CLAUDE.md doc update** is accurate — the known C-level limitation is documented.
- **60 new tests, 187 passing** — comprehensive coverage of the changed behavior.

---

### P1 — `SystemExit` escapes `run_python()` and can crash the API process

**File: `nous/api/tools.py`, line ~3470**

```python
# current
except (asyncio.TimeoutError, ScriptDeadlineExceeded):
    return {"content": [...], "is_error": True}
except Exception as e:                          # ← BUG: BaseException, not Exception
    return {"content": [...], "is_error": True}
finally:
    executor.shutdown(wait=False, cancel_futures=True)
```

**Why this is now a real bug:** P2 removed `SAFE_BUILTINS`, so `import sys; sys.exit()` now works in agent code. `SystemExit` is a `BaseException`, not an `Exception`. When user code calls `sys.exit()`:

1. `_run()`'s `finally` still runs correctly — trace cleared, slot released ✅
2. The `Future` stores `SystemExit` as its exception
3. `asyncio.wait_for` re-raises it
4. `except (asyncio.TimeoutError, ScriptDeadlineExceeded)` — miss
5. `except Exception as e` — **miss** (`SystemExit` is not `Exception`)
6. `SystemExit` propagates out of `run_python()` and up to the FastAPI/Starlette handler, which also uses `except Exception` — **the uvicorn worker process can exit**

**Fix — one line:**
```python
except BaseException as e:                      # catches SystemExit, KeyboardInterrupt, etc.
    return {
        "content": [{"type": "text", "text": f"Error: {type(e).__name__}: {e}"}],
        "is_error": True,
    }
```

**Required test:**
```python
async def test_sys_exit_is_contained(self, run_python_tool):
    """sys.exit() must not escape run_python — now that imports work, this is live."""
    result = await run_python_tool(code="import sys\nsys.exit(42)")
    assert result.get("is_error") is True
    text = result["content"][0]["text"]
    assert "SystemExit" in text
```

---

### P2 — `except BaseException` in agent code can swallow `ScriptDeadlineExceeded` and disable the tracer

This is lower severity (unusual pattern in agent code, and the asyncio `wait_for` grace-period eventually orphans the thread), but worth documenting:

When `_tracer` raises inside `exec()`, CPython injects the exception at the current execution point **and then disables the frame's trace hook** (`f_trace = NULL`). So if agent code does:

```python
try:
    while True: pass
except BaseException:
    result = "I caught the deadline"
```

`ScriptDeadlineExceeded` is caught, the trace hook is cleared, and the script continues executing with **no active deadline enforcement**. It will run until the outer `asyncio.wait_for(timeout + GRACE)` gives up and abandons the thread.

Test `test_timeout_not_swallowed_by_except_exception` (line 572) covers `except Exception` — it should add a variant for `except BaseException`.

---

### P3 — 2 failing tests in `test_config.py` (pre-existing)

PR description acknowledges these are environment-sensitive (bare `Settings()` picks up operator env vars). Not a blocker, but worth hardening separately — `monkeypatch.delenv` on the relevant vars would make them hermetic.

---

### Verdict

The implementation is conceptually correct and the test suite is thorough. The one merge blocker is the `SystemExit` escape path (P1 above), which is a one-line fix (`Exception` → `BaseException`) plus one test. Everything else is ready to ship.

tfatykhov:
**Note:** The two `CHANGES REQUESTED` reviews above are complementary — both should be addressed before merge:

1. **#5079070514** (Codex): `sys.settrace(None)` hook bypass + `SystemExit` escape
2. **#5079102108** (Nous): `SystemExit` escape details + `except BaseException` swallow path for `ScriptDeadlineExceeded`

Summary of what needs fixing:
- Shadow `sys.settrace` / `sys.setprofile` in exec namespace (no-ops) so user code cannot uninstall the deadline hook
- Catch `BaseException` (not just `Exception`) in the outer `run_python` try/except to contain `SystemExit` / `KeyboardInterrupt`
- Add tests for both: `sys.settrace(None)` + spin loop, and `sys.exit(42)`

tfatykhov:
## ✅ P1 Fixes Verified — Ready to Merge

The Claude Code fix job (`job-20260725-153428-33c9c30b`) completed successfully. Both P1 blockers from the earlier review are resolved:

### Fix 1: `sys.settrace` bypass blocked ✅
- Monkey-patches `sys.settrace` and `sys.setprofile` in the worker thread to reinstall the tracer after any disable attempt
- New test: `test_settrace_bypass_blocked` — `sys.settrace(None)` + infinite loop still times out and releases its concurrency slot

### Fix 2: `BaseException` escape blocked ✅
- Added final `except BaseException` catch block that translates `SystemExit` / `KeyboardInterrupt` to `is_error` responses (prevents API process crash)
- New tests: `test_system_exit_is_error`, `test_keyboard_interrupt_is_error`

### CI
- **Lint & Format:** ✅ pass
- **Tests:** ✅ pass (63 run_python + 175 core)

Latest commit: `4d5864d fix(run_python): block settrace bypass; translate BaseException to is_error`

**Verdict: APPROVED — safe to merge** 🚀

PR #578, opened 2026-07-29
fix(heart,brain): correlate episodes/decisions via session_id; delete two dead join tables

tfatykhov:
Three findings from a 3-agent review of why `heart.episode_decisions` and `heart.episode_procedures` were both empty in prod.

## 1. Root cause: a three-line omission

`api/tools.py` built `RecordInput` **without `session_id`** — although `RecordInput.session_id` has always existed (`brain/schemas.py`) and the dispatcher already injects `_session_id` for five other tools. All 287 populated `session_id` values in prod come from the deliberation path; the `record_decision` tool has set **zero, ever**. That NULL is what left episode↔decision association with no substrate.

**Gated behind `NOUS_DECISION_SESSION_ID_ENABLED`, default OFF.** Enabling it changes a *write* path: session-scoped decisions enter `DeliberationProtocol._is_duplicate`'s pool for the first time, where a match silently drops the deliberation decision at DEBUG level. Flag off ⇒ byte-identical to today.

## 2. The join rule, settled empirically

Defined once in `nous/brain/graph_constants.py`, measured against all 87 matchable prod decisions:

| rule | pairs | covered | lost | ambiguous |
|---|---|---|---|---|
| naive equality | 148 | 87/87 | 0 | **15** |
| strict window | 52 | 52/87 | **35** | 0 |
| **window + 60s grace** | **87** | **87/87** | **0** | **0** |

An exact 1:1 function. The two sessions that broke naive equality — `claude-code-f044-codesign` (6 episodes × 8 decisions = 48 pairs) and `claude-snn-smoketest-consult` (4 × 7 = 28) — resolve to 8-for-8 and 7-for-7.

Both extra terms are load-bearing, each falsified by measurement:

- **`agent_id` equality** — `session_id` is NOT agent-namespaced (`heartbeat-<hex>` / `subtask-<hex>` are generated identically per agent), so without it a same-named session on another agent materialises a cross-agent pair.
- **60s grace** — `pre_turn` records a deliberation decision at step 4 and creates the episode at step 5, so a decision can predate its own episode. On single-episode sessions (unambiguous ground truth): of 72 pairs, 28 fall before episode start, median 0.4s, max 2.2s. 60s and 5min give identical results, so the constant is not load-bearing.

Open episodes are additionally bounded by `LEAD(started_at) OVER (PARTITION BY agent_id, session_id)`.

## 3. Two dead paths deleted

- **`heart.episode_procedures`** (migration 067) — zero rows, zero readers, writers called only from tests. Its concept shipped on a different substrate: Laplace-smoothed effectiveness over `procedures.success_count/failure_count`, live with 13,402 success / 1,344 failure across 62 of 70 procedures and five consumers. Populating it would have created a second contradictory representation.
- **`EpisodeSignal` + `Brain.get_episode_for_decision`** — unwired by audit HD-4 (2026-06-09) for mapping the hardcoded-`"success"` episode outcome onto every decision in a session. Prod has 711 success / 96 NULL / 10 abandoned / **zero failure** episodes, so reviving it would relabel essentially every linked decision `success @ 0.8` — the corruption #577 reversed across 138 rows. Deleted rather than left dormant.
- **`heart.episode_decisions`** (migration 068) — superseded by the join.

## Incidental fixes
- `_build_decision_context` was N+1 (`Brain.get` per decision, each opening its own pool session and eager-loading four unused relationships) → one query, four fields.
- That block lands **outside** the `<transcript>` wrapper in instruction position and is not echo-screened; now delimited with `<decisions>`. Harmless before only because it was always empty.
- Corrected a false comment in `graph_densifier.py` claiming nothing else rebuilds these edges — `decision_graph_linker.py:165-173` does, and all 3 prod edges are its output.

## Verification
Full-suite failure-set diff on **real Postgres**: **77 failures before and after, zero regressions, zero fixed, +9 new tests.** (Two earlier baseline runs reported false regressions because pytest aborted collection when an untracked new test file survived `git stash`; instrumented exit codes caught it.)

## Honest limits
- The retrieval payoff is **unmeasured**. `discussed_in` edges are verifiably *consumed* by Stage 2 of `run_recall_pipeline`, but that they *improve* answers is a hypothesis. Per F044, edge existence is the lever where weight is not — plausible, unproven. Wants an eval-DB A/B before the flag is flipped.
- The 718 historical decisions with NULL `session_id` can **never** be linked; there is no backfill. This fixes the forward flow only.
- Writing `discussed_in` permanently de-orphans those decisions, excluding them from `backfill_orphan_decisions` thereafter. That does not revert with the commit.


tfatykhov:
@Codex - review latest

PR #582, opened 2026-08-03
fix(retrieval,eval): make leg failures reportable, stop caching failed expansions, land N3 filter dark

tfatykhov:
Implements **N1, N2, N3, N7, N8** from the 2026-08-02 external fix list (60-query prod clone + 130 gold-labelled MAB CR questions).

Every `file:line` in that list was re-verified against `be6361e` before implementing. **Three references did not survive** — see *Not implemented* below.

### N1 — `heart.recall` swallowed leg failures (`heart.py:1076-1082`)

A per-type sub-search failure was logged and dropped with **nothing on the return path**, so a caller could not distinguish "the fact leg found nothing" from "the fact leg crashed." A store two migrations behind raises `UndefinedColumnError` on the fact leg; the run completes, returns chunks and episodes, and produces plausible numbers.

Adds an optional caller-supplied `stage_errors` dict filled with `heart_recall_<type>` keys, mirroring the accumulator pattern one layer up (`acc.stage_errors` to `PipelineStats.n_stage_errors`). The pipeline now passes its accumulator through.

Fail-open **behaviour is unchanged** — the defect was invisibility, not the fail-open. Omitting the dict is byte-identical for every existing caller. This also closes the `PipelineStats` docstring that explicitly deferred it ("Surfacing those would require Heart instrumentation; deferred to a follow-up").

### N2 — failed query expansions were cached permanently (`query_expansion.py:234`)

`_cache_put` ran unconditionally, including when `_call_haiku` returned `[]` and `final == [query]`. One transient failure disabled expansion for that query hash for the full cache TTL. The guard condition already existed 8 lines below protecting the success log; the cache write had the same problem and no guard.

Prod was clean (0 degenerate entries in 1,844) — this is prevention, not remediation.

### N3 — adjacency boost credited structural anchors its sibling excludes (`retrieval_pipeline.py:1468`)

The boost's edge query filtered only `relation != 'contradicts'`. Its sibling `_record_recall_reactivation` at `:1553` — same file, near-identical query — carries `AND extraction_method IS DISTINCT FROM 'deterministic'`. Added verbatim.

Measured on a 60-query prod clone, 4 disjoint strata:

| metric | delta | p |
|---|---|---|
| recall@10 | **+0.0900** | 2.7e-5 |
| MRR | +0.0239 | 3.3e-6 |
| nDCG@10 | +0.0709 | 1.2e-6 |

**Framed as removing a known regression, not an improvement** — the gain over disabling the boost *entirely* is NOT established (off, filtered, and matched-control are statistically tied; only current prod is clearly worst). **The mechanism is NOT established**: the intuitive "structural stars dominate the degree sum" account was tested and failed (degree concentration 0.511 with vs 0.497 without). The added comment states the measurement, not a cause.

Ships behind `NOUS_GRAPH_ADJACENCY_BOOST_EXCLUDE_DETERMINISTIC`, **default OFF**, per the F084/F085/F086 land-dark convention. Also deletes the stale comment citing `spreading_activation.py:103` — that line is now a docstring about MAX aggregation, not a filter.

**Flip-time note carried into CLAUDE.md:** monitor degree-distribution drift. LTP reinforcement plus a degree-based boost is a rich-get-richer loop no frozen-corpus eval can observe.

### N7 — `nous_eval` measured a window production does not use

`recall_deep` does not truncate — median served length is ~77 rows, all reaching the model. Six of ~nine legs band below the rank-20 cutline and are unreachable by any top-10/20/30 eval, so prior nulls on F085/F086/spreading are **unverified, not settled**.

`retrieved_ids` already held the full served list, so the harness was discarding this at *scoring* time, not failing to collect it. No re-runs needed. Adds:

- `recall@served` + `mean_served`
- a recall curve over k in {3, 5, 10, 20, 40, 60, served}
- `leg_visibility()` — labels rows from the pipeline's **own existing provenance markers** (`metadata["retrieval_leg"]`, `metadata["stage_origin"]`, `source`, `type == "chunk"`), and flags legs whose median rank sits below the scoring cutline

Surfaced in both markdown and JSON. Smoke output reproduces the report's shape:

```
| config   | R@3   | R@5   | R@10  | R@20  | R@40  | R@60  | R@served |
| baseline | 0.000 | 0.000 | 0.000 | 0.000 | 1.000 | 1.000 | 1.000    |

| leg           | rows | median rank | best rank | visible@10 |
| exemplar      |   10 |        45.5 |        41 | NO         |
| keyed         |   10 |        35.5 |        31 | NO         |
| chunk         |   20 |        20.5 |        11 | NO         |
| heart_primary |   10 |         5.5 |         1 | yes        |
```

This is the substrate N9 needs: an arm can now assert leg visibility before reporting a null.

### N8 — F085/F086 bands calibrated against the scale, not the distribution

`keyed_fact_leg_score` (`config.py:448`) and `exemplar_leg_score` (`:494`) are both 0.55 — mid-range on the RRF [0,1] **scale**, but roughly the **10th percentile** of the observed **distribution** (n=2,834: min 0.4700, p10 0.5200, median 0.7824). Sorted insertion places a row where its *score* puts it, so the leg lands at rank 41-53. Strictly better than tail-appending, still invisible.

Documented as a **deliberate below-the-read-zone choice** for land-dark legs rather than a unit mismatch, so N7's check now enforces a stated intent. No ranking change — raising toward the median is an explicit decision needing an A/B.

## Not implemented, with reasons

**N4 — needs no repo change.** The code default is already `False` (`config.py:1194`) and `test_f044_tinyhippo_lite.py:349` asserts it. The drift was in the untracked local `.env`, which had that test **red**; corrected there. The prod `.env` is a separate file the owner must apply.

Related finding: at least two more flags drift the same way (`NOUS_RRF_K=30`, `NOUS_GRAPH_HUB_AUTOSURFACE_ENABLED=true`), each with its own failing default-asserting test. Those tests construct `Settings()` and read the ambient `.env`, so they cannot do their job — the guard written to catch exactly this drift was drowned in noise from the others. Not fixed here (out of scope).

**N6 — premise falsified, would have been a no-op.** The load-bearing claim "Nothing populates `RetrievalPlan.queries`" is false:

- `intent.py:208/218` append on **every** non-short-circuit path
- `layer.py:647` builds the plan, `:723` passes it to `context.build`
- `context.py:237-241` consumes it into `_query_texts`, used at `:743/787/886/1110/1265` **in preference to** `_default_query`

So `context.py:733-736` is the *fallback*, not the live query. The actual live pre-turn query is a join over a **set of at most 10 regex-extracted keywords** (`intent.py:145`) — word order destroyed, set-ordered, capped at 10. That is a different (and arguably worse) problem than the one described.

The measured effect may still be real, but it was measured on `recall_deep` with harness-constructed query strings — a path that never touches `context.py`. Needs re-diagnosis against the path that actually serves prod turns before any code change.

**N5 — design task, not a mechanical fix.** The item explicitly forbids a new global constant and requires allotment derived from store composition; its own data measures prod net-negative at higher k (consistent with #579/#580/#581). Needs prod measurement.

## Testing

- **28 new tests** (`tests/test_fixlist_n1_n3_n7.py`) pinning each defect, each flag default, and the flag-off byte-identity of N3
- `tests/eval/` — 204 passed
- Pre-existing failures in `test_heart.py` / `test_rrf_search.py` / `test_f065_configs.py` verified **identical at the stashed baseline** (unreachable Postgres + `.env`-vs-code-default drift)

## Deploy notes

All runtime changes are behaviour-preserving by default. `NOUS_GRAPH_ADJACENCY_BOOST_EXCLUDE_DETERMINISTIC` is the only new flag and defaults OFF; it needs a second-corpus A/B before flipping. N1/N2 take effect on deploy — N1 adds observability only, N2 changes only whether a *failed* expansion is written to cache.


tfatykhov:
@codex review

All three findings were real and shared one root cause — the N1/N7 instrumentation was computed but never reached its consumer. Fixed in `71cc77f`.

**P1 — Heart leg errors discarded by the eval runner.** Correct, and pointed: this was the same defect one layer above the code N1 fixed. `_run_one` returned only the three boolean flags, so `run_matrix` dropped `stats.n_stage_errors` and every report showed plausible metrics with no sign a leg had crashed.

- Added `QrelResult.stage_errors`, populated from `stats.n_stage_errors`.
- `_run_one` now emits `stage_error_<stage>` integer counters alongside the booleans.
- I deliberately did **not** take the "mark the affected qrel" option via `error`: that excludes the qrel from aggregates, which hides the partial run rather than reporting it. N1's intent is that a partial retrieval stays visible *and* countable.
- The aggregation loop needed a `bool`-before-`int` isinstance check — `bool` subclasses `int`, so without it the existing stage flags would be summed as 1/0 instead of counted as occurrences. Pinned by a test.

**P1 — visibility cutoff ignored `--top-k`.** Correct, and it inverted the feature's own verdict: a k=20 run with a leg at median rank 15 was reported `visible@10: NO` despite having been measured. `top_k` is now threaded from the CLI through `render_markdown` / `render_json` into `leg_visibility`. Test asserts both directions (`visible@10` vs `visible@30` on the same data).

**P2 — leg provenance not serialized.** Correct — the only copy was the ephemeral markdown. `render_json` now persists the aligned per-qrel `retrieved_legs`, per-qrel `stage_errors`, the computed `leg_visibility` rows, and the run's `top_k` so the artifact self-describes the depth it was scored at and consumers can recompute visibility at any k.

Tests: 7 new (35 in this file). `tests/eval` 211 passed; integration + report suites green. The one remaining local failure (`test_f065_configs`) is pre-existing `.env`-vs-code-default drift, verified identical at the stashed baseline and called out in the PR body.

tfatykhov:
@codex review

All four verified against the code before fixing. Fixed in `72aa984`.

**P1 — spreading rows never classified.** Confirmed at `retrieval_pipeline.py:2057-2077`: `source` is `"spreading_activation"` *and* `metadata["stage_origin"]` is `brain_graph`/`heart_graph_memory` on the same row, so my `stage_origin`-first ordering meant the label was unreachable. This was the worst of the four — it blinded N7 for one of the exact legs the report exists to measure. The spreading check now precedes `stage_origin`, with a test pinning that non-spreading `graph_expanded` rows still resolve to `heart_graph_memory` so the reorder steals nothing from the graph legs.

**P1 — declared depth vs computed depth.** Confirmed: `_metrics_table`, `_recall_curve_table`, `render_json`'s `_metrics_to_dict`, and the delta table all called `compute_metrics` at its default 10 while the artifact recorded the configured `top_k`. `top_k` is now threaded into every metric computation in both renderers, and the markdown columns are labelled at the real depth (`P@30`/`R@30`/`nDCG@30`, header `scored at k=30`). A test asserts the *value* changes with depth, not just the label — gold at rank 25 gives `R@k` 0.000 at k=10 and 1.000 at k=30. Worth noting for future readers: `MetricsResult.p_at_10` etc. are historical key names, not depths; the depth is now recorded explicitly beside them.

**P1 — partial runs invisible in markdown.** Confirmed — I had added the counters to JSON only, so the operator-facing report was the one consumer still showing a healthy table over a crashed leg. Added a "Partial retrieval detected" banner that renders *before* the metrics tables (test asserts the ordering), names the failed stages, and marks the numbers invalid for comparison.

**P2 — `eval_runs` payload built independently.** Confirmed at `retrieval.py` — `_metrics_compact` never saw the report file's structures. It now carries `top_k`, `r_at_served`, `recall_curve`, the computed `leg_visibility` rows, and `n_qrels_partial`, so historical analysis doesn't depend on the report file surviving.

Tests: 9 new (44 in this file); `tests/eval` + integration 220 passed. CI was green on `71cc77f` and is re-running for this commit.

One recurring theme across both rounds worth stating plainly: every finding has been "the data is computed but some consumer never receives it." That was the shape of N1 itself, and it has now recurred at the runner, the markdown, and the eval_runs payload. If you see a fourth instance, it's likely a real pattern in how this harness fans out rather than isolated misses.

tfatykhov:
@codex review

All three verified before fixing. Fixed in `e458b8a`. Notably two of the three were defects I introduced in earlier rounds, not pre-existing issues — the review is catching regressions in the fixes themselves, which is the useful case.

**P1 — duplicate telemetry reported as a crashed run.** Confirmed at `retrieval_pipeline.py:1131-1140`, whose own comment calls graph corroboration *"signal, not noise"*. My round-2 blanket `stage_error_` prefix would have made every healthy graph-enabled eval render the new partial-run banner and declare its own metrics invalid. That is strictly worse than having no banner — a warning that fires on success trains operators to ignore it, which would have destroyed the N1 signal I was trying to add. Added `_NON_ERROR_STAGE_COUNTERS`: those keys route to `stage_info_*` and are excluded from `QrelResult.stage_errors`. Tests pin both directions — duplicates-only renders no banner, a real failure still does.

**P1 — visibility used the wrong statistic.** Agreed, and this was a design error rather than a wiring slip. A pooled median answers "where does this leg's mass sit", not "did the scoring window observe this leg" — and those diverge exactly where it matters, for large-allotment legs whose head scores while their tail drags the median down. Given the report instructs operators to treat `visible=False` nulls as inconclusive, the old statistic would have discounted valid measurements.

Visibility is now per-qrel (a leg is observed on a qrel when any of its rows lands at rank ≤ cutoff), aggregated into `participation_rate` and reported with `n_qrels_present` / `n_qrels_within_cutoff` in the markdown and both JSON payloads. `median_rank`/`best_rank` remain as diagnostics. The regression test is the case your example describes: rank 1 plus a 20-30 tail, median above the cutline, `participation_rate == 1.0`, `visible=True`.

**P2 — the expansion guard had a hole.** Confirmed: `_fuse` dedupes on `c.lower().strip()`, so a case-only echo gives non-empty `cleaned` and `final == [query]`, and the no-op was still cached. Gate is now `len(final) > 1`. Worth flagging that the original fix list proposed exactly `if len(cleaned) > 0` — your version is a genuine improvement on the source document, not just on my patch.

One scope note so it isn't silent: I also moved the success log two lines below onto the same `len(final) > 1` gate. It shared the old predicate for the identical purpose, and leaving `>0` and `>1` adjacent would be both confusing and wrong in the same way — logging "expansion success" for a result that fused back to the bare query is the same misreport in a different channel. Say the word if you'd rather that stayed untouched.

Tests: 8 new (51 in this file); 286 passed across eval + integration + expansion suites.

tfatykhov:
@codex review

Both correct. Fixed in `691082c`.

**P1 — participation denominator.** You're right, and this one stings: it was the inversion of the problem round 3 set out to fix. I replaced a statistic that could call an observed leg invisible with one that could call a barely-present leg fully observed — still answering a question the operator didn't ask, just failing in the opposite direction. Dividing by `n_qrels_present` conditions the rate on the leg having emitted at all, so 1-of-100 read `1.00`.

Denominator is now `n_qrels_evaluated` (all valid qrels, errored excluded). `n_qrels_present` stays alongside so emission coverage and in-window coverage remain separately legible — they answer different questions and collapsing them is what caused this. New field is in the markdown table and both JSON payloads. Tests pin 1-of-10 → 0.10, errored-qrel exclusion, and full coverage still 1.00.

**P2 — delta labels.** Confirmed: `_delta_table` renders `MetricsResult` field names directly, and those names carry a historical `_at_10` suffix while holding whatever depth `compute_metrics` was handed. I parameterized the aggregate headers in round 2 and missed this renderer. The three depth-bearing rows are relabelled at `top_k`; `mrr`/`p_at_1`/`p_at_5` keep their names, and a test asserts the k=10 rendering is byte-identical to before.

Tests: 6 new (57 in this file); 280 passed across eval + integration + expansion + pipeline suites.

Standing note on the run: every finding across four rounds has been real — no false positives — and the last two rounds have been catching defects in my own fixes rather than in the original PR. Both of those were in the same place: the statistic and its denominator in `leg_visibility`. If you have another angle on that function specifically, it's the part of this change I'd most want a fifth look at.

tfatykhov:
@codex review

Correct, and this was the best catch of the five rounds. Fixed in `f337cc3`.

`leg_visibility` derived its universe of legs from observed rows, so a leg emitting zero rows on every qrel never entered the aggregation and was omitted from the report entirely — no row, no `0.00`. That is the most extreme case of an unobserved arm, and it was the single case the report stayed silent about. An operator reading a null for the keyed or exemplar leg would find no keyed row and nothing telling them the null was uninformative.

Worth naming plainly: this is the exact failure mode the whole N1/N7 change set exists to remove — a signal computed from what was present, blind to what was absent — reproduced *inside* the diagnostic built to remove it. Silence read as "nothing to report" precisely when the arm contributed nothing.

**Fix.** `leg_visibility` takes `expected_legs` and seeds an entry per enabled leg, so silent legs render at `participation 0.00` / `n_rows=0`. `RunResult.expected_legs` carries the set from `_expected_legs()`, derived from the settings actually used for that config — from master flags rather than observed rows, since the whole point is naming legs that produced nothing. `spreading_activation_enabled` is `"auto"|"true"|"false"` and `"auto"` resolves at runtime against graph density, so anything but an explicit `"false"` counts as attempted; a test pins both directions.

Rendering: silent legs are labelled `*(silent)*` with em-dashes for median/best rank rather than `0.0`/`0`, which would imply a rank that never existed. `median([])`/`min([])` are guarded, with a test for the empty-input edge.

Tests: 5 new (62 in this file); 289 passed across eval + integration + expansion + pipeline + F044 suites.

Tally across the run: 16 findings, all real, zero false positives. Rounds 3-5 were all defects in my own fixes, and four of those five landed in `leg_visibility` or its consumers — which is a fair signal that the function was the riskiest thing in this PR and worth the extra passes rather than merging at round 2.

tfatykhov:
@codex review

Both correct. Fixed in `de8e626`.

**P1 — the gate contradicted its own report.** This is the one that would have mattered most in practice. `_stage_error_summary` declares a partial run's metrics "invalid for comparison", but `decide_gate_f050` computed a normal delta anyway and `retrieval.py` returns exit 0 when it passes — so a crashed fact leg could produce a **passing automated merge signal** underneath a report saying the comparison was meaningless. Prose that CI cannot read is not a safeguard; I added the banner in round 2 and never connected it to the one output that acts on it.

Added Rule 0: any stage error on a gate-eligible qrel in either gated config fails the gate, naming the config and count. It runs before the aggregate rule so a partial run reports *why* rather than surfacing as a delta miss. Test asserts healthy runs are unaffected — identical arms still fall through to Rule 1 rather than short-circuiting on Rule 0.

**P2 — leg nesting.** Confirmed at `retrieval_pipeline.py:1173` (spreading) and `:980-982` (Stage 2b): every graph-derived leg sits under `graph_recall_enabled`, so sub-flags alone don't mean "attempted". For `graph_off`, `spreading_activation_enabled` keeps its `"auto"` default and round 5's seed listed spreading as enabled-but-silent — labelling a deliberately disabled arm as one that failed to emit, which inverts the point of the seed. `_expected_legs` now requires the master switch for every graph leg, and `heart_graph_memory` additionally requires `cross_type_linking_enabled` to match the full Stage 2b condition.

Tests: 6 new (68 in this file); 280 passed across eval + integration + expansion + pipeline suites.

Running tally: 18 findings, all real, zero false positives. Rounds 3-6 have all been defects in my own fixes, and the pattern is consistent enough to name — each round I added a signal, and the next round found somewhere that signal failed to reach a consumer, or reached one that misread it. N1 was "a crashed leg looks empty"; round 6 is "a crashed leg still passes the merge gate." Same defect, four layers apart.

tfatykhov:
@codex review

Correct. Fixed in `91f0166`.

Verified at `retrieval_pipeline.py:968-982`: Stage 2 (which emits `stage_origin="heart_graph"`) sits inside the same condition as Stage 2b, including `cross_type_linking_enabled`. With graph on and cross-type linking off it cannot run, so round 6's seed listed it as enabled-but-silent — the same inversion you caught last round, one leg over.

`heart_graph` now requires both flags alongside `heart_graph_memory`. One deliberate asymmetry worth flagging so it doesn't read as an oversight: `brain_graph` still requires only the master switch, because Stage 4 at `:1173` is gated on `graph_recall_enabled` alone. The three graph legs have genuinely different guards and a test now pins that they aren't collapsed together.

Tests: 1 new (69 in this file); 245 passed across eval + integration.

Process note, since this is round 7 and I'd rather say it than let the loop run silently: the runtime changes (N1 heart leg reporting, N2 expansion cache, N3 flagged-dark filter, N8 band docs) have been stable since round 1 — every finding from round 2 onward has been in `nous_eval`, which is dev-only tooling excluded from the prod image. I've raised with the repo owner the option of splitting the eval harness into its own PR so the runtime fixes can land while the harness keeps iterating. Either way I'll keep addressing findings here until they stop or that call is made.

tfatykhov:
@codex review

**Scope reduced.** Owner call after 8 rounds: split the PR. `6ffc054` removes the leg-visibility verdict from this PR entirely; it moves to draft #583.

Your three round-8 findings are all valid and all preserved there — I'm not dismissing them, I'm declining to patch them a fifth time in the same place.

**Why.** `_expected_legs` predicted which legs the pipeline would attempt by re-deriving pipeline control flow from config flags, in a second location. That's a duplicated invariant, and your F2 showed it can't be correct even in principle: whether the one-hop fallback runs depends on whether spreading succeeded *at runtime*. Rounds 5, 7 and 8 were each a tightening of that flag model, and each time you found another producer or branch it failed to capture — `brain` never seeded, `brain_graph` seeded when the fallback is skipped, `heart_graph_memory` also produced by Stage 4's untyped path. Those aren't three bugs; they're one wrong design emitting bugs on demand.

#583 fixes it structurally: the pipeline reports `attempted_legs` from its own accumulator (which already carries most of the signal), and `_expected_legs` is deleted rather than corrected.

**Removed here:** `LegVisibility`, `leg_visibility()`, `_expected_legs()`, `RunResult.expected_legs`, `_leg_visibility_table()`, the `leg_visibility` blocks in `render_json` + `_metrics_compact`, and their tests.

**Kept here** — everything with zero open findings:
- **N1** `heart.recall` stage_errors → `PipelineStats` → `QrelResult` → markdown partial-run banner → F050 gate Rule 0 (your R1/R2/R3/R6 findings, all fixed)
- **N2** expansion cache guard at `len(final) > 1` (your R3 fix)
- **N3** deterministic-edge filter, default-OFF flag
- **N8** band calibration documented
- **N7** `recall@served`, `mean_served`, recall curve over k, `top_k` threaded through both renderers, delta labels at the real depth (your R2/R4 fixes)
- raw leg provenance: `_leg_of()` **including your R2 spreading-ordering fix**, and `QrelResult.retrieved_legs` serialized 1:1 with `retrieved_ids` — so #583's analysis is reconstructable at any depth without re-running anything

Tests: 55 in this file; 327 passed across eval + integration + expansion + pipeline suites. CI green on the parent commit.

For the record across this review: **22 findings, all real, zero false positives.** Several were regressions in my own fixes, and catching those is exactly what made the split the right call rather than merging at round 2 on a green CI.

tfatykhov:
@codex review

Correct, and this one lands harder than a normal P2. Fixed in `769e220`.

Verified at `nous/api/tools.py:445`: the Brain section is emitted only when the call is `search_all` or explicitly includes `"decision"`, so a `memory_types=["fact"]` qrel can surface decisions via graph expansion or spreading that the formatter drops. Counting every `retrieved_id` overstated `mean_served` and let `R@served` credit a gold sitting in the dropped section.

That matters more than the severity suggests, because `R@served` is N7's headline metric and its entire claim is *"this is what the model actually got"*. A metric that overstates in exactly the direction of its own thesis is the worst kind to ship.

**Fix.** Adds `QrelResult.served_ids`, computed by `_served_ids()` from the qrel's memory_types. Empty means no narrowing, so `search_all` qrels are byte-identical. Existing metrics (MRR, P@K, R@K, nDCG) still derive from `retrieved_ids` — the change is scoped to the two served metrics so the blast radius stays inside N7 rather than silently moving numbers the rest of the harness reports.

Two deliberate choices worth naming:
- `_served_ids` mirrors **one** narrow condition, not the whole bucketing logic. Reproducing the formatter wholesale is the duplicated-invariant pattern that just cost us the split; one rule with a test pinned to `tools.py:445` is the smaller debt.
- Heart-routed graph/spreading rows (`stage_origin="heart_graph_memory"`) are **kept**, because they do render. A test pins that distinction so a future edit can't collapse "spreading row" into "dropped row".

The proper fix is the formatter reporting the IDs it rendered — one source of truth — which is the same redesign as the leg-visibility work. Noted on #583 rather than half-built here.

Tests: 5 new (60 in this file); 273 passed across eval + integration + pipeline suites.

tfatykhov:
@codex review

All three correct, and all three are consequences of my round-9 fix. `0d2b488` resolves them by removing the metric rather than patching it a third time.

- **empty-vs-unset (P1):** `served_ids or retrieved_ids` treated "the formatter rendered nothing" as "no narrowing applied" — re-creating the exact overstatement I'd just removed, in the sentinel I introduced to remove it.
- **decision-only qrels (P1):** the mirror image. My early return counted everything when `"decision" in memory_types`, but Stage 4 returns non-decision neighbors the Heart section drops for a decision-only search. And you're right that such qrels are live in `tests/fixtures/eval_smoke.jsonl`.
- **not serialized (P2):** no consumer could verify the corrected value, so they'd recompute the inflated one from the only IDs present.

**The pattern is the one that already cost us the split.** `_served_ids` re-implemented `_format_pipeline_text`'s section-eligibility rules in a second place — the same mistake as `_expected_legs`, made two hours after I wrote a commit message saying the proper fix was for the formatter to report its own rendered IDs. I shipped the mirror anyway and it overstated in two different directions across two rounds.

A metric whose entire claim is *"this is what the model received"* cannot rest on a second-hand model of what the formatter renders. So `r_at_served` / `mean_served` / `served_ids` move to #583, where the formatter reports rendered IDs and the metric becomes exact by construction.

**Kept:** `recall_curve` over k ∈ {3,5,10,20,40,60}. It carries N7's actual finding — a fixed-k report is blind to most of what the pipeline returns — and it's computed over `retrieved_ids`, the *same basis* as `r_at_10` / `ndcg_at_10` / MRR. It makes no claim the harness can't substantiate, which is exactly what went wrong with `@served`.

N7's acceptance is therefore partially deferred, and the report now says what it measures instead of shipping a number that overstates in the direction of its own thesis.

Tests: 55 in this file; 268 passed across eval + integration + pipeline suites.

PR #583, opened 2026-08-03
feat(eval): leg visibility from pipeline-reported attempted legs (N7 follow-up)

tfatykhov:
Rebuilt on current `main`. Implements the redesign this PR was opened to hold.

## The problem it fixes

The predecessor derived *"which legs ran"* from config flags **inside the eval harness**, reproducing pipeline control flow in a second place. That cannot be correct: the one-hop graph fallback is skipped when spreading activation succeeds, decided **at runtime**, so no combination of flags predicts it.

Four review rounds each tightened the flag model, and each time another producer or branch escaped it:

| round | escape |
|---|---|
| 5 | silent legs omitted entirely |
| 7 | `heart_graph` also needs `cross_type_linking_enabled` |
| 8 | `brain` never seeded; `brain_graph` seeded when the fallback is skipped; `heart_graph_memory` also produced by Stage 4 |

Those aren't three bugs. They're one wrong design emitting bugs on demand.

## The fix: the producer reports what it did

`_PipelineAccumulator.attempted_legs` is marked at each stage's **entry**, before its work runs, so a leg that executed and produced nothing is still named. Surfaced as `PipelineStats.attempted_legs`.

Eleven marks, including the runtime branch the flag model could not represent:

```python
if use_spreading:
    acc.attempted_legs.add("spreading_activation")
...
if not use_spreading:
    acc.attempted_legs.add("brain_graph")
    acc.attempted_legs.add("heart_graph_memory")   # untyped Stage 4 neighbours
```

**All three round-8 findings resolve by construction, not by patch.**

## Harness

`QrelResult.attempted_legs` carries the per-qrel report; `RunResult.attempted_legs` unions across the config. `leg_visibility(per_qrel, cutoff, attempted_legs)` seeds silent legs from that. **`_expected_legs` is not reintroduced.**

Two earlier corrections were independently right and are retained:
- visibility is computed **per qrel** — a leg's own long tail must not hide its head
- the participation denominator is **all valid qrels**, not the qrels where the leg emitted (1-of-100 reads `0.01`, not `1.00`)

## Still out of scope

`recall@served` needs the **formatter** to report the IDs it rendered — a different producer, a separate change. Two harness-side derivations of it both overstated; a third attempt doesn't belong here.

## Testing

14 new tests. Four assert the **producer** side directly rather than only the consumer:
- every label `_leg_of` can emit has a corresponding mark
- the fallback marks *both* labels it can emit
- spreading is marked inside its own branch, not beside the fallback

302 passed across eval + integration + pipeline suites. The one remaining local failure (`test_f065_configs`) is pre-existing `.env`-vs-code-default drift.

### DIFF
--- a/nous/api/retrieval_pipeline.py
+++ b/nous/api/retrieval_pipeline.py
@@ -765,11 +765,15 @@ async def _run_stages(
     if getattr(settings, "keyed_fact_leg_enabled", False) and (
         search_all or "fact" in search_types
     ):
-        acc.attempted_legs.add("keyed")
         try:
             vocab = await heart.facts.entity_key_vocabulary()
             candidates = extract_entity_candidates(query, vocab=vocab)
             if candidates:
+                # Marked here, not at the block: a query yielding no entity
+                # candidates never reaches the fetch, and reporting it as a
+                # keyed leg that ran and found nothing is the same false
+                # attribution corrected for the graph seed loops.
+                acc.attempted_legs.add("keyed")
                 acc.keyed_leg_used = True
                 acc.keyed_results = await heart.facts.fetch_by_entity_keys(
                     candidates, limit=getattr(settings, "keyed_fact_leg_k", 8)
@@ -940,7 +944,6 @@ async def _run_stages(
         and (search_all or "fact" in search_types)
         and _is_classification_shaped(query, getattr(settings, "exemplar_max_query_words", 64))
     ):
-        acc.attempted_legs.add("exemplar")
         try:
             if await heart.facts.has_exemplars():
                 acc.exemplar_leg_used = True
@@ -950,6 +953,10 @@ async def _run_stages(
                 embedder = getattr(heart, "_embeddings", None)
                 q_vec = (await embedder.embed(query)) if embedder is not None else None
                 if q_vec:
+                    # Marked only once BOTH gates pass (a non-empty exemplar
+                    # store and a usable query vector); either failing means
+                    # no fetch is issued, so the leg was not attempted.
+                    acc.attempted_legs.add("exemplar")
                     # Codex r8: exclude the F071 already-in-context fact set from
                     # the fetch so ids the assembly-time F071 filter will drop do
                     # NOT spend the K budget (which would starve fresh
--- a/nous/api/retrieval_pipeline.py
+++ b/nous/api/retrieval_pipeline.py
@@ -721,7 +721,6 @@ async def _run_stages(
         # (chunks_searched stays False) from "flag on but stage failed"
         # (chunks_searched True + n_stage_errors["chunk_recall"] non-zero).
         acc.chunks_searched = True
-        acc.attempted_legs.add("chunk")
         try:
             acc.chunk_results = await _search_episode_chunks(
                 heart=heart,
@@ -740,6 +739,9 @@ async def _run_stages(
                 # conflating them IS the bug.
                 limit=settings.episode_chunk_recall_limit,
                 settings=settings,
+                # The helper marks "chunk" only where it actually queries —
+                # the vector path returns early with no embedder/vector.
+                attempted=acc.attempted_legs,
             )
         except Exception:
             # Match the sibling stages' pattern (spreading_activation,
@@ -1255,9 +1257,6 @@ async def _run_stages(
                 )
 
         if use_spreading:
-            # Spreading branch taken. Its rows are labelled
-            # "spreading_activation" regardless of stage_origin.
-            acc.attempted_legs.add("spreading_activation")
             try:
                 from nous.brain.schemas import NeighborResult
                 from nous.brain.spreading_activation import (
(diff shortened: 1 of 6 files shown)


tfatykhov:
@codex review

Correct. Fixed in `8debc02` — and the finding pointed at something one layer deeper than serialization, so I fixed both.

You're right that the config-wide union loses exactly the runtime decision this PR exists to preserve: spreading for source A, one-hop fallback for source B, and a consumer filtering to B sees `spreading_activation` in the union with no way to tell it was never attempted there.

**1. Serialization (the literal ask).** `render_json` now persists `q.attempted_legs` per qrel, so source- and query-level visibility is reconstructable from the archived report alone.

**2. The computation, which was wrong the same way.** `leg_visibility` now defaults `attempted_legs` to the union over the qrels **actually passed in**, rather than requiring a caller-supplied set. Your example is reachable in-process too — anyone calling `leg_visibility` on a source-filtered subset while passing the config-wide union gets the same misattribution, and per-source reporting is precisely that call shape. A filtered call is now correct by default instead of correct only if the caller remembers to narrow the union as well. An explicit set still wins, for callers deliberately widening beyond the given qrels.

The second is the one I'd have regretted missing. Fixing only the artifact would have left a correct-looking JSON file produced by a computation that could still misreport in memory.

Tests: 3 new (17 in this file) — per-qrel serialization, a filtered subset not inheriting another source's legs, and explicit-set override. 240 passed across eval + integration suites; CI was green on the parent commit.

tfatykhov:
@codex review

All three correct, and they are one mistake with three instances. Fixed in `15cd51d`.

I marked `attempted_legs` where each stage's `if` opens — but every one of those blocks is reachable in a configuration where its work loop iterates an **empty seed list** and issues no neighbour query. On exactly those runs the leg was reported attempted-and-silent, which is the false attribution this mechanism exists to prevent, reintroduced by the mechanism itself.

- **`heart_graph`** — the outer block also fires on chunk-only retrieval via `acc.chunk_results` (present for Stage 2b's benefit), where the `acc.heart_results[:3]` loop makes no query. Now gated on a fact/episode seed existing.
- **`heart_graph_memory`** — an explicit procedure/censor scope reaches the block with non-empty `acc.heart_results` but produces no seeds, since Path A accepts only fact/episode rows and chunks. Marker moved *after* seed construction, gated on non-empty `mem_seeds`.
- **`brain_graph` + `heart_graph_memory`** — reachable via `heart_fact_seeds` alone (fact-only retrieval, spreading off) with `decision_results` empty. Now gated on an expandable decision, mirroring the loop's own `score is None` skip.

The correction underneath all three: **"attempted" must mean the leg issued its query, not that an enclosing conditional was entered.** That is the only reading under which "attempted but silent" is a true statement — and I had the weaker one, which is how a leg that never ran could be reported as having run and returned nothing.

Worth noting this is the same failure the flag-derived predecessor had, one level in: it inferred leg execution from configuration; I inferred it from block entry. Both are proxies for "did this code path actually do its work", and both were wrong in the cases where the proxy and the reality diverge. The gates now sit on the seed lists the loops actually consume.

Tests: 1 new, asserting each marker is preceded by its seed guard (so a future edit cannot quietly hoist one back to block entry). It also widened an earlier window-based assertion that the new comments had pushed the marks out of. 297 passed across pipeline + eval + integration suites.

tfatykhov:
@codex review

Both correct — same class as last round, and these were the two remaining instances. Fixed in `cea873c`.

- **`keyed`** — a query yielding no entity candidates never reaches `fetch_by_entity_keys`, so ordinary no-match queries were reported as a keyed leg that ran and found nothing. Marker moved inside `if candidates:`, beside `keyed_leg_used`.
- **`exemplar`** — an empty exemplar store *or* a missing query vector both skip `fetch_exemplars_by_vector`. Marker moved inside `if q_vec:`, so it denotes an actual retrieval attempt rather than a classification-shaped query merely arriving.

Rather than wait to be told a third time, I re-checked the remaining five markers against the same question — *does a query definitely get issued once this line runs?* `heart_primary`, `chunk`, and `brain` all issue theirs unconditionally after the mark; `keyed_r2` calls `entity_keys_for_facts` immediately. Those are correct as they stand. The test now asserts the seed guard for every gated marker, so none can be hoisted back to block entry by a later edit.

This is the third round on one idea, and the through-line is worth naming: every version of this instrumentation has failed by marking a **proxy** for "the leg did its work" rather than the work itself. The flag-derived predecessor used configuration; then I used block entry; then the try-block top. Each proxy is correct until the case where it diverges from reality, and each round has been that case. The markers now sit on the branch that immediately precedes the query.

tfatykhov:
@codex review

Both correct. Fixed in `1108dc6`.

One of these falsifies a claim I made in my last reply. I wrote that `chunk` "issues its query unconditionally after the mark" — I asserted that from the call site without reading the callee, and it is wrong: `_search_episode_chunks` returns `[]` before any SQL when there is no embedder or the query embeds empty. That was a verification I claimed and did not do.

- **`chunk`** — the caller genuinely cannot know which path ran (the vector path returns early; the hybrid path still runs keyword-only in the same conditions). The marker moves **into** the helper via an optional `attempted` sink, added at each of the two points it is about to query. Same producer-reports-its-own-work shape as the rest of this PR, one level deeper.
- **`spreading_activation`** — `spreading_activation_search` returns immediately on empty seeds without issuing its CTE, reachable with `graph_recall_max_expand=0` and no heart fact seeds. Marker moved after seed construction, gated on non-empty `seeds`.

Then, instead of asserting it again, I **read the callee** for every remaining marker:

| leg | why it's sound |
|---|---|
| `heart_primary` | `heart_types` is always a subset of `_recall`'s `search_map` keys and the block is guarded on it being non-empty, so `search_map` cannot be empty and the query runs |
| `brain` | `brain.query` is the next line |
| `keyed_r2` | `entity_keys_for_facts` immediately, with `r1_ids` non-empty by the enclosing gate |

Four rounds on one idea, and the shape has been identical every time: I mark a **proxy** for "this leg did its work," you find the case where the proxy and the work diverge. Configuration, then block entry, then the top of the try, then the call site of a helper that early-returns. The general correction is that only the frame that issues the query knows whether it issued one — which is why `chunk` now marks inside the helper rather than around it.

298 passed across pipeline + eval + integration suites.

PR #585, opened 2026-08-08
feat(dag): F087 durability spine — durable result delivery, wall-clock reaper, live token accounting

tfatykhov:
## Summary

The DAG state machine core is sound — failure cascade, fix-stage recovery (F066.1), deferral backstops and workspace containment all work and are well covered. **Every gap this PR closes is at the edges**, where the subsystem meets the rest of Nous. No existing step of `_advance_dag` is rewritten; all changes are additive seams.

### P1 — Result delivery to the main loop did not exist

`DAGOrchestrator.__init__` accepted `bus` and assigned `self._bus` (`orchestrator.py:79`), then never read it. `grep emit|publish` across `nous/dag/` returned zero matches. `_check_dag_completion` wrote `result_summary` to a row and stopped: no event, no notification, no memory write. A multi-hour DAG completed in total silence, and the only way to learn the outcome was to poll `dag_manage` or open the dashboard.

**Reaching terminal and being delivered are now separate transitions.** Migration `069` adds `delivered_at` / `delivery_attempts` / `delivery_error` to `execution_dags` (plus a partial index on the sweep predicate). `_check_dag_completion` is unchanged; a new `tick()` phase drains `status IN (completed, failed, partial, cancelled) AND delivered_at IS NULL`. Because the work queue is a table rather than in-memory state, a process that dies between the two writes re-delivers on the next tick — at-least-once, surviving restart.

> The event bus **cannot** provide this guarantee: `EventBus.emit` drops on `QueueFull` and never blocks (`events.py:180-184`). Simply wiring the long-dead `self._bus` would have *looked* like a fix while still losing notifications under load. That is why durability lives in the database and the bus is only one leg.

**Three legs** in new `nous/dag/delivery.py`, each independently flagged and independently guarded so one failing never suppresses the others:

| Leg | Flag (default) | Required? |
|---|---|---|
| Bus emit `dag.completed` / `dag.failed` | `NOUS_DAG_DELIVERY_BUS_ENABLED` (on) | best-effort |
| Agent-authored prose summary | `NOUS_DAG_DELIVERY_AGENT_SUMMARY_ENABLED` (**off**) | best-effort, bounded, template fallback |
| Telegram push | `NOUS_DAG_DELIVERY_TELEGRAM_ENABLED` (on) | **required iff token AND chat id configured** |

The requiredness rule is the subtle part. Telegram is the leg the user actually sees, so a transient HTTP failure must bring the sweep back — but when Telegram is *unconfigured*, making it required would burn every attempt for every DAG against a channel that does not exist. Bus and summary stay best-effort because retrying a notification the user already received is worse than dropping an event.

Retries are bounded by `NOUS_DAG_DELIVERY_MAX_ATTEMPTS` (5). On exhaustion the DAG is marked delivered **with `delivery_error` set** — it stops looping, but the failure stays visible on the row instead of vanishing.

The agent-summary leg runs as a normal cognitive turn, so its episode lands in Heart on its own. That is how a finished DAG reaches the next conversation's context without a separate memory-write leg.

### P1 — No wall-clock backstop on `running` nodes

`_effective_timeout` was consumed only at launch (handed down to the subtask) and inside `_poll_awaiting_checks`. Nothing checked elapsed time for a node in `status='running'` — enforcement was fully delegated to the subtask executor. Since `reclaim_stale` runs once at worker start and only touches rows already past their timeout, a subtask orphaned by a crash *with time left on the clock* left the node `running` forever, which kept its DAG `running` forever, which consumed one of `MAX_ACTIVE_DAGS = 5` forever. Five orphans brick `dag_create` permanently.

`_reap_overrun_nodes` fails a running node past `_effective_timeout(node) + grace`, tearing down the primitive **before** marking it failed (F064.1 ordering) so a still-live subtask stops burning tokens and holding a worker slot.

**This defaults ON** (`NOUS_DAG_NODE_REAPER_ENABLED`) — a backstop that ships dark is not a backstop, and it only fires in an already-broken state. The 300 s grace (`NOUS_DAG_NODE_TIMEOUT_GRACE_SECONDS`) exists so it never preempts the primitive's own, richer error.

### P1 — Token budget enforcement was unreachable code

`DAGStore.update_dag_tokens` (`store.py:364`) had **no caller in `nous/`** — only two tests. `tokens_consumed` was therefore structurally always `0`, so the `ratio >= 1.0` branch in `_advance_dag` step 2 had never once executed in production. `token_budget` was a knob the agent could set that did nothing.

Now wired from `Subtask.tokens_in + tokens_out` on the running→terminal edge. `_sync_subtask_node` is re-entrant across ticks, so migration `069` adds `dag_nodes.tokens_counted` and the add is gated on winning a false→true flip — the guard is in the UPDATE's `WHERE` clause, not a read-then-write, so two concurrent ticks cannot both add. The trigger is the *subtask* reaching terminal rather than the node, so the `awaiting_check` path (subtask done, node still polling a shell command) is counted too.

Accounting is **live**; enforcement is **dark** (`NOUS_DAG_TOKEN_BUDGET_ENFORCEMENT_ENABLED=false`). Flipping both at once would start cancelling DAGs for anyone who had set `token_budget` casually. The over-budget WARN logs either way so budgets can be sized before enabling the cancel.

> Same PR, opposite defaults — decided by blast radius, not by a blanket land-dark rule.

### P2 — The tick was silently unwired when heartbeat is off

`main.py` wired the tick `if heartbeat_runner is not None` but `register_dag_tools` ran unconditionally. With `NOUS_HEARTBEAT_ENABLED=false` the agent created DAGs that launched wave-0 and then sat forever, with no error surfaced anywhere.

`DAGOrchestrator` gains `clock_wired` (explicit, set by whoever installs the tick — inferring from `last_tick_at` would false-negative during the first interval). `main.py` logs at ERROR, `dag_create` refuses with a message naming the cause, and `dag_manage list` distinguishes "nothing is running" from "nothing can ever run".

## Test plan

- **33 new tests** across `tests/test_dag_delivery.py` (leg isolation, Telegram-unconfigured path, summary timeout fallback, message truncation, crash-resume, bounded give-up, batch cap) and `tests/test_dag_durability.py` (reaper fires past grace / not inside it / cancels primitive first / releases the active-DAG slot, token idempotency across re-sync, clock-wiring refusal).
- **DAG suite: 241 passed, 1 skipped.**
- **Zero regressions, verified by diff** rather than by absolute count: stashing the working tree and re-running the identical `dag or subtask or heartbeat or event` selection gives **40 failures before → 29 after, zero new, 11 fixed**. The 29 remaining are pre-existing and unrelated (sqlite-harness issues in `test_f061_*`, `test_heart`, `test_fact_enhancements`).
- Migration `069` applied against **real Postgres 17** inside a rolled-back transaction — `ALTER` / `CREATE INDEX` / `ALTER` all clean, no dev-DB mutation.

### Test hygiene fixed en route

The DAG suite had **11 pre-existing failures on any machine with a prod-shaped `.env`**: fixtures called bare `Settings()`, which inherited `NOUS_DAG_STALL_DETECTION_ENABLED=true` and tripped `store.create`'s stall-vs-timeout validator. `test_dag_store.py` already stated the intent in a comment ("so the tests don't care about ambient `NOUS_DAG_NODE_*` env") but implemented it incompletely — `_env_file=None` completes it. A suite that cannot run green on the author's machine masks real regressions.

**No eval-DB validation run:** F087 touches orchestration only, not recall / sleep / retrieval / compaction, so the eval-DB gate does not apply.

## Operator note — unrelated live misconfiguration found while investigating

The prod `.env` carries `NOUS_DEFAULT_STALL_TIMEOUT=3000`, but the real setting is **`NOUS_DAG_NODE_DEFAULT_STALL_TIMEOUT`**. pydantic-settings silently ignores the unknown name, leaving the 600 s default in force while `NOUS_DAG_NODE_DEFAULT_TIMEOUT` is raised to 6000 s for long-running work. With stall detection enabled, nodes are being killed after **600 s of inactivity** rather than the intended 3000 s. Not addressed by this PR (it is a config fix), documented in CLAUDE.md next to the flag.

## Out of scope

- Moving the tick off the heartbeat loop onto its own asyncio task — the heartbeat remains the orchestrator's only clock; this PR only makes that loud instead of silent.
- Phase-2 Critic gate evaluation — `gate` nodes still auto-pass.
- F064.4 / F064.5 v2 consumers.
- `DAGOrchestrator._bus` is now redundant (delivery owns the bus) but was left in place rather than churn the constructor signature and its tests.

Spec: `docs/superpowers/specs/2026-08-08-dag-durability-spine-design.md`


tfatykhov:
@codex please re-check — all 4 P2s addressed in b3c78c3, fixed as two families rather than four patches.

**Family A — the token roll-up's two-phase write** (P2 *keep the claim and roll-up in one transaction* + P2 *refresh the budget total*). Both stemmed from splitting the claim from the add. `count_node_tokens` is replaced by `claim_and_add_node_tokens`: the false→true claim, `tokens_used`, and the DAG's `tokens_consumed` increment now commit in one transaction, so a crash between them can no longer leave a node permanently marked counted with nothing added. `_count_node_tokens` also refreshes `dag.tokens_consumed` in memory — `_advance_dag` loads `dag` before `_sync_node_statuses`, so the same-tick budget check was reading a stale total and dispatching the successor wave. `dag` is threaded through `_sync_subtask_node` (optional, so direct callers still work).

**Family B — retry-state hygiene** (P2 *clear stale delivery errors* + P2 *avoid regenerating summaries*). `mark_delivered` now always writes `delivery_error`, so a successful retry clears what `bump_delivery_attempt` left behind — a row can no longer report both success and an error. And an agent-authored summary is cached to the new `execution_dags.delivery_summary` (added to migration 069) and reused on retry; `DeliveryOutcome.summary_authored` ensures only text that actually cost a turn is persisted, never the cheap template, and the write lands *before* the outcome is recorded so a crash in between still banks the LLM turn.

5 new tests covering each. DAG suite 246 passed / 1 skipped; migration re-validated against real Postgres 17 in a rolled-back transaction.

tfatykhov:
@codex please re-check — both round-2 P2s addressed in 9dc6cbe. They shared one root cause I had missed: the retry paths were never considered when the per-attempt and per-outcome bookkeeping was added.

**Token claim is per ATTEMPT, not per node.** Cleared `tokens_counted` at all three points a node is reset for re-execution — the retried node in `retry_node()`, the selectively-unblocked downstream nodes (a `cancel_cascade` node may have been *running and consuming* when it was cancelled), and the fix-stage `retry_as_is` branch, which is the automatic sibling of the manual path and had the identical bug. `tokens_used` now accumulates (`DAGNode.tokens_used + tokens`) rather than overwrites, since a retried node genuinely spent both attempts and the DAG total accumulates the same way — overwriting would make the node disagree with the DAG it rolls up into.

**Delivery state is per OUTCOME.** New `DAGStore.reopen_delivery()` clears all four delivery columns when `retry_node` reactivates a failed/partial DAG. You're right that this defeated the feature on exactly the DAGs an operator cared enough about to retry. The cached summary is cleared too — it described the previous outcome and must not be reused for the new one.

Incidental: the fix-stage branch's `dict[str, Any]` became `dict[str, object]`, clearing the pre-existing `F821` (the module imports only `TYPE_CHECKING, Literal`) on a line this commit already touched.

3 new tests. DAG suite 249 passed / 1 skipped.

tfatykhov:
@codex please re-check — both round-3 P2s addressed in a4e302b.

**Reopen delivery atomically with reactivation.** You're right, and this one stings: my previous commit added the delivery reset but committed it separately from the status update, reintroducing on a new path the exact two-phase-write hazard this PR fixed for token accounting in round 1. The failure mode is worse than either endpoint — a crash in between leaves a *running* DAG carrying a stale `delivered_at`, so it finishes normally after restart and is then silently never announced. `reopen_delivery` is replaced by `reactivate_for_retry`, which sets `status`/`started_at` and clears all four delivery columns in one transaction. Status and delivery state are one logical transition and now commit as one.

**Retry token accounting for terminal nodes.** Also correct. `_count_node_tokens` swallows its exceptions by design so bookkeeping can't derail the status sync — but `_sync_subtask_node` then terminalizes the node, and later ticks only sync nodes still `running`, so one transient DB error stranded those tokens permanently. New `_reconcile_token_accounting` sweeps terminal subtask nodes still carrying `tokens_counted=False` and re-attempts the claim each tick. It is idempotent through the same flag, so it goes quiet once everything is accounted, and it runs *before* the budget check so a recovered count reaches enforcement on the same tick. When the subtask row is gone its usage is unknowable, so the node is claimed at zero rather than leaving the sweep spinning every tick for the life of the DAG.

3 new tests. DAG suite 253 passed / 1 skipped; ruff clean on the new files.

tfatykhov:
@codex please re-check — both round-4 findings addressed in 1bfe1fa.

**P1 — backfill terminal DAGs as delivered.** Correct and the most serious finding on this PR. `delivered_at` is nullable, so on upgrade every pre-existing terminal DAG read as undelivered, and with delivery on by default the sweep would have pushed the whole backlog to Telegram five per tick. I measured it against the local dev DB: **`UPDATE 580`** — roughly an hour of continuous notifications, and prod is larger. Migration 069 now backfills terminal rows to `COALESCE(completed_at, created_at)` immediately after the `ADD COLUMN` and before the index, so only outcomes that terminalize *after* the upgrade are announced. Verified against real Postgres 17 on seeded rows of all four terminal statuses plus the NULL-`completed_at` edge case — all suppressed, while a still-running DAG correctly stays eligible for its eventual real announcement.

**P2 — terminal DAGs unreachable for reconciliation.** Also correct, and it defeated the round-3 fix at exactly the moment it mattered. `_reconcile_token_accounting` is reached only via `_advance_dag`, which `tick()` runs only for pending/running DAGs — so a DAG whose accounting failed on its completion tick was terminalized by `_check_dag_completion` immediately after, and the promised next-tick retry never ran. The delivery sweep now reconciles each DAG before delivering it, which is the last point before the number is published. Guarded, so a reconciliation failure degrades to announcing the total we have rather than blocking delivery.

1 new test. DAG suite 254 passed / 1 skipped; ruff clean on the new files.

tfatykhov:
@codex please re-check — round-5 P2 addressed in 2e27143, but resolved at the root rather than as proposed. Flagging the deviation explicitly.

**The inconsistency you identified is real.** A DAG whose accounting only recovers in the pre-delivery sweep is announced `completed`, whereas accounting that landed a tick earlier would have run `_handle_budget_exceeded` and persisted `partial`. Same DAG, different final status, decided by whether a transient DB error happened to occur.

**I did not re-run enforcement after late reconciliation, because that cements the wrong half.** Enforcement exists to stop *future* work — it cancels pending/ready/awaiting_check nodes. A DAG whose nodes all completed has no future work to stop, so labelling it `partial` claims work was skipped when none was. The early path was the incorrect one, not the late one.

So `_handle_budget_exceeded` now **declines** when the overage curtailed nothing (no cancellable node, none still running) and lets `_check_dag_completion` finalize normally. Both timings then agree, which resolves your inconsistency, and a fully-successful DAG stops being mislabelled. The overage is not lost: it logs at WARNING and the delivery template renders `Tokens: used/budget` whenever a budget is set.

Two notes:
- `cancelled_any` had been computed and never read since F038 — a standing ruff `F841`, and exactly the signal this branch needed. Now cleared.
- The handler returns `bool` so the caller knows whether enforcement took over. My first cut missed that and the new test caught it: `_advance_dag` returns unconditionally after `_handle_budget_exceeded`, so declining meant `_check_dag_completion` never ran and the DAG would have sat `running` forever. It now falls through only when enforcement declines.

If you think `partial` is the correct label for an over-budget DAG that completed every node, say so and I'll invert it — but then the early path needs to be the one that changes, not the late one.

1 new test; `test_budget_exceeded` is unaffected (it does cancel a pending node, so enforcement still takes over). DAG suite 255 passed / 1 skipped.

tfatykhov:
@codex review

The previous review attempt failed with `Unknown error` rather than completing, so re-requesting. Current head is `2e27143`; CI is green (Lint + Tests) on it.

Since the last completed review (round 5), one change landed: `_handle_budget_exceeded` now declines when the overage curtailed nothing, and returns `bool` so `_advance_dag` only stops advancing when enforcement actually took over. Rationale and the explicit deviation from the suggested remedy are in the comment above.

tfatykhov:
@codex please re-check — both round-6 P2s addressed in fa988e7. Both were consequences of my own round-4/round-5 fixes, which is the pattern this review has been surfacing all along.

**Preserve earlier curtailment across ticks.** Correct, and a regression I introduced last commit. "Did enforcement curtail anything" is a property of the DAG's *history*, not of the current tick: an over-budget wave with both running work and pending successors cancels the successors on tick 1 and returns to let the running nodes finish, so on tick 2 the tick-local `cancelled_any` sees nothing to cancel, declines, and `_check_dag_completion` labels the DAG `cancelled` instead of `partial`. It now seeds from nodes already cancelled carrying the budget marker. That marker is promoted to a module-level `_BUDGET_CANCEL_ERROR` constant, since it is now load-bearing state rather than a bare log string.

**Require terminal reconciliation before delivery.** Also correct. `_reconcile_token_accounting` returned `None` whether or not the claims actually landed — and since `_count_node_tokens` swallows its errors, nothing upstream could tell. It now returns whether every terminal node is accounted for, reading back `node.tokens_counted` as the only honest signal, and the sweep defers when incomplete.

**One deliberate bound on that second fix:** deferral only happens while attempts remain, and the **final attempt publishes regardless**. An unannounced DAG is a worse outcome than a slightly low token count — this PR exists to guarantee the announcement, so blocking it indefinitely on bookkeeping would invert the priority it is trying to establish. The fallback logs at WARNING.

2 new tests. DAG suite 257 passed / 1 skipped; ruff clean on the new files.

tfatykhov:
@codex please re-check — both round-7 findings addressed in e94cd42. I verified both premises against the code before fixing.

**P1 — delivery writes could stamp a reactivated DAG.** Confirmed: `retry_node` (orchestrator.py:333) and `cancel_dag` (:312) do **not** take `self._lock` — only `tick` and `start_dag` do — so an operator retry genuinely interleaves with the delivery await, which lasts as long as a Telegram round-trip or a 120-second summary turn. All three delivery writes (`mark_delivered`, `save_delivery_summary`, `bump_delivery_attempt`) are now compare-and-swap, fenced on the DAG still being terminal **and** still undelivered.

I fixed it with the WHERE-clause discipline already used by `claim_and_add_node_tokens` rather than by taking the lock in `retry_node`: locking would serialize correctly but block an operator action for up to two minutes behind a summary turn. `mark_delivered` now returns whether it applied, so the sweep distinguishes "delivered" from "superseded" and logs it instead of silently assuming success.

**P2 — token claims are one-shot, so only settled subtasks may be claimed.** Also confirmed: `SubtaskManager.cancel()` flips running→cancelled without flushing worker token state, and its own docstring notes the worker "is not asyncio-cancelled mid-call". The claim now runs only when the subtask is settled; a still-pending/running subtask leaves the node retryable rather than freezing zeros as final.

I did treat `cancelled` as settled, which is a judgment call worth flagging: the row is terminal and the worker's eventual `complete()`/`fail()` is a no-op against the terminal guard, so its persisted counters really are the last word — under-reported, but never going to improve. Leaving it unsettled would spin the sweep every tick for the life of the DAG and defer every cancelled DAG's notification to the attempt cap, for an identical numeric result. Say the word if you read that differently.

2 new tests. DAG suite 259 passed / 1 skipped.

tfatykhov:
@codex please re-check — both round-8 findings addressed in d8b68bf.

**P1 — fence on a generation, not on terminal+undelivered.** You're right, and the failure mode is the same permanent exclusion my last fence was meant to prevent, just one step later: if the retried work re-terminalizes before the original await returns, the row is terminal *and* undelivered again, so the stale write claims the new outcome.

`execution_dags` gains `delivery_generation` (migration 069) — an optimistic-concurrency token naming *which* outcome a delivery is for. `reactivate_for_retry` bumps it, the sweep pins the value when it loads the DAG, and all three delivery writes fence on it. A superseded delivery's failure also no longer burns the replacement outcome's retry budget or parks a stale error on it: `bump_delivery_attempt` returns 0 and the caller logs and discards.

**P2 — bank the old attempt before resetting its claim.** Also correct, and it applies to `retry_node` as well as the fix stage: `retry_node` nulls `subtask_id` while the fix-stage relaunch overwrites it, so either way an attempt whose roll-up had failed transiently — or whose subtask had not settled yet — becomes invisible to reconciliation, and its usage is gone from both totals. New `_account_before_retry` runs first on both paths. Best-effort by design: it sits on the operator's retry path, and bookkeeping must not block a retry.

3 new tests. DAG suite 262 passed / 1 skipped; migration re-validated against real Postgres 17 in a rolled-back transaction (`UPDATE 575` backfilled).

tfatykhov:
@codex please re-check — round-9 P2 addressed in 954ecce, plus the systematic pass it should have prompted two rounds ago.

**The finding is right, and the meta-point is the real defect.** I enumerated all three `tokens_counted` reset sites in my round-2 commit message, then covered only two of them in round 8. You found the third. That is me fixing the sites named in a review comment instead of the family I had already identified.

So rather than patch and wait for round 10, I audited every mutation site behind the three invariants:
- **`tokens_counted` writes** — 3 reset sites (retried node, downstream unblock, fix-stage retry); all three now call `_account_before_retry` first.
- **`subtask_id` writes** — 2; both safe. The relaunch at `_launch_subtask_node` is only reachable via pending, and every path to pending either never launched (`_defer_node`, `_recover_stale_ready_nodes`, initial create) or banked first.
- **delivery-column writes** — 3 fenced on generation, plus `reactivate_for_retry`, which is correctly unfenced because it *is* the invalidator.

I also wrote the invariants into the orchestrator module docstring: DAG outcome (generation), node attempt (`tokens_counted`), subtask settlement. These three lifetimes are what nearly every finding on this PR reduced to, and their mutation sites aren't discoverable from any single function — each was found separately during review. Recording them means the next person doesn't rediscover them one comment at a time.

1 new test. DAG suite 263 passed / 1 skipped.

tfatykhov:
@codex — round 10: first finding fixed in 2ccc026, second one I'm pushing back on with evidence.

## Preserve the old attempt when banking fails — fixed, but per-path

You're right that best-effort banking plus an unconditional reset means silent permanent loss. But making the reset conditional *everywhere* was also wrong, and the existing suite caught it: `test_retry_recovers_cancel_cascade_descendants` went red because a cascade-cancelled descendant whose subtask row still read `running` was refused and **stranded forever**. That trades a bookkeeping loss for an availability loss, in a PR whose entire purpose is removing silent dead ends.

So the response is now chosen by what refusing actually costs:
- **directly-retried node** → raise; the operator just tries again, and these failures are transient.
- **fix stage** → defer *before* the attempt bump, so deferral is free. (My first cut checked after the bump and would have silently burned the parent's only fix attempt.)
- **downstream descendants** → proceed, logging at ERROR. They get no automatic second chance. The loss is now loud rather than silent, which I take to be the substance of your objection.

## Cancelled subtasks as settled — holding, with evidence

I checked the code rather than re-arguing. `complete()` and `fail()` write `tokens_in`/`tokens_out` **only** inside an UPDATE guarded by `status NOT IN ('cancelled','completed','failed')` (`nous/heart/subtasks.py:162`), and there is **no incremental token-write path anywhere** — counters are written once, at completion. So a cancelled row's counters can never change again, by construction.

That makes both options numerically identical: claiming yields `+0`; excluding yields no contribution, also `+0`. Neither recovers the lost usage. What differs is behaviour — excluding means `tokens_counted` stays false forever, so `_reconcile_token_accounting` re-fetches that subtask every tick for the life of the DAG and reports incomplete, which defers **every** cancelled-node DAG's notification to the attempt cap before publishing anyway.

So I read "irreversibly claim stale counters" as describing a loss that already happened upstream. The real defect is that **`SubtaskManager.cancel()` does not flush in-flight token usage** — that's in the subtask subsystem, not this reconciliation, and fixing it here is not possible. I'd rather file it than paper over it inside F087, which is already 10 rounds deep.

If you disagree, the concrete thing that would change my mind is a path where a cancelled subtask's counters can still be updated — if that exists I've misread the guard and I'll invert this immediately.

DAG suite 265 passed / 1 skipped.

tfatykhov:
Filed the cancel-flush root cause as #586 — `SubtaskManager.cancel()` does not flush in-flight token usage.

That issue carries the evidence from my previous comment (the terminal guard at `nous/heart/subtasks.py:162`, no incremental token-write path) plus three candidate fixes, and notes the property that makes it worth doing: the more aggressively the system cancels to control spend, the more spend it fails to see.

It is deliberately out of scope here — the defect is in the subtask subsystem and cannot be fixed from F087's reconciliation, which only reads counters that are already lost by the time it runs.

tfatykhov:
@codex please re-check — all three round-11 findings fixed in 7cc6ae0. One correction to your summary comment first.

**Correction:** the summary says manual retry now raises for "every downstream node that will be unblocked." It does not, deliberately. Only the *directly retried* node raises. Descendants proceed with an ERROR log, because refusing there strands a recoverable node permanently — your own suite caught that as `test_retry_recovers_cancel_cascade_descendants` when I did make it uniform. The per-path rationale is in the module docstring.

**1. Legacy path discarded the usage — this one made the feature inert by default, and it's the most serious of the three.** `_execute_legacy` received `_usage` and called `complete()` without `tokens_in`/`tokens_out`; `subtask_hardening_enabled` defaults **False**, so that is the default path, and F087's roll-up reads exactly those columns. The headline accounting fix was therefore recording zero for every DAG subtask out of the box — and freezing it, since the claim is one-shot. Same class as the original "`update_dag_tokens` has no callers" finding, which is a pointed thing to have missed twice. Now persisted as `input_tokens`/`output_tokens`/`tool_calls`, matching the hardened path.

**2. Reaper could kill work that never started.** `node.started_at` is set at dispatch; the subtask stays `pending` until dequeued. With 2 workers and a five-deep queue that gap readily exceeds timeout+grace. It now checks the primitive's status and skips a queued subtask — the lookup sits after the elapsed check, so the healthy path is unaffected.

**3. `cancel_dag` rewrote terminal `partial` DAGs.** `partial` was missing from the early-return list. The check now uses the shared `_TERMINAL_DAG_STATUSES` constant so the cancel path and the delivery sweep cannot drift apart again.

2 new tests. DAG suite 267 passed / 1 skipped. I also stash-diffed the whole `dag or subtask` selection since #1 touches `subtask_worker.py`: 23 failures before, 23 after, identical set — all pre-existing.

tfatykhov:
@codex please re-check — round-12 P2 fixed in ac9da7f.

You're right that my previous fix was a half-measure. Sparing a `pending` subtask protects it only while it waits; the moment a worker dequeues a long-queued task its status flips to `running` while `node.started_at` still holds the dispatch time, so the next tick reaps work that just began.

The underlying error was measuring from the wrong clock, so I fixed the measurement rather than adding a second status guard. `node.started_at` is now treated as what it actually is — an **upper bound**, useful only as a cheap pre-filter, since `subtask.started_at` (set at dequeue, `subtasks.py:109`) is always >= it, so a node inside budget by the dispatch clock is certainly inside it by the execution clock. Past that pre-filter the decision uses the subtask's real `started_at`. The healthy path still costs no extra query.

2 new tests, deliberately paired so the fix can't silently disable the backstop: a task dispatched 5000s ago but executing for 10s is spared, and one genuinely executing for 900s is still reaped.

DAG suite 269 passed / 1 skipped.

tfatykhov:
@codex please re-check — round-14 lock finding fixed in ad857d0; the second one I'm recording as unfixable from that call site rather than papering over.

**Delivery sweep outside the state-machine lock — fixed.** You're right and this one is squarely my error, not a subtle interaction: I put blocking network I/O inside a mutex meant for in-memory state transitions. Worst case is `batch_size` × summary timeout = 5 × 120s holding `_lock`, during which `start_dag` cannot acquire it and `dag_create` hangs after having already inserted its DAG.

The sweep now runs after the lock is released, with its own `_delivery_lock` so sweeps still serialize against each other and a DAG cannot be announced twice by overlapping ticks. When a sweep is in flight the next tick **skips** rather than queueing, so ticks can't pile up behind a summary turn. Both new tests drive the real concurrency rather than asserting structure: `start_dag` completes while a delivery is parked mid-await, and two overlapping ticks announce a DAG exactly once.

**Legacy failure-path token usage — real, but not fixable here.** I checked before responding: `run_turn` attaches no usage to the exceptions it raises, so at the point `fail()` is called the number genuinely does not exist. The success branch was fixable because it was discarding usage it already had; this branch has nothing to discard. The hardened executor avoids it entirely via its `HardenedRunState` side channel, which survives the exception.

Logged on #586 with two additional candidate fixes — give the legacy path an equivalent side channel, or retire it by defaulting `subtask_hardening_enabled` on, which `_execute_legacy`'s own docstring ("Removed in PR-6 once flag is locked on") suggests is the intended direction anyway. I'd rather that be a deliberate decision than a hurried change in a PR that is already 14 rounds deep.

DAG + f066 suites: 309 passed / 1 skipped — and I verified with `--collect-only` that the selection actually includes the f066 files this time, having reported green over three red commits earlier by trusting a `-k` filter that silently excluded them.

tfatykhov:
@codex please re-check — both round-15 findings fixed in e57c6ac. Both are follow-ons from my round-14 fix, and the second is one that fix made *more* likely.

**Detach the sweep from tick completion.** You're right, and my round-14 fix was incomplete for a reason worth naming: I fixed lock contention without following the await chain up into the caller. `HeartbeatRunner._loop` does `await dag_orchestrator.tick()`, so awaiting the sweep inside `tick()` still stalled DAG advancement, reaping, dynamic sync, digest and tuning. The sweep now runs on its own task; `tick()` returns as soon as the state machine is done. `_delivery_lock` still prevents overlap, and a strong reference prevents mid-flight GC silently dropping a notification.

Detaching had a consequence the tests surfaced immediately, which I think is worth flagging since it's a genuine production concern rather than test noise: a DAG run failed with `no such table` because a sweep outlived the fixture teardown — i.e. in production, an in-flight sweep meeting a closed pool at shutdown. `shutdown_components` now drains it after the heartbeat stops launching new ones but before the pool closes, bounded at 30s so a hung Telegram call can't block shutdown. An abandoned sweep re-delivers after restart, which is what the durable queue is for.

**Fence token reconciliation against retry resets.** Also correct. `claim_and_add_node_tokens` now takes `expected_subtask_id` and matches it in the WHERE clause, so a stale snapshot cannot re-claim a freshly reset flag using the old subtask's tokens — which would have double-added that attempt *and* left the flag true so the replacement could never be counted. Same optimistic-concurrency discipline as `delivery_generation`, applied to the attempt rather than the outcome.

1 new test (tick returns while a delivery is parked mid-await) plus every delivery-observing test updated to drain the detached sweep. DAG + f066: 310 passed / 1 skipped. CI verdict pending — I'll confirm it rather than assume.

tfatykhov:
@codex please re-check — round-16 P2 fixed in ab0ff44.

Correct, and worth naming precisely: **both of those tests are mine**, written in round 14 when `tick()` still awaited the sweep. Round 15 detached it and left them latently flaky. They await `tick_task` / `first` — but the tick completing early is now exactly the property under test, so awaiting it does not wait for the sweep's database writes, and the fixture can tear the DB down mid-`mark_delivered()`. That reproduces the same `no such table` the detach already caused once.

The part I'd flag for myself: they *passed*, which is worse than failing. A green run on those two was not evidence of correctness, only of timing. Both now drain explicitly via `wait_for_delivery()`, and I verified by repetition rather than a single pass — 3 consecutive runs, 28 passed each.

CI on the previous head (`e57c6ac`) came back **success**. DAG + f066 suites: 310 passed / 1 skipped.

tfatykhov:
@codex please re-check — both round-17 findings fixed in e033668. Both were in `_reap_overrun_nodes` and both were mine.

**Defer when the execution-clock lookup fails.** Correct. `except Exception: subtask = None` swallowed the error and fell through to the dispatch clock — re-creating the exact false positive the execution-clock check was added to prevent, triggered by nothing worse than a transient DB blip. Reaping is destructive and this backstop is never urgent to the second, so a failed read now defers to the next tick rather than deciding from a clock known to be wrong.

**Preserve a failure that wins the race.** Also correct, and the asymmetry was arbitrary on my part: I handled `completed` because that was the case in front of me, not because `failed` was different. Both are self-reached terminal states, cancel() equally declines to touch them, and falling through replaced the primitive's real error with the generic wall-clock text — lost permanently, since terminal nodes are never re-synced.

Both now route through `_sync_subtask_node`. The distinction I've made explicit in the comment is **who ended it**: `completed`/`failed` means the row carries the truth; `cancelled` is the reaper's own doing, and there the wall-clock error is the honest description — so a subtask actually killed by the reaper still reports the timeout, as you asked.

2 new tests. The failure-race one asserts both directions — the real reason survives *and* the generic text is absent — because asserting only the former would pass on a message that concatenated both.

DAG + f066: 312 passed / 1 skipped.

tfatykhov:
@codex please re-check — round-18 P2 fixed in f409568.

Correct, and this is the sibling of your previous finding, in the same function, which I touched **in the same commit** and fixed only one of. `except Exception: settled = None` treated a failed read as proof the reaper won and overwrote the primitive's real terminal outcome with the generic wall-clock error. It now defers, which converges: the subtask is already cancelled, so a later tick either syncs its persisted outcome or reaps once the read succeeds.

Rather than patch the reported line, I audited every `_subtask_mgr.get` call site and every swallow-to-None in the orchestrator. This was the last one feeding a destructive decision — the rest either already defer (`_reconcile_token_accounting`, the pre-cancel lookup) or are cleanup paths where swallowing is correct (`_cancel_node`).

Worth saying plainly: this is the same "fixed the instance, not the family" pattern as round 9, which I had already written down as a lesson and then repeated on two adjacent lines of one function. The audit above is me trying to actually apply it rather than restate it.

1 new test — a failed confirmation leaves the node `running` and un-errored for the next tick, rather than stamping the generic failure. DAG + f066: 313 passed / 1 skipped. CI on `e033668`: success.

tfatykhov:
@codex please re-check — both round-19 findings fixed in a616310.

**Defer when cancellation leaves the subtask running.** Correct. `_cancel_node` swallows its errors, so a confirmation read of `running` means cancellation did not take — and failing the node there abandons a worker that is still executing, whose result is then discarded since terminal nodes are never re-synced.

Bounded deliberately: unbounded deferral would re-create the exact wedge this backstop exists to prevent, because an orphaned row whose worker died never settles. After 5 ticks the node is failed anyway, with an error stating that cancellation could not be confirmed and the subtask may still be running — rather than claiming a clean timeout it cannot vouch for.

**Three existing reaper tests went red on this, and they were the ones at fault.** They mocked `cancel()` succeeding while the subtask's status never changed — a world where cancellation silently failed. Production `cancel()` moves a running row to `cancelled`. I added a helper that models the real transition, so they now assert against production behaviour. Worth noting the implication: their previous green was not evidence the reaper worked.

**Retry mutation window.** Fixed by reordering rather than one large transaction: `reactivate_for_retry` now runs FIRST, before any node reset. That makes the DAG non-terminal immediately — so it leaves `get_undelivered_terminal_dags`' domain entirely — and bumps the generation for anything already in flight. Same window closed, materially less risk than threading every per-node reset through a single transaction. If you think the transaction is still warranted I'll do it, but I'd rather not restructure the mutation path this late without a concrete case the reorder misses.

3 new tests. DAG + f066: 315 passed / 1 skipped. CI on `f409568`: success.

tfatykhov:
@codex please re-check — all three round-20 findings fixed in 902994c.

**Retry is now one transaction, and my pushback last round was wrong.** I argued the reorder was equivalent and asked for a concrete case it missed; you supplied it. Reactivation-last let the sweep announce an outcome built from half-reset nodes; reactivation-first let an ordinary tick load a *running* DAG whose nodes were all still terminal, re-finalize it via `_check_dag_completion`, and strand a pending node inside a terminal DAG. The invalid state is the split itself, so no ordering removes it — only atomicity does. New `DAGStore.apply_retry` commits every node reset plus the status/generation/delivery transition together.

**The reap counter leaked across attempts.** `_reap_defer_counts` was cleared only inside the reaper's own branches, so a node settling normally via `_sync_subtask_node` kept its count, and a later retry reusing the node id inherited it — hitting the cap early and abandoning a live replacement worker. Now cleared for any node that is not running, which covers both settling and being reset to pending (the reaper runs before dispatch in the tick).

**Cancellation confirmation was subtask-only.** Check nodes now get the same confirmed, bounded deferral, with an error saying the dynamic check could not be disabled rather than claiming a clean timeout. Worth noting this is the third variant of the same shape — `_cancel_node` swallowing a failure that the caller then treats as success — and I fixed it for subtasks twice without asking what else that function cancels.

2 new tests. DAG + f066: 317 passed / 1 skipped.

